PR #61, opened 2021-08-20
Accept OPENSEARCH_URL for address

### LINKED ISSUE #49: Accept either ELASTICSEARCH_URL or OPENSEARCH_URL
To support backward compatibility, if addresses are not specified in config during new client api, we will look for addresses either  from ELASTICSEARCH_URL or OPENSEARCH_URL environment variable. if we find both, then will exit with error message that both can't be set .

VijayanB:
### Description
Accept OPENSEARCH_URL and ELASTICSEARCH_URL for backward compatibility, but not together.
 
### Issues Resolved
#49
 
### Check List

### DIFF
--- a/opensearch.go
+++ b/opensearch.go
@@ -42,9 +42,9 @@ import (
 	"sync"
 	"time"
 
+	"github.com/opensearch-project/opensearch-go/internal/version"
 	"github.com/opensearch-project/opensearch-go/opensearchapi"
 	"github.com/opensearch-project/opensearch-go/opensearchtransport"
-	"github.com/opensearch-project/opensearch-go/internal/version"
 )
 
 var (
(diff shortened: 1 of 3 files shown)


opensearch.go
```diff
@@ -143,14 +148,20 @@ func NewDefaultClient() (*Client, error) {
...
 // to configure the addresses; use a comma to separate multiple URLs.
 //
+// It's an error to set both OPENSEARCH_URL and ELASTICSEARCH_URL.
+//
 func NewClient(cfg Config) (*Client, error) {
 	var addrs []string
 
 	if len(cfg.Addresses) == 0 {
-		addrs = addrsFromEnvironment()
+		envAddress, err := getAddressFromEnvironment()
```
jmazanec15, line 160: This may be a Go question, but why use envAddress as opposed to addrs directly?

VijayanB: Thats correct :) 
":=" defines new variable on left side. Other alternate is

```
var err error
addrs, err = getAddressFromEnvironment()

```
Do you prefer this?

jmazanec15: Right, what I meant was that in Go, is this the preferred convention? 

I lean towards the approach that you just mentioned, but Im wondering if there is a convention for this?

VijayanB: AFAIK, shorter declaration is preferred since it will infer type and default value by compiler. This is also preferred, if in future some one updates the return type, they don't have to update the caller as well.

jmazanec15 (approved):

PR #81, opened 2022-01-05
Support Signing Request feature

### LINKED ISSUE #75: [FEATURE] Add IAM authentication option
If people want to connect to an Amazon OpenSearch Service cluster, they have to figure out how to use this client alongside some third-party signing library and/or the AWS SDK, which is non-trivial due to how we (by design) conceal the underlying HTTP requests that the client makes. We should offer IAM signing as an option, just like the OpenSearch CLI does.

**Is your feature request related to a problem?**
If you want to use this client with an Amazon OpenSearch Service cluster that has IAM authentication rather than basic authentication, good luck.

**What solution would you like?**

When initializing the client, an AuthType option. If basic (or null), accept username and password. If IAM, accept options for access key, secret key, session token, region, and service.

**What alternatives have you considered?**

An additional Go IAM signing library for use on top of the client. But given that it would only work with the client, it seems better and easier to just add it to the client.

**Do you have any additional context?**

https://github.com/opensearch-project/OpenSearch/issues/1400

VijayanB:
### Description
* Add support to sign request
* Add support to sign using AWS Sigv4
 
### Issues Resolved
#75 
 
### Check List


opensearchtransport/opensearchtransport.go
```diff
@@ -439,6 +451,13 @@ func (c *Client) setReqAuth(u *url.URL, req *http.Request) *http.Request {
 	return req
 }
 
+func (c *Client) signRequest(req *http.Request) error {
+	if c.signer != nil {
```
dblock, line 455: Is this intentionally passing through, should it fail if there's no signer?

signer/aws/aws.go
```diff
@@ -0,0 +1,66 @@
...
+//OpenSearchService AWS OpenSearchService Name
+const OpenSearchService = "es"
+
+// Signer is a interface that will implement opensearchtransport.Signer
+type Signer struct {
+	session session.Session
+	service string
+}
+
+//NewSigner returns an instance of Signer
```
dblock, line 32: Nit: space after `//`, there's a few others like this.

signer/aws/aws.go
```diff
@@ -0,0 +1,66 @@
...
+	}
+	return &Signer{
+		session: *awsSession,
+		service: OpenSearchService,
+	}, nil
+}
+
+//SignRequest signs the request using SigV4
+func (s Signer) SignRequest(req *http.Request) error {
+
```
dblock, line 46: remove blank line, maybe auto-format the file

signer/aws/aws.go
```diff
@@ -0,0 +1,66 @@
...
+	signer := v4.NewSigner(s.session.Config.Credentials)
+	return sign(req, s.session.Config.Region, s.service, signer)
+
+}
+
+func sign(req *http.Request, region *string, serviceName string, signer *v4.Signer) (err error) {
+	if region == nil || len(*region) == 0 {
+		return fmt.Errorf("aws Region cannot be empty")
```
dblock, line 54: lowercase region

dblock (approved):
Some smallish comments, looks good to my Go knowledge.

signer/aws/aws.go
VijayanB: Ack

signer/aws/aws.go
VijayanB: Ack

signer/aws/aws.go
VijayanB: Ack

dblock (approved):

jmazanec15 (approved):

PR #96, opened 2022-03-19
golangci-lint integration

dzyanis:
### Description
[golangci-lint](https://golangci-lint.run/) is a Go linters aggregator. It helps improve the code quality and prevents bugs. I'd recomede to integrate golangci-lint and then enable linter one by one.
 
### Issues Resolved
There are many issues with formatting, misspellings and even potentially dangerous bugs.
 
### Check List

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,7 @@
 tmp/
 *.test
 
-#jetBrains editors
+# editors
 .idea
 .vscode
 bin/
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,5 @@ tmp/
 
 #jetBrains editors
 .idea
+.vscode
+bin/
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,0 +1,117 @@
+run:
+  modules-download-mode: readonly
+  issues-exit-code: 0
+
+linters-settings:
+  exhaustive:
+    default-signifies-exhaustive: false
+  gci:
+    sections:
+      - standard
+      - default
+      - prefix(github.com/opensearch-project/opensearch-go)
+  goconst:
+    min-occurrences: 5
+    ignore-tests: true
+  godot:
+    scope: toplevel
+    exclude:
+      - go-sumtype:decl
+      - check interfaces
+    capital: true
+  godox:
+    keywords:
+      - BUG
+      - FIXME
+      - HACK
+  goimports:
+    local-prefixes: github.com/opensearch-project/opensearch-go
+  gosimple:
+    go: "1.18"
+    checks: ["all"]
+  lll:
+    line-length: 130
+    tab-width: 4
+  nolintlint:
+    allow-unused: false
+    allow-leading-space: false
+    allow-no-explanation: []
+    require-explanation: true
+    require-specific: true
+  staticcheck:
+    go: "1.18"
+    checks: ["all"]
+  stylecheck:
+    go: "1.18"
+    checks: ["all"]
+    dot-import-whitelist: []
+    initialisms: []
+    http-status-code-whitelist: []
+  unparam:
+    check-exported: true
+  unused:
+    go: "1.18"
+  whitespace:
+    multi-if: false
+    multi-func: false
+  tagliatelle:
+
+linters:
+  enable-all: true
+  disable:
+    # TODO: enable one by one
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -179,8 +179,11 @@ cluster.clean: ## Remove unused Docker volumes and networks
 	docker network prune --force
 	docker system prune --volumes --force
 
+linters:
+	./bin/golangci-lint run ./... --timeout=5m
 
-
+linters.install:
+	curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s
```
dblock, line 186: This is a moving target, could we install a released version please?

dblock (changes requested):
LGTM, see below.]

I'd prefer if some linters were enabled now and code cleaned up as part of this PR, instead of adding the linter and disabling everything.

Makefile
dzyanis: As I know we even can't specify a version in the config. I took it from the doc https://golangci-lint.run/usage/install/#local-installation

dblock: I would use a tag, e.g. https://github.com/golangci/golangci-lint/tree/v1.45.2 but it's NBD.

dblock (approved):

dzyanis:
> I'd prefer if some linters were enabled now and code cleaned up as part of this, or a subsequent quick follow PR, instead of adding the linter and disabling everything.

Actually, I disabled only those linters which return errors. :-)

I'm going to make other PRs and next one will fix all code style issues. Probably enable linters:
    - gofmt
    - gofumpt
    - gci
    - goimports
    - whitespace
    - lll
    
  @dblock but what do you think about naming in Makefile?
  ```bash
 $ make lint
lint             linters          linters.install
  ```
Don't you think that people can be confused by having `lint` and `linters`?

dblock:
> > I'd prefer if some linters were enabled now and code cleaned up as part of this, or a subsequent quick follow PR, instead of adding the linter and disabling everything.
> 
> Actually, I disabled only those linters which return errors. :-)
> 
> I'm going to make other PRs and next one will fix all code style issues. Probably enable linters: - gofmt - gofumpt - gci - goimports - whitespace - lll
> 
> @dblock but what do you think about naming in Makefile?
> 
> ```shell
> $ make lint
> lint             linters          linters.install
> ```
> 
> Don't you think that people can be confused by having `lint` and `linters`?

🤷  I guess `lint` is better.

Makefile
VijayanB: IMO, we don't need install command as part of make file. We can keep only mostly used command as part of Make file and add instruction or link to how to install golangci-lint at README page. WDYT?

.gitignore
```diff
@@ -2,4 +2,6 @@ tmp/
 *.test
 
 #jetBrains editors
```
VijayanB, line 4:
```suggestion
#editors
```

Makefile
dzyanis: > we don't need install command as part of make file

I think you are right

> add instruction or link on how to install golangci-lint at README page

golangci-lint is optional and I think we can remove it and that's it. Let's start from that.

VachaShah:
@dzyanis Can you please fix the DCO? One of the commits might be missing a signoff.

.golangci.yml
```diff
@@ -0,0 +1,117 @@
...
+  unused:
+    go: "1.18"
+  whitespace:
+    multi-if: false
+    multi-func: false
+  tagliatelle:
+
+linters:
+  enable-all: true
+  disable:
```
VachaShah, line 61: Are we keeping all linters disabled in this PR?

dzyanis: Not really. We enabled all linters but disabled the ones that return warnings. In the next PRs I'm going to fix them and enable one by one

VachaShah: Oh my bad, can you also create an issue with the list of linters to be enabled may be on a checklist? It would be easier for the upcoming PRs.

VachaShah (approved):

VijayanB (approved):

dzyanis:
@VachaShah @VijayanB is there anything else what I should do? Who is responsible for merging?

PR #103, opened 2022-04-28
Add setting for allow_leading_wildcard to search

nevins-b:
### Description
Allow for setting, disabling, `allow_leading_wildcard` on a search

### Issues Resolved

### DIFF
--- a/opensearchapi/api.search.go
+++ b/opensearchapi/api.search.go
@@ -63,6 +63,7 @@ type SearchRequest struct {
 
 	AllowNoIndices             *bool
 	AllowPartialSearchResults  *bool
+	AllowLeadingWildcard       *bool
 	Analyzer                   string
 	AnalyzeWildcard            *bool
 	BatchedReduceSize          *int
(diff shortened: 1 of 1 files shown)


dblock:
Thanks! Happy to take this with some tests, please?

nevins-b:
@dblock Sure, I didn't see a place to add tests. I could add something to the benchmark tests but that didn't feel right. Suggestion of where I would add some?

dblock:
Not a go expert nor have I looked in detail into opensearch-go to say where those tests should go, but looks like there's a bunch of tests that _are_ running (https://github.com/opensearch-project/opensearch-go/runs/6070092469?check_suite_focus=true). I'd like us not to merge untested code, so if you need to introduce a new set of tests and just test this one addition, let's do it?

VachaShah:
@nevins-b Thank you for this PR! May be the tests can go [here](https://github.com/opensearch-project/opensearch-go/blob/main/opensearchapi/opensearchapi_integration_test.go#L44)?

wbeckler:
Hi @nevins-b, do you think you'll want this merged? If nobody adds tests or moves this forward, I propose we close this PR.

Jakob3xD:
While looking into the PR I noticed that `allow_leading_wildcard` does not seems to be a parameter for a search request.
`allow_leading_wildcard` is a filed of `query_string` in the search body. The body is given via the user so adding this to search prams would not work. https://opensearch.org/docs/latest/query-dsl/full-text/query-string/

I suggest closing the PR. @wbeckler

wbeckler:
I'm okay with closing this. Any other maintainers have a thought?

PR #133, opened 2022-06-30
fix: update deps

matelang:
### Description
AWS SDK v2 had an issue that caused erroneous v2 signer behaviour when Content-Length header is set.
See: https://github.com/aws/aws-sdk-go-v2/issues/1728

AWS SDK v2, specifically module `github.com/aws/aws-sdk-go-v2` version `v1.16.6`, released yesterday adds a fix for this issue.

As housekeeping I've updated the other deps as well.


VachaShah (approved):
Thank you @matelang for updating this!

VijayanB (approved):

PR #216, opened 2023-01-17
Adding support and examples for AOSS

### LINKED ISSUE #204: [FEATURE] Add support for OpenSearch Serverless

- OpenSearch serverless service identifier changes from `es` to `aoss`.
- Sigv4 requires a hash
- Info API doesn't exist
- What else?

Document support for OpenSearch Serverless in the compatibility matrix, add integration tests.

Coming from https://github.com/opensearch-project/opensearch-clients/issues/44

VachaShah:
Signed-off-by: Vacha Shah <vachshah@amazon.com>

### Description
Adding support and examples for AOSS.

### Issues Resolved
#204 

### DIFF
--- a/USER_GUIDE.md
+++ b/USER_GUIDE.md
@@ -1,8 +1,8 @@
 - [User Guide](#user-guide)
-  - [Example](#example)
-  - [How to use IAMs as authentication method](#how-to-use-iams-as-authentication-method)
-      - [AWS SDK V1](#aws-sdk-v1)
-      - [AWS SDK V2](#aws-sdk-v2)
+	- [Example](#example)
+	- [Amazon OpenSearch Service](#amazon-opensearch-service)
+			- [AWS SDK V1](#aws-sdk-v1)
+			- [AWS SDK V2](#aws-sdk-v2)
 
 # User Guide
 
@@ -138,7 +138,7 @@ func main() {
 
 ```
 
-## How to use IAMs as authentication method
+## Amazon OpenSearch Service
 
 Before starting, we strongly recommend reading the full AWS documentation regarding using IAM credentials to sign
 requests to OpenSearch APIs.
@@ -223,7 +223,7 @@ func main() {
 
 #### AWS SDK V2
 
-AWS SDK V2 for Go can be used for various service options. To create a client, please see the example below:
+Use the AWS SDK v2 for Go to authenticate with Amazon OpenSearch service.
 
 ```go
 package main
@@ -239,7 +239,7 @@ import (
 	requestsigner "github.com/opensearch-project/opensearch-go/v2/signer/awsv2"
 )
 
-const endpoint = "" // e.g. https://opensearch-domain.region.com or serverless collection endpoint
+const endpoint = "" // e.g. https://opensearch-domain.region.com or Amazon OpenSearch Serverless endpoint
 
 func main() {
 	ctx := context.Background()
@@ -271,7 +271,7 @@ func main() {
 
 	indexName = "go-test-index"
 
-    // Define index mapping.
+	// Define index mapping.
 	mapping := strings.NewReader(`{
 	 "settings": {
 	   "index": {
--- a/USER_GUIDE.md
+++ b/USER_GUIDE.md
@@ -283,6 +283,7 @@ func main() {
 	// Create an index with non-default settings.
 	createIndex := opensearchapi.IndicesCreateRequest{
 		Index: indexName,
+		Body:  mapping,
 	}
 	createIndexResponse, err := createIndex.Do(context.Background(), client)
 	if err != nil {
--- a/USER_GUIDE.md
+++ b/USER_GUIDE.md
@@ -1,6 +1,8 @@
 - [User Guide](#user-guide)
-    - [Example](#example)
-    - [How to use IAMs as authentication method](#how-to-use-iams-as-authentication-method)
+  - [Example](#example)
+  - [How to use IAMs as authentication method](#how-to-use-iams-as-authentication-method)
+      - [AWS SDK V1](#aws-sdk-v1)
+      - [AWS SDK V2](#aws-sdk-v2)
 
 # User Guide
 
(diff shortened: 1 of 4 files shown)


dblock:
I fully expect this not to be needed soon for AOSS, this is work in progress server-side.

dblock:
The root API will be (re)added to OpenSearch Serverless.

I propose we remove the product check altogether, similar to https://github.com/opensearch-project/opensearch-ruby/issues/130. We try to do a version check to see if the client is compatible with the server. I believe this was inherited from Elastic that was trying to maintain some sense of version compatibility without semver, but since we follow semver it's easy to know now whether your client will work against a server or not. My concern is that a version request every single time a new instance of a client is created is not insignificant overhead (in a web app example this could end up making a root request every time a web app serves a client search request!), and is an unexpected side effect when users try to use the client.

VachaShah:
> I fully expect this not to be needed soon for AOSS, this is work in progress server-side.

Sounds good! I removed the content-length change from this PR.

VachaShah:
> The root API will be (re)added to OpenSearch Serverless.
> 
> I propose we remove the product check altogether, similar to [opensearch-project/opensearch-ruby#130](https://github.com/opensearch-project/opensearch-ruby/issues/130). We try to do a version check to see if the client is compatible with the server. I believe this was inherited from Elastic that was trying to maintain some sense of version compatibility without semver, but since we follow semver it's easy to know now whether your client will work against a server or not. My concern is that a version request every single time a new instance of a client is created is not insignificant overhead (in a web app example this could end up making a root request every time a web app serves a client search request!), and is an unexpected side effect when users try to use the client.

Sure! I removed the product info check altogether.

VachaShah:
@dblock This is ready for review.

USER_GUIDE.md
```diff
@@ -1,6 +1,9 @@
 - [User Guide](#user-guide)
-    - [Example](#example)
-    - [How to use IAMs as authentication method](#how-to-use-iams-as-authentication-method)
+	- [Example](#example)
+	- [How to use IAMs as authentication method](#how-to-use-iams-as-authentication-method)
+			- [AWS SDK V1](#aws-sdk-v1)
+			- [AWS SDK V2](#aws-sdk-v2)
+			- [AWS SDK V2 with OpenSearch Serverless](#aws-sdk-v2-with-opensearch-serverless)
```
dblock, line 6: Tabs vs. spaces? 🤔

USER_GUIDE.md
```diff
@@ -283,3 +286,91 @@ func main() {
 }
 
 ```
+
+#### AWS SDK V2 with OpenSearch Serverless
```
dblock, line 290: We should merge the managed and serverless portions, these are the same now no?

dblock (changes requested):
I think we should make a separate PR with an entry that removes the version check, so we can make it very visible. Should we major increment version, too?

For docs I'd say managed and serverless work the same and the only thing you want to change is the service signing name. Let's say that in the user guide?

USER_GUIDE.md
VachaShah: My bad, fixed

USER_GUIDE.md
VachaShah: Combined them.

VachaShah:
> I think we should make a separate PR with an entry that removes the version check, so we can make it very visible. Should we major increment version, too?
> 
> For docs I'd say managed and serverless work the same and the only thing you want to change is the service signing name. Let's say that in the user guide?

Created a separate PR #219 for removing the version check and incrementing the version. 
Updated the user guide.

USER_GUIDE.md
```diff
@@ -1,6 +1,8 @@
 - [User Guide](#user-guide)
-    - [Example](#example)
-    - [How to use IAMs as authentication method](#how-to-use-iams-as-authentication-method)
```
dblock, line 3: I know we already had this here, but shouldn't it be "IAM"?

More generally I think we might want to align this on other docs and say "Amazon OpenSearch Service" as a title.

USER_GUIDE.md
```diff
@@ -221,32 +223,39 @@ func main() {
 
 #### AWS SDK V2
 
+AWS SDK V2 for Go can be used for various service options. To create a client, please see the example below:
```
dblock, line 226: I don't think we need to be polite in docs, would remove "please".

I also prefer succinct versions, so "Use the AWS SDK v2 for Go to authenticate with Amazon OpenSearch service."

USER_GUIDE.md
```diff
@@ -221,32 +223,39 @@ func main() {
...
+	"github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/config"
-	"github.com/opensearch-project/opensearch-go/v2/opensearchapi"
+	opensearch "github.com/opensearch-project/opensearch-go/v2"
+	opensearchapi "github.com/opensearch-project/opensearch-go/v2/opensearchapi"
 	requestsigner "github.com/opensearch-project/opensearch-go/v2/signer/awsv2"
 )
 
-const endpoint = "" // e.g. https://opensearch-domain.region.com
+const endpoint = "" // e.g. https://opensearch-domain.region.com or serverless collection endpoint
```
dblock, line 242: Naming-wise we call it Amazon OpenSearch Serverless

USER_GUIDE.md
```diff
@@ -260,26 +269,51 @@ func main() {
 		log.Fatal("client creation err", err)
 	}
 
-	ping := opensearchapi.PingRequest{}
+	indexName = "go-test-index"
 
-	resp, err := ping.Do(ctx, client)
+    // Define index mapping.
```
dblock, line 274: align

dblock (approved):
Looks good! I have some potential nits for docs, feel free to ignore them and merge.

VachaShah:
@dblock I addressed the comments for documentation, it needs your approval once more.

dblock (approved):

USER_GUIDE.md
```diff
@@ -260,26 +269,51 @@ func main() {
...
 
-	resp, err := ping.Do(ctx, client)
+	// Define index mapping.
+	mapping := strings.NewReader(`{
+	 "settings": {
+	   "index": {
+	        "number_of_shards": 4
+	        }
+	      }
+	 }`)
```
VijayanB, line 281: Where is this mapping used?

signer/awsv2/sdkv2signer.go
```diff
@@ -68,6 +68,7 @@ func (s *awsSdkV2Signer) SignRequest(r *http.Request) error {
 	}
 
 	hash, err := hexEncodedSha256OfRequest(r)
+	r.Header.Set("X-Amz-Content-Sha256", hash)
```
VijayanB, line 71: Shall we move this to constant?

USER_GUIDE.md
VachaShah: My bad, I added its usage

signer/awsv2/sdkv2signer.go
VachaShah: Since its just used here, I kept it as a string

VachaShah:
@VijayanB @dblock Review again please?

dblock (approved):

Issue #236, opened 2023-02-21
[FEATURE] SnapshotDelete use []string


Jakob3xD:
### Is your feature request related to a problem?
[SnapshotDeleteRequest](https://pkg.go.dev/github.com/opensearch-project/opensearch-go/v2@v2.2.0/opensearchapi#SnapshotDeleteRequest) currently accepts `string` for Snapshot. Most other Requests use `[]string` and I don´t  see a reason why only `string` is used. The API itself uses `[]string`.
https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/action/admin/cluster/snapshots/delete/DeleteSnapshotRequest.java#L56

### What solution would you like?
Change the type from `string` to `[]string`

### What alternatives have you considered?
Otherwise the user need to join the slice into a string by doing `strings.Join(snapshots, ",")`

### Do you have any additional context?
Nope

PR #237, opened 2023-02-21
SnapshotDeleteRequest use []string instead of string

Jakob3xD:
### Description
SnapshotDeleteRequest use []string instead of string

### Issues Resolved
Closes [236](https://github.com/opensearch-project/opensearch-go/issues/236)


VachaShah (commented):
Thank you for this fix @Jakob3xD! Can you also add a test for the same?

Jakob3xD:
@VachaShah Are there any existing tests I can orientate on that handle snapshots or maybe something else, so I can see how you want the tests to look like and what structure they should have?

VachaShah:
> @VachaShah Are there any existing tests I can orientate on that handle snapshots or maybe something else, so I can see how you want the tests to look like and what structure they should have?

@Jakob3xD All the API tests currently exist in [opensearchapi_integration_test.go](https://github.com/opensearch-project/opensearch-go/blob/main/opensearchapi/opensearchapi_integration_test.go).

dblock (changes requested):
This is a breaking change for the client. Can we avoid it and have both versions of `string` and `string[]` and implement the former using the latter everywhere one may want to pass a single item?

Jakob3xD:
This is a breaking change but only because this was wrong in the first place. I should have opened my issue as bug and not as feature. IMO it is also easier to convert a string to []string by just doing `[]string{"Hello"}` instead of converting a []string to string by doing `strings.Join(stringvar, ",")` which needs to import the strings lib.

dblock:
@Jakob3xD You're right, still before I merge, is there a way to make the change backwards compatible? Does it make any sense to have two fields `Snapshot` and `Shapshots` for example?

Jakob3xD:
@dblock IMO it does not make sense to have both fields as it add unneeded overhead.
Other snapshot requests are snapshot specific and are therefore only a string.

I did't find more problems with wrong types.

@VachaShah to add integration tests for snapshots, I need to create a snapshot repository which would require setting [path.repo](https://opensearch.org/docs/latest/tuning-your-cluster/availability-and-recovery/snapshots/snapshot-restore/#shared-file-system) in Opensearch and I am not sure how this is done in the github workflows. If you can help me enabling this in the CI I would add integration tests for:
- snapshotRepo create/delete
- Snapshot create,delete,restore,clone

dblock:
Thanks @Jakob3xD for hanging in here with me. FYI, I am not trying to be difficult, I just want to make sure we are doing the right thing.

1. There's one important reason: you're introducing a breaking change. Every breaking change per semver requires a major version increment, that's a promise we have made - someone may be using the wrong type today and we will break them with this change. Or am I incorrect and the API doesn't work at all today?
2. I would be ok with _any_ test that exercises this code. Something that mocks the server response is acceptable. But integration tests would be better. If you can't figure out how to do it I can try looking into it or maybe @VachaShah knows too.

Jakob3xD:
I understand your point. It wouldn´t  make sens to do a mayor release for this. Maybe we keep this PR back until the next mayor release?

As sad I am happy to write the tests. The only thing I need help with is setting the [path.repo](https://opensearch.org/docs/latest/tuning-your-cluster/availability-and-recovery/snapshots/snapshot-restore/#shared-file-system) in the Opensearch config so I can create the snapshot repository.

dblock:
> I understand your point. It wouldn´t make sens to do a mayor release for this. Maybe we keep this PR back until the next mayor release?

Sure, but what's the reason not to add a backwards compatible version, and marking the wrong interface deprecated?

> As sad I am happy to write the tests. The only thing I need help with is setting the [path.repo](https://opensearch.org/docs/latest/tuning-your-cluster/availability-and-recovery/snapshots/snapshot-restore/#shared-file-system) in the Opensearch config so I can create the snapshot repository.

I am not sure how this works, give something a try and we'll help you when you get blocked?

wbeckler:
Today, people might be sending in a comma-joined string like `"snapshot1,snapshot2,snapshot3"` as the parameter, since the function is just comma-joining the array. 

Maybe we create a v3 branch of the client today and add this there, while marking the current type declaration as deprecated in the v2 client.

I kind of agree with @Jakob3xD that adding an alternative function with the new type is a bit of overhead. We would carry around `SnapshotsDeleteRequest` until it was deprecated, forcing a change on devs who would be scratching their heads about the repeat declaration. I think it's a better dev experience to let people use the comma-join workaround and add arrays to v3.

VachaShah:
Hi @Jakob3xD sorry for the delay! As @dblock mentioned, atleast unit testing would be good to add where it mocks the server. For integration tests, how about adding the path.repo to the [docker file](https://github.com/opensearch-project/opensearch-go/blob/main/.ci/opensearch/docker-compose.yml) which spins up OpenSearch cluster in the CI?

Jakob3xD:
@VachaShah I added some integration tests for Snapshot functions and added the setting to the docker file.
Works when testing locally.

dblock:
Assuming the new integration tests work, I think we should make this as a breaking change and take #246 along the way and major increment the client version.

dblock:
I think we should introduce an UPGRADING.md in this project similar to https://github.com/opensearch-project/opensearch-py/blob/main/UPGRADING.md that will spell breaking changes. Want to add one as part of this PR @Jakob3xD?

Jakob3xD:
The integration test with unrelased Opensearch fails, as the path setting is missing. If i see it correctly it builds opensearch from source and starts without any given settings. The normal Integration tests with docker works fine. Should I try adding the setting to the workflow https://github.com/opensearch-project/opensearch-go/blob/main/.github/workflows/test-integration-unreleased.yml#L52 ?

@dblock Yes, I will add the file.

dblock:
> The integration test with unrelased Opensearch fails, as the path setting is missing. If i see it correctly it builds opensearch from source and starts without any given settings. The normal Integration tests with docker works fine. Should I try adding the setting to the workflow https://github.com/opensearch-project/opensearch-go/blob/main/.github/workflows/test-integration-unreleased.yml#L52 ?
> 
> @dblock Yes, I will add the file.

Yes, do whatever you need to do to get a green build!

UPGRADING.md
```diff
@@ -0,0 +1,10 @@
+- [Upgrading Opensearch GO Client](#upgrading-opensearch-go-client)
+    - [Upgraading to >= 3.0.0](#upgrading-to->=-3.0.0)
+        - [opensearchapi](#opensearchapi-snapshot-delete)
+
+# Upgrading Opensearch GO Client
+
+## Upgraading to >= 3.0.0
+
+### opensearchapi snapshot delete
+SnapshotDeleteRequest and SnapshotDelete changed the argument `Snapshot` type from `string` to `[]string`.
```
dblock, line 10: Maybe give a before/after example. I know it sounds like overkill but it's best to help users as much as we can.

dblock (commented):
I think we need to also add 3.0 to https://github.com/opensearch-project/opensearch-go/blob/main/COMPATIBILITY.md.

dblock (dismissed):
I am good with this change!

@wbeckler This is the first time we're diverging a client version per semver with the server. Any concerns?

UPGRADING.md
Jakob3xD: Updated the file.

dblock (dismissed):

dblock (approved):

dblock:
@Jakob3xD looks like integration tests are failing :( care to take a look please?

Issue #220, opened 2023-01-23
[BUG] Unclear how to get a detailed error


dblock:
### What is the bug?

Go noob here. Following USER_GUIDE to implement a basic demo in https://github.com/dblock/opensearch-go-client-demo. How do I get the detailed error from the client?

```
$ go run main.go 
2023/01/23 11:21:29 indices.create: cannot retrieve information from OpenSearch
exit status 1
```

Jakob3xD:
I am not a maintainer but maybe this helps you.
```
func main() {
	resp, err := client.Indices.Create(index_name, client.Indices.Create.WithWaitForActiveShards("1"))
	if err != nil {
		log.Fatal("indices.create: ", err)
	}
	defer resp.Body.Close() // Avoid Memory leak
	if resp.IsError() { // Check the StatusCode
		return OpensearchRespError(resp)
	}
	log.Print(resp)
}

func OpensearchRespError(resp *opensearchapi.Response) error {
	body, err := io.ReadAll(resp.Body) // Read the Reponse Body
	if err != nil {
		return err
	}
	return fmt.Errorf("status: %v, msg: %s", resp.StatusCode, string(body)) // Return the StatusCode + Message/Body as Error
}
```
https://github.com/opensearch-project/opensearch-go/blob/v2.2.0/opensearchapi/opensearchapi.response.go#L100-L102

dblock:
I didn't realize I was supposed to _also_ check `resp.IsError()` (!).

Am I doing it right https://github.com/dblock/opensearch-go-client-demo/commit/5698c5530a12835a536bf325eefad06e41141207?

Is there really no way to avoid all this error checking boilerplate?

Jakob3xD:
In your first 4 change lines you return resp. For the others you return `err` what should be nil, so you would need to log `resp` instead. Moreover it would make sens to do this in functions to avoid code duplication.

Personally I prefer creating the requests and executing it in an extra function that handles the errors.
I have a function that looks like this:
```
func main {
	req := &opensearchapi.IndicesCreateRequest{
		Index:               index_name,
		WaitForActiveShards: "1",
	}
	if err := OpensearchDo(ctx, osClient, req); err != nil {
		fmt.Sprintf("%s", err)
		os.Exit(1)
	}

}
func OpensearchDo(ctx context.Context, osClient *opensearch.Client, req opensearchapi.Request) error {
	resp, err := req.Do(ctx, osClient)
	if err != nil {
		return err
	}
[...]

PR #246, opened 2023-03-03
Draft: Add Err() function to Response for detailed errors

Jakob3xD:
### Description
Add Err() function to Response and a new Error Type for a better and more detailed error handling. 

### Issues Resolved
Related https://github.com/opensearch-project/opensearch-go/issues/220


opensearchapi/opensearchapi.response.go
```diff
@@ -96,19 +117,41 @@ func (r *Response) Status() string {
...
 
 // IsError returns true when the response status indicates failure.
-//
 func (r *Response) IsError() bool {
 	return r.StatusCode > 299
 }
 
+// Err returns an error when the response status indicates failures.
+func (r *Response) Err() error {
+	if r.StatusCode > 299 {
```
dblock, line 126: Should we reuse `IsError()` here, and add explicit tests for it?

dblock (commented):
I like this a lot! Let's add something to USER_GUIDE too?

opensearchapi/opensearchapi.response.go
Jakob3xD: Using IsError() now and also update USER_GUID

opensearchapi/opensearchapi_response_internal_test.go
```diff
@@ -103,6 +105,50 @@ func TestAPIResponse(t *testing.T) {
...
+		}
+
+		err := res.Err()
+		if err == nil {
+			t.Errorf("Expected error for response: %s", res.Status())
+		}
+		var errTest *Error
+		if !errors.As(err, &errTest) {
+			t.Errorf("Expected error to be of type opensearchapi.Error: %T", err)
+		}
```
dblock, line 149: Ensure that `.Reason` and such are correct?

USER_GUIDE.md
```diff
@@ -300,8 +311,11 @@ func main() {
...
 	deleteIndexResponse, err := deleteIndex.Do(context.Background(), client)
 	if err != nil {
-		log.Println("failed to delete index ", err)
-		log.Fatal("delete index response body read err", err)
+		log.Println(fmt.Sprintf("failed to delete index: %s", err))
+		log.Fatal(fmt.Sprintf("delete index response body read err: %s", err))
+	}
+	if err := deleteIndexResponse.Err(); err != nil {
+		fmt.Println(fmt.Sprintf("failed to delete index: %s", err))
 	}
```
dblock, line 319: In the guide I want to see how to get error reason, index uuid, etc.

opensearchapi/opensearchapi_response_internal_test.go
Jakob3xD: Added an `if` to verify the content of the custom error.

USER_GUIDE.md
```diff
@@ -104,9 +110,12 @@ func main() {
 
 	searchResponse, err := search.Do(context.Background(), client)
 	if err != nil {
-		fmt.Println("failed to search document ", err)
+		fmt.Println(fmt.Sprintf("failed to search document: %s", err))
 		os.Exit(1)
 	}
+	if err := searchResponse.Err(); err != nil {
+		fmt.Println(fmt.Sprintf("failed to search document: %s", err))
+	}
```
dblock, line 118: I think this is net worse than what we had - we now have to check for error twice? Shouldn't `err` the first time be what we put into `searchResponse.Err()`?

dblock (commented):
I think we're close, but something feels wrong about having to now check for error twice. WDYT?

USER_GUIDE.md
dblock: That's what I am saying, shouldn't we make `.Do` care about the actual response?

As a user I want to write this:

```go
createIndexResponse, err := createIndex.Do(context.Background(), client)
if err != nil {
  // err here is either a network error, or the error from the response
  fmt.Println(fmt.Sprintf("failed to create index: %s", err))
  os.Exit(1)
}
```

WDYT?

Jakob3xD: Depends on what the lib maintainers wants.
Pro:
- The lib has a standard way on handling response errors, which prevent other users from doing it by them self
- You only need to check one error
- Adds the possibility to have a custom error type for some requests

Cons:
- Existing users with their own implementation needs to check the returned error. For example if the status code 404 is accepted. Currently this would be an error. Otherwise the question turns up, if the accepted response code should be a config setting to each Do request?

Jakob3xD: https://github.com/opensearch-project/opensearch-go/compare/main...Jakob3xD:opensearch-go:opensearchapi-response-error

Still missing some stuff but this is how it could look like.

dblock: I like the pros and I like the implementation. The fact that it's easier for beginners is huge.

Is this (the con) a breaking change for existing users? If so we'll need to major increment the version of the client.

opensearchapi/opensearchapi.error.go
```diff
@@ -0,0 +1,57 @@
+// SPDX-License-Identifier: Apache-2.0
+//
+// The OpenSearch Contributors require contributions made to
+// this file be licensed under the Apache-2.0 license or a
+// compatible open source license.
+//
+// Modifications Copyright OpenSearch Contributors. See
+// GitHub history for details.
+
+// Licensed to Elasticsearch B.V. under one or more contributor
```
dblock, line 10: You can remove the ES grant if it's all new code.

dblock (approved):
This is so much better!

Let's add an UPGRADING.md like in https://github.com/opensearch-project/opensearch-go/pull/237?

opensearchapi/opensearchapi.error.go
Jakob3xD: Done

dblock (dismissed):
@Jakob3xD Thank you so much for hanging in here with me with lots of my asks! I really appreciate your time and effort.

In https://github.com/opensearch-project/opensearch-net/issues/175#issuecomment-1475383403 yesterday (.NET client) I learned that it has a `ThrowExceptions` method, and that there's such a thing as "go style error handling" ;) @Xtansia is changing this here a bad idea? Should we also implement a `ThrowExceptions` option like in the .NET client in the GO client, and make this behavior optional (and thus backward compatible)? (AFAIK I dislike that option a lot because I feel like the client should just do the right thing - why would I ever want to check for errors in 2 different places? but I could be wrong).

@VachaShah care to CR, please?

Xtansia:
> In [opensearch-project/opensearch-net#175 (comment)](https://github.com/opensearch-project/opensearch-net/issues/175#issuecomment-1475383403) yesterday (.NET client) I learned that it has a `ThrowExceptions` method, and that there's such a thing as "go style error handling" ;) @Xtansia is changing this here a bad idea? Should we also implement a `ThrowExceptions` option like in the .NET client in the GO client, and make this behavior optional (and thus backward compatible)? (AFAIK I dislike that option a lot because I feel like the client should just do the right thing - why would I ever want to check for errors in 2 different places? but I could be wrong).

Calling what the .NET client does by default "Go style handling" is really a bit of a misnomer as it's not the same as idiomatic Go error handling, but it's how it's described to in the .NET client docs for `ThrowExceptions()`. Really it's just the difference between exceptions and manually checked errors. 

It is somewhat tricky in the fact we have multiple classes of errors with differing severity or semantics:
- Network error / DNS error / connection timeout - We couldn't contact the server at all
- SSL/TLS error - We tried communicating with something, but it isn't something we trust
- HTTP server error - We successfully communicated, but the actual server application is misbehaving
- HTTP client error - We successfully communicated, but what we sent it was junk
- HTTP auth error - We successfully communicated, but we're not allowed to do what we asked
- JSON deserializer error - We successfully communicated, but either the server sent us junk or our mapping for the shape is wrong
- OpenSearch application level errors:
  - Request we sent was of the correct shape, but the content was an invalid configuration (e.g. bad value, missing value)
  - 404 Not Found etc. e.g. trying to search an index that doesn't exist

I'm not super familiar with what's idiomatic for comparable clients/SDKs written in Go so I can't comment too directly on what's the right call here. The argument in exception based languages like C#, and presumably the reasoning for the .NET clients design choice, is usually that things like a network error or JSON deserializer error are "exceptional" errors (ie. we don't generally expect them to happen, given a reliable network & retries, and a correct mapping) however something like a 404 is not unexpected in the general case (ie. a race between indexing and reading a document).

Jakob3xD:
@dblock 
Golang does not have things like raise and except. Instead it has panic and recover but they should only be used on unrecoverable failures. IMO as a user I would not expect such a lib to panic. Instead I would always assume to get an error which I can check and handle. `ThrowExceptions` is not the normal golang way.

dblock (commented):
I am making the executive decision to merge this. As a user I don't want to check for 2 errors, one is enough and this accomplishes that very well. Thanks for contributing @Jakob3xD!

dblock (approved):

dblock:
@Jakob3xD I tried sending you an e-mail about something, maybe it went to spam? :) mind dropping me a note to dblock[at]dblock[dot]org or dblock[at]amazon[dot]com? Thaaanks!

Jakob3xD:
Closes https://github.com/opensearch-project/opensearch-go/issues/220

dblock:
Might just be an older client version @wbeckler.

dblock:
@Xe ranted about the ES client in https://xeiaso.net/blog/elasticsearch - I think we fixed the error handling part fully in this PR, @Jakob3xD care to take a look at the blog post? anything I've missed?

Jakob3xD:
I wouldn't say we cover it fully. In Go it is common to have  pre-defined error types that you can match on. So for example you have the error type ErrIndexNotFound which the user can then use to match against his error. But as this is hard to maintain our implementation is more generic but at least existing 😅. We do the json to go part for the user and return one error so you don't need to check the response for errors your self.

Beside that I can understand his anger. That's why I want to rework the lib to make it more idiomatic.

dblock:
> Beside that I can understand his anger. That's why I want to rework the lib to make it more idiomatic.

Please!

MaxKsyunz:
Was this PR reverted? Not seeing it in main or the 2.3 release.

dblock:
Yeah, but I don't know what happened here, @VachaShah?

VachaShah:
> Yeah, but I don't know what happened here, @VachaShah?

Thats weird, I don't see this commit but no revert happened for this as well. Taking a deeper look.

VachaShah:
Was this supposed to be a breaking change?

dblock:
https://github.com/opensearch-project/opensearch-go/commit/2f58bf6cb3afa0dfa62a1349924b6e0ebe01a4da
Yes, looks like this was a breaking change.

dblock:
> Was this PR reverted? Not seeing it in main or the 2.3 release.

It's on main though, but was reverted from 2.3 as breaking?

VachaShah:
Yeah I reverted before 2.3 release since this was a breaking change to do the next 2.x release.

PR #252, opened 2023-03-20
Update signer v1 example for AOSS

harrisonhjones:
### Description

Like https://github.com/opensearch-project/opensearch-go/pull/216 but for the v1 signer.

### Issues Resolved

I did not open one but could if needed. Originally tried solving it via https://github.com/aws/aws-sdk-go/pull/4761 but this way is more "right".

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - Bumps `github.com/aws/aws-sdk-go` from 1.44.180 to 1.44.195
 - Bumps `github.com/aws/aws-sdk-go` from 1.44.176 to 1.44.185
 - Bumps `github.com/aws/aws-sdk-go` from 1.44.132 to 1.44.180
+- Bumps `github.com/stretchr/testify` from 1.8.1 to 1.8.2
 
 ### Added
 - Github workflow for changelog verification ([#172](https://github.com/opensearch-project/opensearch-go/pull/172))
--- a/go.mod
+++ b/go.mod
@@ -6,6 +6,6 @@ require (
 	github.com/aws/aws-sdk-go v1.44.210
 	github.com/aws/aws-sdk-go-v2 v1.17.5
 	github.com/aws/aws-sdk-go-v2/config v1.18.15
-	github.com/stretchr/testify v1.8.1
+	github.com/stretchr/testify v1.8.2
 	golang.org/x/net v0.7.0 // indirect
 )
--- a/go.sum
+++ b/go.sum
@@ -41,8 +41,8 @@ github.com/stretchr/objx v0.4.0/go.mod h1:YvHI0jy2hoMjB+UWwv71VJQ9isScKT/TqJzVSS
 github.com/stretchr/objx v0.5.0/go.mod h1:Yh+to48EsGEfYuaHDzXPcE3xhTkx73EhmCGUpEOglKo=
 github.com/stretchr/testify v1.7.1/go.mod h1:6Fq8oRcR53rry900zMqJjRRixrwX3KX962/h/Wwjteg=
 github.com/stretchr/testify v1.8.0/go.mod h1:yNjHg4UonilssWZ8iaSj1OCr/vHnekPRkoO+kdMU+MU=
-github.com/stretchr/testify v1.8.1 h1:w7B6lhMri9wdJUVmEZPGGhZzrYTPvgJArz7wNPgYKsk=
-github.com/stretchr/testify v1.8.1/go.mod h1:w2LPCIKwWwSfY2zedu0+kehJoqGctiVI29o6fzry7u4=
+github.com/stretchr/testify v1.8.2 h1:+h33VjcLVPDHtOdpUCuF+7gSuG3yGIftsP1YvFihtJ8=
+github.com/stretchr/testify v1.8.2/go.mod h1:w2LPCIKwWwSfY2zedu0+kehJoqGctiVI29o6fzry7u4=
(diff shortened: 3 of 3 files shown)


signer/aws/aws.go
```diff
@@ -25,51 +27,74 @@ import (
...
+const emptyBodySHA256 = "2cf24dba5fb0a30e26e83b2ac5b9e29e1b161e5c1fa7425e73043362938b9824"
+
+// Signer is an interface that will implement opensearchtransport.Signer
 type Signer struct {
 	session session.Session
 	service string
 }
 
-// NewSigner returns an instance of Signer for AWS OpenSearchService
+// NewSigner returns an instance of Signer for configured for AWS OpenSearchService.
```
dblock, line 41: The correct names would be "Amazon OpenSearch Service" and "Amazon OpenSearch Serverless", maybe we can update?

signer/aws/aws.go
```diff
@@ -25,51 +27,74 @@ import (
...
 		return err
 	}
-	_, err = signer.Sign(req, bytes.NewReader(buf), serviceName, *region, time.Now().UTC())
-	return
+
+	return nil
+}
+
+func hexEncodedSha256(b []byte) string {
+	hasher := sha256.New()
```
dblock, line 94: This can never fail?

signer/aws/aws.go
```diff
@@ -25,51 +27,74 @@ import (
 // OpenSearchService AWS OpenSearchService Name
 const OpenSearchService = "es"
 
-// Signer is a interface that will implement opensearchtransport.Signer
+// OpenSearchServiceServerless AWS OpenSearchService Serverless Name
+const OpenSearchServiceServerless = "aoss"
+
+const emptyBodySHA256 = "2cf24dba5fb0a30e26e83b2ac5b9e29e1b161e5c1fa7425e73043362938b9824"
```
dblock, line 33: We could avoid declaring a constant for this and ensuring the sign function works for a `null` body (and produces this value). Probably shorter, but I don't feel too strongly about it.

signer/aws/aws.go
```diff
@@ -25,51 +27,74 @@ import (
...
-// NewSignerWithService returns an instance of Signer for given service
+// NewSignerWithService returns an instance of Signer for a given service.
 func NewSignerWithService(opts session.Options, service string) (*Signer, error) {
 	if len(strings.TrimSpace(service)) < 1 {
 		return nil, errors.New("service cannot be empty")
 	}
 	awsSession, err := session.NewSessionWithOptions(opts)
 	if err != nil {
-		return nil, fmt.Errorf("failed to get session from given option %v due to %s", opts, err)
+		return nil, fmt.Errorf("failed to get session from given options %v: %v", opts, err)
```
dblock, line 54: Should this be `%v: %s`?

dblock (changes requested):
%v: %v looks like a typo/bug

signer/aws/aws.go
harrisonhjones: It looks like hasher.Write can. Will address.

signer/aws/aws.go
harrisonhjones: no. The common go style for error decorating is: `fmt.Errorf("failed to foo: %v", err)`. I was updating this to match here.

signer/aws/aws.go
harrisonhjones: Will address

signer/aws/aws.go
harrisonhjones: I did that originally but figured this actually saved us some unnecessary calculation.

harrisonhjones (commented):
Thanks for taking a look @dblock . I'll address you comments asap.

harrisonhjones:
@dblock addressed your concerns.

signer/aws/aws.go
```diff
@@ -22,54 +24,85 @@ import (
 	v4 "github.com/aws/aws-sdk-go/aws/signer/v4"
 )
 
-// OpenSearchService AWS OpenSearchService Name
+// OpenSearchService Amazon OpenSearch Service Name
```
VachaShah, line 27: You could keep it as "Amazon OpenSearch Service" and for L30 as "Amazon OpenSearch Serverless".

signer/aws/aws.go
```diff
@@ -22,54 +24,85 @@ import (
...
 const OpenSearchService = "es"
 
-// Signer is a interface that will implement opensearchtransport.Signer
+// OpenSearchServerless Amazon OpenSearch Serverless Name
+const OpenSearchServerless = "aoss"
+
+const emptyBodySHA256 = "2cf24dba5fb0a30e26e83b2ac5b9e29e1b161e5c1fa7425e73043362938b9824"
```
VachaShah, line 33: You can skip the empty body hash since the signer assigns the empty body hash if the request body is null, looking at https://github.com/aws/aws-sdk-go/blob/main/aws/signer/v4/v4.go#L708

VachaShah (commented):
Thank you @harrisonhjones for adding this!

It would be good to add this to the user guide as well: https://github.com/opensearch-project/opensearch-go/blob/main/USER_GUIDE.md#aws-sdk-v1

dblock (approved):
I'm good with this! @VachaShah has a good point that the signing code already signs an empty body, so let's avoid doing it twice? 

+1 on users guide and CHANGELOG please?

signer/aws/aws.go
harrisonhjones: As far as I can tell that's only if `includeSHA256Header ` is true which isn't for `aoss`, only some s3 services.

signer/aws/aws.go
harrisonhjones: Also:

1. Do we _want_ to export these constants? They aren't exported in the v2 signer.
2. If we do perhaps they should be `OpenSearchServiceName` and `OpenSearchServerlessName` to clarify they are `Name`s.

PR #253, opened 2023-03-20
Add point in time api

### LINKED ISSUE #251: [FEATURE] Add point_in_time API
Currently the lib does not support the `point_in_time api` used for pagination searching.
https://opensearch.org/docs/latest/search-plugins/point-in-time-api/

Implement the new API Endpoints described in the docs.

Jakob3xD:
### Description
Add  the Point in Time API.
https://opensearch.org/docs/latest/search-plugins/point-in-time-api/

### Issues Resolved
Closes #251 

### DIFF
--- a/opensearchapi/api.pointintime.create.go
+++ b/opensearchapi/api.pointintime.create.go
@@ -46,10 +46,10 @@ func newPointInTimeCreateFunc(t Transport) PointInTimeCreate {
 
 // ----- API Definition -------------------------------------------------------
 
-// Point In Time ets you run different queries against a dataset that is fixed in time.
+// PointInTimeCreate let you create a pit for searching with pagination
 type PointInTimeCreate func(o ...func(*PointInTimeCreateRequest)) (*Response, *PointInTimeCreateResp, error)
 
-// PointInTimeRequest configures the Point In Time API request.
+// PointInTimeCreateRequest configures the Point In Time Create API request.
 type PointInTimeCreateRequest struct {
 	Index []string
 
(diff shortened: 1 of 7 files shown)


Jakob3xD:
@dblock does this lib aim to be a low level lib or should it be more user friendly?
I am asking because I added some custom Response types for Point In Time requests and also avoid the user to build the delete body.

dblock:
> @dblock does this lib aim to be a low level lib or should it be more user friendly? I am asking because I added some custom Response types for Point In Time requests and also avoid the user to build the delete body.

I definitely think we want to minimize the amount of work a user has to do. So as much user friendly as possible, but also letting users go deep into low level constructs.

opensearchapi/api.pointintime.create.go
```diff
@@ -0,0 +1,260 @@
+// SPDX-License-Identifier: Apache-2.0
+//
+// The OpenSearch Contributors require contributions made to
+// this file be licensed under the Apache-2.0 license or a
+// compatible open source license.
+//
+// Modifications Copyright OpenSearch Contributors. See
+// GitHub history for details.
+
+// Licensed to Elasticsearch B.V. under one or more contributor
```
dblock, line 10: This is new code? We don't need a ES grant, remove this license part below.

dblock (approved):
Small nit.

opensearchapi/api.pointintime.create.go
Jakob3xD: Question for me as I am not into such license things.
Can I remove the ES grant if it is new code but follows the same scheme and has similarity to code that has ES grant? As the `Withxyz()` functions are all the same.

dblock: IANAL, but if you _copied_ any code with such license then we should probably keep it. But if it's _new code_ then we don't need it. Check out https://github.com/opensearch-project/.github/blob/main/CONTRIBUTING.md#license-headers for more info.

Issue #260, opened 2023-03-22
[FEATURE] Add a code/markdown linter


dblock:
### Is your feature request related to a problem?

Add a linter that formats code, markdown, etc.

### What solution would you like?

- A linter that auto-formats code.
- CI that fails if the linter didn't run and offers how to run it.

zethuman:
@dblock can we close this issue?

dblock:
Yes, done in #261, thanks! 

Btw, you can say "Closes #261" in a PR and it will auto-close the issue when the PR is merged.

PR #258, opened 2023-03-21
Format with prettier

harrisonhjones:
### Description

Just a simple "let's format the markdown files" PR. Ran `npx prettier --prose-wrap "always" -w *.md` and committed the results.

### Issues Resolved

None

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #261, opened 2023-03-23
Add markdown linter

zethuman:
### Description
_Add a linter that formats markdown files with prettier_

### Issues Resolved
_Closes [#260]._ 


zethuman:
@dblock can we open settings > actions  > ceneral

dblock:
- please sign DCO
- let's not wrap markdown? https://code.dblock.org/2021/06/07/to-wrap-or-not-to-wrap-in-markdown.html

.github/workflows/lint.yml
```diff
@@ -27,3 +27,25 @@ jobs:
         shell: bash
       - run: go version
       - run: make lint
+
+  prettify:
+    name: Prettify
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
```
dblock, line 35: Why is this needed? The CI runs on the pushed code or PR already AFAIK.

.github/workflows/lint.yml
```diff
@@ -27,3 +27,25 @@ jobs:
...
+        with:
+          fetch-depth: 0
+          ref: ${{ github.event.pull_request.head.ref }}
+      - uses: actions/checkout@v3
+        if: github.event_name == 'push'
+        with:
+          fetch-depth: 0
+      - name: Install npm v8
+        run: npm i -g npm@8
+      - name: Prettify markdown code
```
dblock, line 46: Will this fail if there are code changes? That's the intent of CI (ensure code is prettified).

dblock (changes requested):
Let's add to the developer guide that one is supposed to run a markdown linter before committing?

.github/workflows/lint.yml
zethuman: You mean why used checkout action? 
Initially, I thought to just amend the commit and not create a new one, so I pulled out all the brunches and stories. But with a pull request, there was no reason to pull everything, so the checkout is only for the head. 
So at the push event, I left it as it was with the job on the linter above, if you look, there is also a checkout

.github/workflows/lint.yml
zethuman: Nope, this linter will not show the difference, but immediately starts changing specified files according to the standard. Of course, you can also add a job simply with a `--check` flag, so that it checks for an error if it indicates the difference. But when I read the description of the issue it was about auto-formatting the code in the pipeline, if I'm wrong, please correct me

zethuman:
Add makefile script to check for format and updated developer guide

.github/workflows/lint.yml
dblock: I mean why is `if: github.event_name == 'pull_request'` vs. `if: github.event_name == 'push'` needed. If you combine the two you will get the "current code" (on a PR it will include the PRed code, on main it will include HEAD).

.github/workflows/lint.yml
dblock: I think the developer should be making changes and ensuring code is linted (you should make it easy).
CI is here to enforce it, so this should only run a check.

Makefile
```diff
@@ -66,6 +66,42 @@ lint:  ## Run lint on the package
 		cd "internal/build/" && go mod tidy && go mod download && go vet ./...; \
 	}
 
+package := "prettier"
```
dblock, line 69: Is there really no linter written in go so that we don't need to require `npm`?

dblock (changes requested):

.github/workflows/lint.yml
zethuman: Origin issue https://github.com/opensearch-project/opensearch-go/issues/260 says that there must be "A linter that auto-formats code."
Your comments confused me((
Do you need a linter that just checks and shows the user where something is wrong, or will it check and auto-format the code?

Makefile
zethuman: There probably exists such linter, but I came from another issue #258, since there was no questions of the tool used, I decided to use it to standardize. If you need the instrument to be on go, of course I can take the trouble to tune it ...

.github/workflows/lint.yml
zethuman: Yes, I think I understand your train of thought. Well, I will commit without separation, with a depth of 1

```
- uses: actions/checkout@v3
  with: { fetch-depth: 1 }
```

.github/workflows/lint.yml
dblock: 1. Developers need an explanation of how to run the linter when making code changes. This can be done command-line/integrated into the IDE, etc, instructions should be added to DEVELOPER_GUIIDE. 
2. CI needs to verify that the code is formatted according to a standard. Thus when one makes pull requests CI runs the linter checks and fails if conventions aren't followed.

I don't know what the Go standard linter is, but let's use it. For Ruby it's RuboCop, https://github.com/rubocop/rubocop, for example.

Makefile
zethuman: I searched up everywhere in the web, there is no such package written in go, that can lint markdown files. Even if such exists, it is no longer maintained. 

Forced to admit ChatGPT to admit that there is no such linter after a long conversation...

```
I apologize for my previous responses. You are correct that there is currently no package written in Go that can lint Markdown files.

While there are some Markdown linters that integrate with Go, such as mdl, they are not written in Go themselves.

Again, I apologize for any confusion I may have caused and I hope this clears up any confusion.
```

dblock: ok, use the one we have

please see other comments/asks

zethuman: Summary of our conversation:

* Fetch source code with fetch-depth - 1
* Changed CI linter option to check only and fail if files are invalid
* Add script to Makefile to prettify markdown files

Makefile script checks format of markdown files, in case of invalid files, asks user to decide auto format or leave as is

dblock (approved):
Is there an option to prettier to format tables? Many editors auto-format those these days.

dblock:
Yeah it's NBD, if you look at the diff here you will see that prettier got rid of spaces, so it's harder to read in an editor.

zethuman:
I opened an issue on the developer side of the tool, we will figure it out

Issue #279, opened 2023-04-04
Add API Guides


nhtruong:
(Spawn off of meta Issue: https://github.com/opensearch-project/opensearch-clients/issues/56)
To help drive adoption of OpenSearch, and improve user's experience, we're working on adding more documentation for all clients. One of the strategies mentioned in said issue is writing a collection of articles serving as a User's Guide for each client. This issue will serve as a hub for all issues related to this topic.

Each Guide/Article will be a `good-first-issue` for new contributors to work on. There's already a Ruby version of each article that the articles on this repo should be based upon. Most of the time, we only need to replace the Ruby code blocks with the client code of this repo. Please refer to the [meta issue](https://github.com/opensearch-project/opensearch-clients/issues/56) for more details.

- [x] https://github.com/opensearch-project/opensearch-go/issues/273
- [x] https://github.com/opensearch-project/opensearch-go/issues/274
- [x] https://github.com/opensearch-project/opensearch-go/issues/275
- [x] https://github.com/opensearch-project/opensearch-go/issues/276
- [x] https://github.com/opensearch-project/opensearch-go/issues/277
- [x] https://github.com/opensearch-project/opensearch-go/issues/278

dblock:
@zethuman I linked user guides in https://github.com/opensearch-project/opensearch-go/pull/302

I noticed some inconsistencies in capitalization (e.g. `### Mix and match operations` vs `### Update an Index`) and language (e.g. `Get Metadata` vs. `Indexing ...`). Want to align those?

PR #292, opened 2023-04-10
[GUIDE] Add bulk guide

zethuman:
### Description
_Add bulk guide._

### Issues Resolved
_Closes [#278, #279]._ 


dblock:
You're missing DCO, same as other PRs.

zethuman:
fixed

guides/bulk.md
```diff
@@ -0,0 +1,204 @@
...
+
+import (
+    "github.com/opensearch-project/opensearch-go/v2"
+    "log"
+)
+
+func main() {
+    client, err := opensearch.NewDefaultClient()
+    if err != nil {
+        log.Printf("error occurred: [%s]", err.Error())
```
dblock, line 20: Should we bailing here in samples, e.g. `log.Fatalf`?

dblock (approved):

guides/bulk.md
zethuman: I was guided by the rules from [issue](https://github.com/opensearch-project/opensearch-go/issues/278#issue-1654505429), there was written not to throw errors, or did I misunderstand?

dblock: Maybe, I don't know what's best. 

What's the user experience when `NewDefaultClient()` returns an error with the code as written?

zethuman: In the console there will be a reason why the client was not created and the program will complete successfully, I mean there will be no fatal termination. The user, having read the reason, understands that he did something wrong, and decides the origin reason and tries again.

dblock: In this example the code doesn't do anything else, however in the example below:

```go
movies := "movies"
books := "books"
createMovieIndex, err := client.Indices.Create(movies)
if err != nil {
    log.Printf("error occurred: [%s]", err.Error())
}
log.Printf("response: [%+v]", createMovieIndex)
createBooksIndex, err := client.Indices.Create(books)
if err != nil {
    log.Printf("error occurred: [%s]", err.Error())
}
log.Printf("response: [%+v]", createBooksIndex)
```

You get an error the first time on  `client.Indices.Create(movies)`, then you get an error again the second time on `client.Indices.Create(books). I believe we shouldn't be bailing after the first error, no?

zethuman: I agree about the first one, but again I will refer to the guide ... It's not difficult for me to change it for a fatal mistake. At the expense of the second, at the end of each guide there are cleaning topics, and the previously created modules were cleaned there

PR #300, opened 2023-04-14
Read, close and replace the http Reponse Body

Jakob3xD:
### Description
With this change the http reponse body gets automatically closed by us so the user don't have to do it and it prevents unclosed connections if the user forgot to close it. This wont brake any existing implementations as you can still call `.Close()` on the Body.

### Issues Resolved
https://pkg.go.dev/net/http#Client.Do
> If the returned error is nil, the Response will contain a non-nil Body which the user is expected to close. If the Body is not both read to EOF and closed, the Client's underlying RoundTripper (typically Transport) may not be able to re-use a persistent TCP connection to the server for a subsequent "keep-alive" request. 

### DIFF
--- a/opensearchtransport/opensearchtransport.go
+++ b/opensearchtransport/opensearchtransport.go
@@ -75,13 +75,11 @@ func init() {
 }
 
 // Interface defines the interface for HTTP client.
-//
 type Interface interface {
 	Perform(*http.Request) (*http.Response, error)
 }
 
 // Config represents the configuration of HTTP client.
-//
 type Config struct {
 	URLs     []*url.URL
 	Username string
@@ -113,7 +111,6 @@ type Config struct {
 }
 
 // Client represents the HTTP client.
-//
 type Client struct {
 	sync.Mutex
 
@@ -146,7 +143,6 @@ type Client struct {
 // New creates new transport client.
 //
 // http.DefaultTransport will be used if no transport is passed in the configuration.
-//
 func New(cfg Config) (*Client, error) {
 	if cfg.Transport == nil {
 		cfg.Transport = http.DefaultTransport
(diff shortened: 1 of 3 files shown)


opensearchtransport/opensearchtransport.go
```diff
@@ -407,14 +402,20 @@ func (c *Client) Perform(req *http.Request) (*http.Response, error) {
 			time.Sleep(c.retryBackoff(i + 1))
 		}
 	}
+	// Read, close and replace the http reponse body to close the connection
+	if res != nil && res.Body != nil {
+		body, err := io.ReadAll(res.Body)
+		res.Body.Close()
```
VachaShah, line 408: @Jakob3xD Just a quick question: since we are closing the response body here, would it cause errors for the user if they try to close it again in their application?

dblock:
Looks good, needs a unit test, including to what happens with a double close by the caller.

opensearchtransport/opensearchtransport.go
dblock: See above, "this wont brake any existing implementations as you can still call .Close() on the Body.". Still needs a unit test.

Jakob3xD: To explain it in addition. The NopeCluster returns a io.ReadCloser like the res.Body is and has a Close() func that just returns nil. https://cs.opensource.google/go/go/+/refs/tags/go1.20.3:src/io/io.go;l=678
So you can still call close, but it won't do anything as it is not needed.

Jakob3xD:
> Looks good, needs a unit test, including to what happens with a double close by the caller.

I am not 100% sure what kind of test you expect. Before this MR and after this MR the user can call Close() as often as they want. As the NopCloser just returns nil when calling Close it is obsolete to call it and to test it?

Jakob3xD:
@dblock I added a test. It will fail if the resp.Body is not a NopCloser. Without my added change it would fail with following error:
```
=== RUN   TestTransportBodyClose
    opensearchtransport_integration_test.go:145: Failed to read the reponse body: http: read on closed response body
--- FAIL: TestTransportBodyClose (0.00s)
```

dblock (approved):
Very good.

Issue #296, opened 2023-04-11
[BUG] Connection problems in actions 


zethuman:
### What is the bug?
_One test case always fails in PR with connectivity issue._

### How can one reproduce the bug?
_Just look at any PR, they are all fails within one test case_

### What is the expected behavior?
_Should be success._

### What is your host/environment?
_MacOS Ventura._

### Do you have any screenshots?
<img width="940" alt="image" src="https://user-images.githubusercontent.com/51375666/231030960-957d77e4-4cfe-40f7-8406-3c3f6f0eddef.png">

zethuman:
Of course, this error annoying me, so I take it to work

zethuman:
I think the problem is in the timeout, that is, the cluster does not have time to start when the requests are already arriving. , Need to set the timeout in the pipeline, since timeout in the makefile script does not work. If there are other suggestions, I can check

A real example of how tests behave if timeout is [increased](https://github.com/zethuman/opensearch-go/actions/runs/4706490474/jobs/8347756932) and if it is [decreased](https://github.com/zethuman/opensearch-go/actions/runs/4706477450)

zethuman:
I think it was a coincidence [#303], I'll keep digging

PR #303, opened 2023-04-15
[BUG] Increased timeout in actions between cluster start and test run

zethuman:
### Description
_Increased timeout in actions between cluster start and test run._

### Issues Resolved
_Closes [#296]._ 

### DIFF
--- a/.github/workflows/test-compatibility.yml
+++ b/.github/workflows/test-compatibility.yml
@@ -43,7 +43,9 @@ jobs:
 
       - name: Integration test without security
         if: ${{ matrix.secured == 'false'}}
-        run: make test-integ race=true
+        run: |
+          make test-integ race=true
+          sleep 20
 
       - name: Integration test with security
         if: ${{ matrix.secured == 'true'}}
--- a/.github/workflows/test-compatibility.yml
+++ b/.github/workflows/test-compatibility.yml
@@ -45,6 +45,7 @@ jobs:
         if: ${{ matrix.secured == 'false'}}
         run: |
           make test-integ race=true
+          sleep 20
 
       - name: Integration test with security
         if: ${{ matrix.secured == 'true'}}
--- a/.github/workflows/test-compatibility.yml
+++ b/.github/workflows/test-compatibility.yml
@@ -45,7 +45,6 @@ jobs:
         if: ${{ matrix.secured == 'false'}}
         run: |
           make test-integ race=true
-          sleep 20
 
       - name: Integration test with security
         if: ${{ matrix.secured == 'true'}}


.github/workflows/test-compatibility.yml
```diff
@@ -40,10 +40,12 @@ jobs:
           export OPENSEARCH_VERSION=${{ matrix.entry.opensearch_version }}
           export SECURE_INTEGRATION=${{ matrix.secured }}
           make cluster.clean cluster.build cluster.start
+#          sleep 20
```
dblock, line 43: Looks like you've commented this out instead of actually increasing timeout?

dblock (changes requested):
Can we find a way to wait for the cluster to start in an active loop with a much longer timeout?

Jakob3xD:
> Can we find a way to wait for the cluster to start in an active loop with a much longer timeout?

We could use the testMain function of test?
https://pkg.go.dev/testing#hdr-Main

.github/workflows/test-compatibility.yml
zethuman: Yes, I checked this case without timeout and there is success result. I think it was just a coincidence at the first time. So, it has nothing to do with timeout, I'm trying to dig further. But locally everything works fine for me with the version of the opensearch image the same as on the erroneous. So I'm at a loss as to why this is happening

zethuman: Yeap, I created a loop to test as you suggested. And there are two problems:
1. Status check does not work with a [secure connection (https)](https://github.com/zethuman/opensearch-go/actions/runs/4773852492/jobs/8487335276).
2. Also, there seems to be [some problem](https://github.com/zethuman/opensearch-go/actions/runs/4773852492/jobs/8487335428) with starting the server, as it stops responding to status requests and just closes the connection. 

I keep digging

zethuman:
As far as I understood after research, there is some kind of problem with the image itself or the server where it starts, but what is strange is this random phenomenon. Sometimes it works well, sometimes it doesn't. In general, we need to dig on the server side, why this is happening. However, as a client, we must not run tests until the server is ready. To do this, following the recommendations of the @dblock, I created a script to check the status of the server, if it is not successful, we restart the server. This solution is not the best because we only treat the symptoms and not the disease itself. But as a temporary solution, I think it will not hurt until the main one is solved. If you have any ideas on how to solve it differently, I'm waiting in the comments ...

[Link to the case](https://github.com/zethuman/opensearch-go/actions/runs/4779097970/jobs/8495973096) when the server is not responding and we are trying to restart the service, after we successfully run the tests.

dblock (approved):
Let's merge this because it works around our immediate problem.

@zethuman Every client and every plugin we have has the same problem. Do you know a way to replicate this behavior across all our (client) repos without copy-pasting the entire contents of `cluster.check`? Maybe we can extract it into a tool? Or use an existing tool like https://github.com/dnnrly/wait-for (first hit on google, didn't research more). I'd like this to be a 1-2 liner added everywhere.

zethuman:
@dblock I'm a little off the radar)) 

Yes, regarding the tool, it's a good idea. If other clients have the same problem, it confirms that we have a problem on the server. I suggest adding the `healthcheck` configurations to the docker-compose or to the dockerfile itself, and wait until the container state is `healthy`. If not, docker will restart the container itself until it is `healthy`. What do you think? For me, this is not a crutch and perhaps the best solution in symbiosis with what you sent.

dblock:
That's a good idea. Take it on!

zethuman:
What is the best way to do it, open a new PR or continue with this one?

dblock:
I had merged this PR @zethuman because it worked around our problem! Progress over perfection.

zethuman:
Noted. Thanks

PR #316, opened 2023-05-04
[Feature] Add healthcheck to docker container

### LINKED ISSUE #296: [BUG] Connection problems in actions

Jakob3xD:
### Description
- Add a healthcheck to the docker container that kills it self, forcing it to restart the container
- Add a bash for loop to wait for the cluster to be ready

### Issues Resolved
 Closes #296 

### DIFF
--- a/.ci/opensearch/Dockerfile.opensearch
+++ b/.ci/opensearch/Dockerfile.opensearch
@@ -3,5 +3,11 @@ FROM opensearchproject/opensearch:${OPENSEARCH_VERSION}
 
 ARG opensearch_path=/usr/share/opensearch
 ARG SECURE_INTEGRATION
+ENV SECURE_INTEGRATION=$SECURE_INTEGRATION
 
 RUN if [ "$SECURE_INTEGRATION" != "true" ] ; then $opensearch_path/bin/opensearch-plugin remove opensearch-security; fi
+
+HEALTHCHECK --start-period=20s --interval=30s \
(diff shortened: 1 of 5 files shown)


Jakob3xD:
@dblock this is how it could look like in general, so we don't need to import an action.
We could also remove the for loop that waits for the cluster to be ready by letting the test it self wait for the cluster to be ready. Like it is done with the TestMain functions in https://github.com/opensearch-project/opensearch-go/compare/main...Jakob3xD:opensearch-go:feature/actions/healthcheck

The advantage of the container killing it self, is that it does not need an external tool to restart it, if it is unhealthy, but we then need something to wait for the cluster to be ready. In this PR it is the bash for loop.

The external action from [zethuman](https://github.com/zethuman) handles the cluster restart and waits for the cluster to be ready in one task, which is also fine and solves the issue every lib might have.

Personally I don't have any hard feelings on how we do it, as long as it is inside the opensearch scope and not in any users projects, as we have no control over it.

zethuman:
Personally, I see the only difference is that my solution is in another repository, this one is included in the script. That is, they are identical in mechanism, I can only say that mine looks more elegant or something))

dblock:
I'm going to take this one for now to avoid the dependency. No hard feelings!

dblock (approved):

PR #338, opened 2023-06-22
Data stream stats: Add name only if present

### LINKED ISSUE #337: [BUG] Querying Data Streams Stats with AWS Auth without a name returns a signing error
### What is the bug?
When querying the Data Stream Stats API with AWS Authentication and without setting an index name the AWS Signature is invalid.

Error:
```
The request signature we calculated does not match the signature you provided. 
Check your AWS Secret Access Key and signing method.
```

Setup an AWS Signer and OpenSearch Client and make a request for `/_data_streams/_stats` without setting an index name.
```go
awsCfg, err := config.LoadDefaultConfig(context.Background())
if err != nil {
    return err
}

awsSigner, err := awsv2.NewSigner(awsCfg)
if err != nil {
    return err
}

url := "https://my-opensearch-cluster"
osCfg := opensearch.Config{
    Addresses: []string{url},
    Signer:    awsSigner,
}

client, err := opensearch.NewClient(osCfg)
if err != nil {
    return err
}

resp, err := client.Indices.GetDataStreamStats()
if err != nil {
    return err
}
```

The list of data streams and their stats is returned. The OpenSearch API supports making this request with no name and should return all of the data streams instead.

Linux 6.2.6
Go 1.20.5

No

It looks like the GetDataStreams function always adds the name and thus ends up adding an additional `/` to the path. This `/` is dropped on the server side but does get factored into the signature calculation on the client side causing the mismatch.

DewaldV:
### Description
This makes the GetDataStreamStats request only add the `Name` and path separator (`/`) to the path if Name is not empty.

As it stands when using this function without setting a name the path will be:
```
/_data_stream//_stats
```

This path is then used by the AWS Signer to calculate a signature. When the request is sent the server recalculates the signature but the additional `/` has been dropped by now and we get a signature mismatch. See #337 for details.

I have tested this fix in our own codebase and it resolves the issue as described.

### Issues Resolved
Closes #337 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,6 +51,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 - Corrects curl logging to emit the correct URL destination ([#101](https://github.com/opensearch-project/opensearch-go/pull/101))
 - Corrects handling of errors without an error response body ([#286](https://github.com/opensearch-project/opensearch-go/pull/286))
+- Corrects AWSv4 signature on DataStream Stats with no index name specified ([#338](https://github.com/opensearch-project/opensearch-go/pull/338))
 
 ### Security
 
--- a/go.mod
+++ b/go.mod
@@ -6,6 +6,7 @@ require (
 	github.com/aws/aws-sdk-go v1.44.298
 	github.com/aws/aws-sdk-go-v2 v1.19.0
 	github.com/aws/aws-sdk-go-v2/config v1.18.28
+	github.com/kinbiko/jsonassert v1.1.1
 	github.com/stretchr/testify v1.8.4
 	golang.org/x/net v0.7.0 // indirect
 )
--- a/go.sum
+++ b/go.sum
@@ -33,6 +33,8 @@ github.com/jmespath/go-jmespath v0.4.0 h1:BEgLn5cpjn8UN1mAw4NjwDrS35OdebyEtFe+9Y
 github.com/jmespath/go-jmespath v0.4.0/go.mod h1:T8mJZnbsbmF+m6zOOFylbeCJqk5+pHWvzYPziyZiYoo=
 github.com/jmespath/go-jmespath/internal/testify v1.5.1 h1:shLQSRRSCCPj3f2gpwzGwWFoC7ycTf1rcQZHOlsJ6N8=
 github.com/jmespath/go-jmespath/internal/testify v1.5.1/go.mod h1:L3OGu8Wl2/fWfCI6z80xFu9LTZmf1ZRjMHUOPmWr69U=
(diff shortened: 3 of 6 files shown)


dblock (changes requested):
Looks good, care to add a test, please?

VachaShah (commented):
Thank you @DewaldV! +1 for adding a test for this.

DewaldV:
Thanks for taking a look, I'll add a test to this as requested. :+1: 

Apologies for my delayed response I've been at conference this week.

DewaldV:
@VachaShah @dblock apologies for the delay in getting those tests sorted but I've added them as requested.

The integration test is a bit strange since I had to add additional Data Stream create and delete tests so that I could ensure I was actually getting multiple results from the `/_data_stream/_stats` endpoint as expected. Not sure if you're prefer we split it out to a different test so let me know.

I added a unit test to specifically check the behaviour of setting the `Name` field of the `IndicesGetDataStreamStatsRequest` as that was causing my original issue. I felt the Integration Test didn't really cover that specifically so a unit test would be a good way to ensure the URL is valid.

dblock (dismissed):
This is perfect, my favorite kind of PR that has way more tests than an otherwise trivial fix!

dblock:
@DewaldV see failures in tests + missing license header

DewaldV:
@dblock Apologies that I missed that. Just pushed the license headers!

dblock:
> @dblock Apologies that I missed that. Just pushed the license headers!

Cool. I disabled the PR behavior to require approvals from maintainers to run for 1st time contributors (made it for new GitHub users only)

DewaldV:
Ahh, the test failure I thought might be an issue but I ran a it a few times locally and it seemed OK so I left it as is.

I suspect the `/_data_stream/_stats` endpoint returns the data streams in any order.

When I tested locally it seemed consistent but it's likely consistent for my local cluster only compared to running it somewhere else.

I'll have to refactor that test to change how we do the assertion so its not order dependent.

DewaldV:
Hey @dblock. :wave:

After some research and attempts at fixing the ordering issue in tests I found the [jsonassert](https://github.com/kinbiko/jsonassert) library that tackles exactly these sorts of issues. It might be a very useful tool for other tests in this lib given the usage of JSON for all OpenSearch APIs.

It includes functionality for asserting JSON objects match without considering array order and should fix the flakiness in my original test.

I understand introducing a new dependency isn't ideal and might need more conversation but this is only used in Integration Tests at this point so hopefully it can be considered.

DewaldV:
I've checked the failures on the latest CI run and I can't replicate them locally. 

I'm convinced they are unrelated to my changes and tests and may just need a rerun to pass.

I can do some more digging if needed thoug, let me know what you think. :+1:

dblock (approved):

Issue #331, opened 2023-06-12
[FEATURE] Native JSON support


dblock:
### Is your feature request related to a problem?

Ingesting a document today requires supplying it as an `IoReader`. Starting with an example in https://xeiaso.net/blog/elasticsearch:

```go
data, err := json.Marshal(esEntry)
if err != nil {
	log.Fatalf("failed to marshal entry: %v", err)
}

resp, err := es.Index("site-search-kb", bytes.NewBuffer(data), es.Index.WithDocumentID(entry.ID))
```

Ugh.

### What solution would you like?

Automatically marshal objects to/from JSON. Starting with the canonical example from https://go.dev/blog/json I think I want something that doesn't say "IoStreams" everywhere.

```
m := Message{"Alice", "Hello", 1294706395881547000}
es.Index("site-search-kb", m, es.Index.WithDocumentID(1))
```

### Do you have any additional context?

Came from https://xeiaso.net/blog/elasticsearch

dblock:
@Jakob3xD would love your thoughts!

YuyaAbo:
I went through the code to understand the Issue and found `opensearchutil.NewJSONReader`. Is this sufficient to resolve the Issue? Is it enough to resolve the Issue or is there some functionality missing?

This utility function seems to be used as follows.

https://github.com/opensearch-project/opensearch-go/blob/c04e36a91a64755d6205a64326a65667d51d285d/opensearchutil/json_reader_integration_test.go#L55-L62

dblock:
@YuyaAbo Thank you!

1. Can this be somehow automatic? In this case this would be a feature request.
2. Do we document this anywhere? Either way we should document this in the user guide, then we can close this. Would you like to help?

YuyaAbo:
@dblock Yes, it would certainly be a useful addition if it could be automated.

I will add an introduction to `opensearchutil.NewJSONReader` to the user guide. I will send a PR later.

We will leave this issue as is for the purpose of adding functionality.

tannerjones4075:
Should this issue be closed? #341 seems to have committed a merge request to add the functionality unless I am mistaken.

dblock:
Yes, closing.

PR #341, opened 2023-06-28
Add an introduction to utility function NewJSONReader in USER_GUIDE

YuyaAbo:
### Description

There is a common case where a structure is used in the Body of an API request, which requires conversion to io.Reader. There is a useful utility function available for this use case, but it is not often noticed, so it is included in the user guide.

ref. https://github.com/opensearch-project/opensearch-go/issues/331#issuecomment-1608111897

### Issues Resolved

There is no Issue to be resolved, but the related Issue is https://github.com/opensearch-project/opensearch-go/issues/331

### DIFF
--- a/USER_GUIDE.md
+++ b/USER_GUIDE.md
@@ -86,13 +86,13 @@ func example() error {
 
 	// When using a structure, the conversion process to io.Reader can be omitted using utility functions.
 	document := struct {
-	    Title    string `json:"title"`
-        Director string `json:"director"`
-        Year     string `json:"year"`
-    }{
-        Title:    "Moneyball",
-        Director: "Bennett Miller",
-        Year:     "2011",
+		Title    string `json:"title"`
+		Director string `json:"director"`
+		Year     string `json:"year"`
+	}{
+		Title:    "Moneyball",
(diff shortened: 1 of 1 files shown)


USER_GUIDE.md
```diff
@@ -83,18 +84,22 @@ func example() error {
...
-	    "title": "Moneyball",
-	    "director": "Bennett Miller",
-	    "year": "2011"
-	}`)
+	// When using a structure, the conversion process to io.Reader can be omitted using utility functions.
+	document := struct {
+	    Title    string `json:"title"`
+        Director string `json:"director"`
+        Year     string `json:"year"`
+    }{
```
dblock, line 92: Looks like some tabs/spaces formatting broken here.

dblock (changes requested):
Thank you! Fix formatting and we're good to go?

USER_GUIDE.md
YuyaAbo: Fixed https://github.com/opensearch-project/opensearch-go/pull/341/commits/e1e12f6c616efc5e1c46594ba91dfab18b0c13ea

dblock (approved):

PR #348, opened 2023-07-19
Add opensearch 2.6.0, 2.7.0, 2.8.0 to compatibility test

Jakob3xD:
### Description
Add opensearch 2.6.0, 2.7.0, 2.8.0 to compatibility test

### DIFF
--- a/.github/workflows/test-compatibility.yml
+++ b/.github/workflows/test-compatibility.yml
@@ -19,6 +19,9 @@ jobs:
           - { opensearch_version: 2.3.0 }
           - { opensearch_version: 2.4.1 }
           - { opensearch_version: 2.5.0 }
+          - { opensearch_version: 2.6.0 }
+          - { opensearch_version: 2.7.0 }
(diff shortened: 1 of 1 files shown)


dblock:
Should we be concerned about failures with unreleased versions of OpenSearch?

dblock (approved):

Jakob3xD:
> Should we be concerned about failures with unreleased versions of OpenSearch?

@dblock It seems to me that the current 3.0.0 Snapshot does not return a `X-Opensearch-Version` Header. I am not sure if this is expected but that is the reason of failure. So it is not related to this MR. It is more general as the dependency updates PRs are also failing with the same error.

dblock:
That was added recently, wasn't it?
https://github.com/opensearch-project/OpenSearch/pull/8084/files

Jakob3xD:
> That was added recently, wasn't it?

Yes, I was confused by the job output but the tests use the wrong want/expect order.
The tests only expect the application/json header but now also got the version header.

PR #399, opened 2023-10-19
Adding a guide for making raw json requests using the client

### LINKED ISSUE #395: [FEATURE] Make raw JSON REST requests to OpenSearch

Coming from https://github.com/opensearch-project/opensearch-clients/issues/62, add support for making raw JSON requests.

Add a high level DSL, e.g. `client.get` and `client.post`.

Some of this may already be implemented under https://github.com/opensearch-project/opensearch-go/issues/331. At the very least we need documentation that doesn't use any high level DSL methods.

VachaShah:
### Description
Same as title

### Issues Resolved
Closes #395 

### DIFF
--- a/guides/json.md
+++ b/guides/json.md
@@ -1,0 +1,157 @@
+- [Making Raw JSON REST Requests](#making-raw-json-rest-requests)
+  - [Setup](#setup)
+  - [GET](#get)
+  - [PUT](#put)
+  - [POST](#post)
+  - [DELETE](#delete)
+
+# Making Raw JSON REST Requests
+
+The OpenSearch client implements many high-level REST DSLs that invoke OpenSearch APIs. However you may find yourself in a situation that requires you to invoke an API that is not supported by the client. Use `client.Perform` to do so.
+
+## Setup
+
+Let's create a client instance:
+
+```go
+package main
+
+import (
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"os"
+	"strings"
+
+	"github.com/opensearch-project/opensearch-go/v2/opensearchapi"
+)
+
+func main() {
+	if err := example(); err != nil {
+		fmt.Println(fmt.Sprintf("Error: %s", err))
+		os.Exit(1)
+	}
+}
+
+func example() error {
+	client, err := opensearchapi.NewDefaultClient()
+	if err != nil {
+		return err
+	}
+```
+
+## GET
+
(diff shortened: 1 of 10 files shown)


VachaShah:
This code is tested against OpenSearch cluster.

guides/json.md
```diff
@@ -0,0 +1,118 @@
...
+# Making Raw JSON REST Requests
+
+The OpenSearch client implements many high-level REST DSLs that invoke OpenSearch APIs. However you may find yourself in a situation that requires you to invoke an API that is not supported by the client. Use `client.Perform` to do so.
+
+## GET
+
+The following example returns the server version information via `GET /`.
+
+```go
+infoRequest, err := http.NewRequest("GET", "/", nil)
```
Jakob3xD, line 16: If you don´t want to check if the err is nil, you can omit it by using `_`. 
It would look like this:
```suggestion
infoRequest, _ := http.NewRequest("GET", "/", nil)
```
otherwise:
```suggestion
infoRequest, err := http.NewRequest("GET", "/", nil)
if err != nil {
    fmt.Printf("failed to perform request: %s", err)
    os.Exit(1)
}
```
This repeats for PUT,POST and DELETE.

Jakob3xD (changes requested):

dblock:
Thanks! Can we add a working sample into `samples` as part of this PR?

Jakob3xD (dismissed):

dblock (approved):
LGTM. Can we add a runnable sample along with this?

Jakob3xD (approved):

dblock:
We had a lot of contributions where code didn't work when copy-pasted, so in other clients we started adding samples that users can "just run", e.g. https://github.com/opensearch-project/opensearch-py/tree/main/samples. WDYT?

Jakob3xD:
Added samples for all guids, as I already had them when I updated the guids.

Once this is merge we could replace the code blocks of the guids with permanent links to the files instead of copy paste the code?

dblock:
> Once this is merge we could replace the code blocks of the guids with permanent links to the files instead of copy paste the code?

You could, but other clients don't really do that. I see guides kept to something concise and copy-pasted and working samples. Ideally samples should be compiled and run as integration tests and guide reference code, of course, but that needs tools.

dblock (approved):

dblock:
We forgot to link json.md from USER_GUIDE. Maybe in a future PR?

PR #410, opened 2023-10-31
create code-coverage-comment.yml file to create a comment of code coverage

tannerjones4075:
### Description
added code-coverage-comment.yml file to add a comment to each PR of code coverage
### Issues Resolved

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,8 +1,15 @@
-comment: off
+# https://docs.codecov.com/docs/pull-request-comments
+comment:
+  layout: "diff, flags, files"
+  behavior: default
+  require_changes: false  # if true: only post the comment if coverage changes
+  require_base: false        # [true :: must have a base report to post]
+  require_head: true       # [true :: must have a head report to post]
+  hide_project_coverage: false # [true :: only show coverage on the git diff]
 
 coverage:
   status:
+    project:
+      default:
+        threshold: 5%  # allow 5% coverage decrease
     patch: off
-
-ignore:
-  - "opensearchapi/api.*.go"
--- a/.github/workflows/test-unit.yml
+++ b/.github/workflows/test-unit.yml
@@ -21,7 +21,10 @@ jobs:
         with: { go-version: "${{ matrix.go }}" }
       - run: go version
       - run: make test-unit race=true
-      - uses: codecov/codecov-action@v1
+        if: matrix.os != 'ubuntu-latest'
+      - run: make test-unit race=true coverage=true
+        if: matrix.os == 'ubuntu-latest'
+      - uses: codecov/codecov-action@v3
         with:
           file: tmp/unit.cov
           flags: unit
--- a/.github/workflows/test-integration.yml
+++ b/.github/workflows/test-integration.yml
@@ -27,12 +27,11 @@ jobs:
           for attempt in `seq 25`; do sleep 5; \
           if curl -s localhost:9200; \
           then echo '=====> ready'; break; fi; if [ $attempt == 25 ]; then exit 1; fi; echo '=====> waiting...'; done
-      - run: make test-integ race=true
-      - uses: codecov/codecov-action@v1
+      - run: make test-integ race=true coverage=true
+      - uses: codecov/codecov-action@v3
         with:
-          file: tmp/integration-client.cov
+          file: tmp/integ.cov
           flags: integration
-
 
   secured:
     name: Tests against secure cluster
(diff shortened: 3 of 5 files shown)


tannerjones4075:
I am looking into commenting the report for the PR. I think that I need permissions to post the report. 

`- name: Create Comment Report
      uses: actions/github-script@v3
      with:
        script: |
          const issueComment = context.issue({
            body: 'This is a comment report on the PR.'
          });
          return github.issues.createComment(issueComment);
      env:
        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}`

.github/workflows/code-coverage-comment.yml
```diff
@@ -0,0 +1,17 @@
...
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 2
+      - uses: actions/setup-go@v4
+        with: { go-version: '1.x' }
+      - name: Run coverage
+        run: go test -race -coverprofile=coverage.out -covermode=atomic
+      - name: Upload coverage to Codecov
+        uses: codecov/codecov-action@v3
```
Jakob3xD, line 17: Can you try using the gen-coverage. It combines all Unit and Integration tests. Otherwise we only get the unit test coverage.
```suggestion
        run: make gen-coverage
      - name: Upload coverage to Codecov
        uses: codecov/codecov-action@v3
        with:
          files: ./tmp/total.cov
```

.codecov.yml
Jakob3xD, line 1: IMO the `ignore` can be removed as we also want the coverage on those API endpoints.

Jakob3xD (commented):
You could also remove the cov section from `.github/workflows/test-unit.yml` and `.github/workflows/test-integration.yml` as they both use the v1 version which is deprecated and this workflow would make it obsolete.

Jakob3xD:
Ah, haven't though about that `gen-coverage` also executes the tests. Therefore it fails because no opensearch is running.

.github/workflows/code-coverage-comment.yml
Jakob3xD: You could try adding the coverage to the `.github/workflows/test-integration.yml` workflow as it already starts opensearch but it would then run the test twice.

It might be needed to adjust the Makefile.
The `gen-coverage` executes the `go test` it self. Maybe it makes more sense to adjust the args of integ and unit test to match the current one of `gen-coverage`, so `gen-coverage` can just call the `make test-integ` and `make test-unit`.  Resulting in replacing the `make test-integ` in  `.github/workflows/test-integration.yml`  with `make gen-coverage`.
If this is to weird to understand, I can take a look tomorrow.

tannerjones4075: I have modified the makefile and the codecov action to make test-integ and make test-unit.

tannerjones4075:
> Ah, haven't though about that `gen-coverage` also executes the tests. Therefore it fails because no opensearch is running.

Yes, I was running into this issue. I have tried a several ways. If you could take a look see what I am doing wrong that would be great.

Jakob3xD:
> Yes, I was running into this issue. I have tried a several ways. If you could take a look see what I am doing wrong that would be great.

I hope it was okay that I hijacked your PR. I tested a bit around and changed the Workflows to report coverage for intergration, unit and total coverage.

tannerjones4075:
> > Yes, I was running into this issue. I have tried a several ways. If you could take a look see what I am doing wrong that would be great.
> 
> I hope it was okay that I hijacked your PR. I tested a bit around and changed the Workflows to report coverage for intergration, unit and total coverage.

I appreciate your help. I am glad that we were able to figure it out. It would be nice to get some credit for the PR for the OSCI Program perspective.

Jakob3xD:
@dblock I made some changes to the commits, therefore I don´t want to review my own code. Would you mind having a look? The coverage decreases as we no longer ignore all the imported code :D. Should increase again with my current refactor PR.
Beside I am not familiar with the OCSI Program and how ppl get credit.

.github/workflows/test-unit.yml
```diff
@@ -21,7 +21,10 @@ jobs:
         with: { go-version: "${{ matrix.go }}" }
       - run: go version
       - run: make test-unit race=true
-      - uses: codecov/codecov-action@v1
+        if: matrix.os != 'ubuntu-latest'
+      - run: make test-unit race=true coverage=true
+        if: matrix.os == 'ubuntu-latest'
+      - uses: codecov/codecov-action@v3
```
dblock, line 27: What's the reason not to run code coverage on other platforms? Codecov supports merging multiple uploads, in fact it's better because some tests may not run on all platforms.

dblock (approved):

dblock:
Looks like tests failed :(

.github/workflows/test-unit.yml
dblock: It's good practice to run tests cross platform making sure we don't break anything since we claim support for those platforms. No platform specific stuff today doesn't mean it can't be accidentally introduced.

Issue #470, opened 2024-02-13
[BUG] The Search API (in Version 3.0.0) is not working without setting an Index


dom192:
### What is the bug?
It  seems, that it is not possible to call the Search function of Client:
```go
// Search executes a /_search request with the optional SearchReq
func (c Client) Search(ctx context.Context, req *SearchReq) (*SearchResp, error)
``` 
without setting `Indices` in the `SearchReq`.
			
### How can one reproduce the bug?
Call the Search function without specifying Indices 
```go
	content := strings.NewReader(
	`	{
		"query": {
		  "match_all": {}
		}
	  }`)

	resp, err := client.Search(
		ctx,
		&opensearchapi.SearchReq{
			Indices: []string{},
			Body:    content,
		},
	)
``` 

### What is the expected behavior?
The expected result should be no error and a valid opensearch result, but instead I get an error:
`error msg: failed to json unmarshal body: invalid character '<' looking for beginning of value`
 The root cause of the problem is getting a http status code 400 Bad Request with http body:
`"<html>\r\n<head><title>400 Bad Request</title></head>\r\n<body>\r\n<center><h1>400 Bad Request</h1></center>\r\n<hr><center>nginx</center>\r\n</body>\r\n</html>\r\n"`
Which is not handled right. The code tries to ` json.Unmarshal` the html and generates the misleading error message.

### What is your host/environment?
Ubuntu 22.04.3 LTS
API version 3 (github.com/opensearch-project/opensearch-go/v3/opensearchapi)

### Do you have any screenshots?
no

### Do you have any additional context?
I debugged the issue and found the root cause I think. The problem is the `GetRequest` function:
![image](https://github.com/opensearch-project/opensearch-go/assets/23055909/dfca5e0d-21af-42ab-bd93-4ab490874329)
In case of no Indices is set, path is set to _search
```go
path = "_search"
``` 
this causes the Bad Request response. To fix this I appended a "/" to _search:
```go
path = "/_search"
``` 
after that the everything worked like expected :)

Jakob3xD:
Tanks for reporting this issue.
Yes, the path should be prefixed with `/ `.

Can you please tell what you client looks like? Are you using AWS Signing or something like that, as I can not reproduce it with our integration tests and I want to adjust them so it does not happen again.

dom192:
> Tanks for reporting this issue. Yes, the path should be prefixed with `/ `.
> 
[...]

PR #471, opened 2024-02-14
Fix search request missing a slash when no indices are given

Jakob3xD:
### Description
Add missing `/ ` for search requests when not searching specific indices.

### Issues Resolved
Closes https://github.com/opensearch-project/opensearch-go/issues/470

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 ### Removed
 
 ### Fixed
+- Fix search request missing a slash when no indices are given ([#470](https://github.com/opensearch-project/opensearch-go/pull/469))
 
 ### Security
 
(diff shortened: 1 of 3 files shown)


Jakob3xD:
Draft as I am waiting for a reply on the issue to maybe adjust the integration tests.

dblock:
Definitely needs tests! Thanks for fixing this.

dblock:
@Jakob3xD Want to finish this?

Jakob3xD:
> @Jakob3xD Want to finish this?

Well, I can mark it as ready but I have no idea for testing this in a useful way for all endpoints.
We could add a unit test for each endpoint creating a requests and calling the `GetRequest()` function and check the URL if it stats with a `/`. But I am not sure this works.

dblock:
> > @Jakob3xD Want to finish this?
> 
> Well, I can mark it as ready but I have no idea for testing this in a useful way for all endpoints. We could add a unit test for each endpoint creating a requests and calling the `GetRequest()` function and check the URL if it stats with a `/`. But I am not sure this works.

I would at a minimum add a test for this specific endpoint (this bug). But something that iterates over all APIs would be better.

dblock (dismissed):

dblock (approved):

PR #449, opened 2023-12-18
Remove references to the default credentials of admin:admin

derek-ho:
### Description
Remove references to the default credentials of admin:admin and changes the CI according to default admin credentials changes for the 2.12.0 release.

### Issues Resolved
_List any issues this PR will resolve, e.g. Closes [...]._ 

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #482, opened 2024-03-07
opensearchapi: Added new struct fields introduced from opensearch main

Jakob3xD:
### Description
Adds missing struct fields that fails the validation against unreleased opensearch

### Issues Resolved


Jakob3xD:
This should fix all workflows except the security one which is handled in #449

dblock (approved):

Jakob3xD:
106106d488c6fd0668d53af3d987b2b371542b4a is a type change and therefore a braking change?
If yes then we maybe want to consider stop testing against unrealesed Opensearch as it may bring a lot of braking changes with it and only the changes once we support 3.0.0?

dblock:
@Jakob3xD That is being reverted in https://github.com/opensearch-project/OpenSearch/pull/12543

PR #488, opened 2024-03-13
Adjust tests to work with secure cluster

Jakob3xD:
### Description
- Update workflow action versions
- Move functions from opensearchapi/internal/test to internal/test
- Adjust integration test to work with secure opensearch
- adjust golangci-lint config for internal/ folder
- add missing merge_id field for indices segement request
- change custom_foldername field to pointer as it can be null
- skip render search template test with security due to issue

### Issues Resolved

### DIFF
--- a/internal/test/helper.go
+++ b/internal/test/helper.go
@@ -1,0 +1,156 @@
+// SPDX-License-Identifier: Apache-2.0
+//
+// The OpenSearch Contributors require contributions made to
+// this file be licensed under the Apache-2.0 license or a
+// compatible open source license.
+
+package ostest
+
+import (
+	"context"
+	"crypto/tls"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"os"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/wI2L/jsondiff"
+
+	"github.com/opensearch-project/opensearch-go/v3"
+	"github.com/opensearch-project/opensearch-go/v3/opensearchapi"
+)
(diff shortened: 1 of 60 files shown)


Jakob3xD:
Should be easier to review this MR commit by commit.

internal/test/helper.go
```diff
@@ -0,0 +1,139 @@
...
+	"github.com/stretchr/testify/require"
+	"github.com/wI2L/jsondiff"
+
+	"github.com/opensearch-project/opensearch-go/v3"
+	"github.com/opensearch-project/opensearch-go/v3/opensearchapi"
+)
+
+// NewClient returns an opensearchapi.Client that is adjusted for the wanted test case
+func NewClient() (*opensearchapi.Client, error) {
+	if os.Getenv("SECURE_INTEGRATION") == "true" {
```
dblock, line 29: You should be able to refactor this away by having `ClientConfig()` return `nil` when there isn't one defined.

dblock (approved):
This is good and I'd merge as is. 

I think having `SECURE_INTEGRATION` being referenced as an environment variable in more than one place isn't great. I would set `OPENSEARCH_USERNAME`, `OPENSEARCH_PASSWORD`, and `OPENSEARCH_URL`m, rather than trying to drive everything off the one environment variable to make it more idiomatic, but this can be improved later (or not).

Jakob3xD:
> I think having `SECURE_INTEGRATION` being referenced as an environment variable in more than one place isn't great.

It is now only referenced in the `IsSecure()` function. It makes more sense this way.

dblock (approved):
Good work!

PR #507, opened 2024-04-03
Add security plugin

### LINKED ISSUE #94: [FEATURE] Types for users and roles in opensearchapi
As I can see, currently there is no support for the users/roles API in opensearchapi. Could you please include this API to the client?

Jakob3xD:
### Description
- adds security plugin
- moves errors to opensearch package
- add more error types and parse them by type assertion

The changes made to error are more or less braking changes as they get moved to another package.
Normal use should still work but for uses who try to parse the error with `errors.As` it is braking.

### Issues Resolved
Closes #94 

### DIFF
--- a/.ci/opensearch/Dockerfile.opensearch
+++ b/.ci/opensearch/Dockerfile.opensearch
@@ -4,9 +4,21 @@ FROM opensearchproject/opensearch:${OPENSEARCH_VERSION}
 ARG opensearch_path=/usr/share/opensearch
 ARG SECURE_INTEGRATION
 ENV SECURE_INTEGRATION=$SECURE_INTEGRATION
+ARG OPENSEARCH_INITIAL_ADMIN_PASSWORD
 
+# Some opensearch secuirty settings are only present since 2.8.0 and causes older versions to brake if the setting is present
+# https://apple.stackexchange.com/a/123408/11374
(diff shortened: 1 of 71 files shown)


Jakob3xD:
:partying_face: CI is green. I am going to split all the unrelated stuff into separate PRs.

dblock (dismissed):
Nice. Did you roll this code out by hand?

Jakob3xD:
> Nice. Did you roll this code out by hand?

Yes, as the API spec is IMO not in a great state because of issue like [this](https://github.com/opensearch-project/opensearch-api-specification/issues/202) and my focus is on adding the missing plugins I need. As I have no experience with code generation it would take way more time to add the plugins and iron out all the issues the api spec has. 

I am also working on the ISM plugins but the API spec is totally missing and I am trying to get what I need from the Docs but this is also a lot of pain.

dblock:
@Jakob3xD we have done a lot of work in https://github.com/opensearch-project/opensearch-api-specification, moving it to OpenAPI, adding bodies, and generally are going to push on teams such as ISM to add specs. You can get a lot of mileage "for free" there. Remember that we have 8 clients, so in order to keep up with API changes we have to automate.

dblock:
@Jakob3xD rebase?

dblock (approved):

PR #512, opened 2024-04-05
Move error structs from opensearchapi to opensearch to be reused by plugins

Jakob3xD:
### Description
- Move error structs from opensearchapi to opensearch including tests
- Add new errors structs
- Move parseError function from opensearchapi to opensearch error.go as ParseError
- Change ParseError function to do type assertion to determine the correct error type
- Adjust tests to new errors
- Handle non json errors

This is a braking change as the error types are moved to another project and also get renamed.
Needed as preparation for plugin integrations.

### Issues Resolved


dblock (changes requested):
Looks like a publicly-facing backwards-incompatible breaking change? Unless you want to inherit these error types in old names and mark those deprecated, let's bump the version to 4 as part of it?

Jakob3xD:
> Looks like a publicly-facing backwards-incompatible breaking change? Unless you want to inherit these error types in old names and mark those deprecated, let's bump the version to 4 as part of it?

I adjusted the upgrading.md to version 4.0.0. Do you want the release done in this PR or after all my other PRs are merged?
I'd prefer after all the others are merged.

dblock:
> > Looks like a publicly-facing backwards-incompatible breaking change? Unless you want to inherit these error types in old names and mark those deprecated, let's bump the version to 4 as part of it?
> 
> I adjusted the upgrading.md to version 4.0.0. Do you want the release done in this PR or after all my other PRs are merged? I'd prefer after all the others are merged.

After. Usually the version number should be set on working code as early as possible so that nobody confuses it with released code of the previous version.

dblock (approved):
Merging this.

Would it make sense to make some of things like `ParseError` or `UnmarshalJSON` private only?

PR #518, opened 2024-04-08
Updates USER_GUIDE.md and /_samples/

mstephenson6:
### Description

Cleans up example code so it runs and shows a complete cycle of index creation, document indexing, searching, document deletion, and index deletion.

### DIFF
--- a/_samples/search.go
+++ b/_samples/search.go
@@ -7,13 +7,16 @@ package main
 
 import (
 	"context"
+	"crypto/tls"
 	"encoding/json"
 	"fmt"
+	"net/http"
 	"os"
 	"strconv"
 	"strings"
 	"time"
 
+	"github.com/opensearch-project/opensearch-go/v3"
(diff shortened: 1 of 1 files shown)


mstephenson6:
> Thanks!
> 
> I believe this matches what we have in https://github.com/opensearch-project/opensearch-go/tree/main/_samples, so that should be updated as well, and made sure to run. Rebase please, too.

@dblock - Yeah, I think the files in `/_samples` need attention for login at least, `DefaultNewClient()` might not work at all with the v2.13 strong password requirement? I'll line them up with the USER_GUIDE.md change.

mstephenson6:
@dblock - this is cleaned up and ready for review. Thanks for pointing out the other samples. I updated all of them to work with the default strong password in v2.13, and the error-handling code in USER_GUIDE.md now uses `opensearch.Error` from its new package on main.

dblock (approved):

dblock:
TY!

Issue #74, opened 2021-10-20
[FEATURE] Types for ISM in opensearchapi


dbason:
**Is your feature request related to a problem?**
Currently there's no support for the ISM APIs in opensearchapi.  This means to use that feature you need to manually extend the API yourself

**What solution would you like?**
It would be nice if this was included.

**What alternatives have you considered?**
Currently we have written our own implementation.

wbeckler:
@dbason at the moment nobody is working on this. If you have a working implementation, would you be willing to share it as a PR? There is an open question of where to put the plugin methods and how to name them. I think it would be useful to see a working implementation.

saketjain01:
Hello Everyone,

Curious to know if there is any further information available on support of this feature in opensearchapi-go. Or any pointers to the alternatives available to implement ISM policy for data streams in golang.

Appreciate any help here.

salamer:
any progress?

dblock:
@salamer AFAIK nobody is working on this, please feel free to pick it up

PR #524, opened 2024-04-11
Add ISM Policy

Jakob3xD:
### Description
- add ISM Policy

### Issues Resolved
Closes https://github.com/opensearch-project/opensearch-go/issues/74

### DIFF
--- a/plugins/ism/api_test.go
+++ b/plugins/ism/api_test.go
@@ -1,0 +1,260 @@
+// SPDX-License-Identifier: Apache-2.0
+//
+// The OpenSearch Contributors require contributions made to
+// this file be licensed under the Apache-2.0 license or a
+// compatible open source license.
+//
+//go:build integration
+
+package ism_test
+
+import (
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	ostest "github.com/opensearch-project/opensearch-go/v3/internal/test"
+	"github.com/opensearch-project/opensearch-go/v3/opensearchapi"
+	"github.com/opensearch-project/opensearch-go/v3/plugins/ism"
+	osismtest "github.com/opensearch-project/opensearch-go/v3/plugins/ism/internal/test"
+)
+
+func TestClient(t *testing.T) {
+	t.Parallel()
+	client, err := osismtest.NewClient()
+	require.Nil(t, err)
+
+	osClient, err := ostest.NewClient()
+	require.Nil(t, err)
+
+	failingClient, err := osismtest.CreateFailingClient()
+	require.Nil(t, err)
+
+	testPolicy := "testPolicy"
+	testIndex := []string{"test_policy"}
+
+	t.Cleanup(func() { client.Policies.Delete(nil, ism.PoliciesDeleteReq{Policy: testPolicy}) })
+	_, err = client.Policies.Put(
+		nil,
+		ism.PoliciesPutReq{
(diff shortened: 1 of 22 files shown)


internal/test/helper.go
```diff
@@ -118,6 +118,20 @@ func SkipIfNotSecure(t *testing.T) {
 	}
 }
 
+// SkipIfPluginMissing skips a test if the required plugin is missing
+func SkipIfPluginMissing(t *testing.T, client *opensearchapi.Client, plugin string) {
```
dblock, line 122: This may be a bit dangerous since we can be having a passing build, but actually no tests for a given plugin running because of some setup problem.

plugins/ism/api_test.go
```diff
@@ -0,0 +1,261 @@
...
+				},
+			},
+		},
+		{
+			Name: "Change",
+			Tests: []clientTests{
+				{
+					Name: "with request",
+					Results: func() (osismtest.Response, error) {
+						return client.Change(nil, ism.ChangeReq{Indices: testIndex, Body: ism.ChangeBody{PolicyID: testPolicy, State: "delet"}})
```
dblock, line 161: "delet"?

dblock (approved):
LGTM, not sure about that delete below.

I really think you should look into code generation.

internal/test/helper.go
Jakob3xD: The issue is the test against unreleased opensearch where only the core opensearch gets build so not plugin is installed.
The other option would be to check for an ENV or something like that. Maybe a combination of both. If the ENV is set it is okay if the plugins are missing and if the ENV is missing it fails?

dblock: I think you want to be explicit about it in all cases. For example, run a different test suite or tests in different folders in unreleased version vs. non unreleased version?

dblock (changes requested):
Let's get rid of the skip (https://github.com/opensearch-project/opensearch-go/pull/524#discussion_r1565609941) and replace it with something explicit?

internal/test/helper.go
Jakob3xD: I am not sure I can follow your idea but it sounds like you want to use the build tags for testing. So we can test opensearchapi only for unreleased opensearch and test opensearch,security,ism for the other things?
This would require us to tag each file.

dblock: I am not familiar with the mechanisms for this in Go. Tags would work.

In other languages/frameworks we have things like test suites or simply different folders, where you can explicitly say "run all tests" or "run tests in this folder".

Jakob3xD: I added the unreleased option to the make test-integ run to only run tests in the opensearchapi folder.
The opensearchapi commits are issues that came up during the tests.

dblock (approved):
This is much butter, and you can merge as is if you want.

That said, I think the level of indirection of unreleased -> testintegpath is making it unnecessarily complicated. Why not just pass an optional test integ path?

```
make test-integ race=true testintegpath="./opensearchapi/..."
```

This still wouldn't convey the semantics of what we're trying to say, which is that when plugins are installed we want to run integration tests for plugins. I'd expect to be able to do this:

```
make test-integ // runs all, default, for developers
make test-integ-min // tests for a min distribution of OpenSearch (runs in all)
make test-integ-plugins // tests for plugins, does not include -min (runs in all except unreleased)
```

(Please tell me if I am annoying ;))

Jakob3xD:
Merging it but I will create a follow up PR for the testing stuff.

PR #542, opened 2024-05-09
Fix empty request body on retries with compression enabled

### LINKED ISSUE #541: [BUG] Error http: ContentLength=401 with Body length 0
### What is the bug?
I'm getting lots of these errors when indexing documents into a data stream:
`http: ContentLength=401 with Body length 0`

Here's my code:
```go
func (logger *MyLogger) SendLogEvent(event *MyLogEvent) error {
	// Marshal the struct
	s, err := json.Marshal(event)
	if err != nil {
		return WrapError("failed to marshal log event", err)
	}
	// Do the request
	req, err := opensearchapi.IndexReq{
		Index: logger.logsIndexName,
		Body: bytes.NewReader(s),
	}.GetRequest()
	if err != nil {
		return WrapError("failed to create OpenSearch index request", err)
	}
	resp, err := logger.client.Perform(req)
	// Make sure to close the body so the connection can be reused
	if resp != nil && resp.Body != nil {
		defer resp.Body.Close()
	}
	if resp == nil {
		return err
	}
	if err != nil || resp.StatusCode != 201 {
		// Log the error
		return errors.New("failed to log event, statuscode " + resp.Status)
	}
	// Still read the body so the connection can be reused
	io.Copy(io.Discard, resp.Body)
	// Success
	return nil
}
```

No error.

Debian 12 and Linux Mint 21.3
opensearch-go v4.0.0

merlinz01:
### Description
Make the transport use a new reader for each retry when gzip compression is enabled.
Also be more explicit about doing so for non-compressed requests.
Test included.

### Issues Resolved
Closes #541.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 ### Security
 
 ### Dependencies
-- Bump `github.com/aws/aws-sdk-go` from 1.51.21 to 1.51.31 ([#534](https://github.com/opensearch-project/opensearch-go/pull/534), [#537](https://github.com/opensearch-project/opensearch-go/pull/537))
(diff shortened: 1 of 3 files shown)


merlinz01:
I forgot to create a new branch for the PR. Please disregard.

PR #543, opened 2024-05-09
Fix empty request body on retries with compression enabled

### LINKED ISSUE #541: [BUG] Error http: ContentLength=401 with Body length 0

merlinz01:
### Description
Make the transport use a new reader for each retry when gzip compression is enabled.
Also be more explicit about doing so for non-compressed requests.
Test included.

### Issues Resolved
Closes #541.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ### Fixed
 
+- Fixes empty request body on retry with compression enabled. ([#542](https://github.com/opensearch-project/opensearch-go/pull/542))
+
 ### Security
 
 ### Dependencies
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ### Fixed
 
-- Fixes empty request body on retry with compression enabled. ([#542](https://github.com/opensearch-project/opensearch-go/pull/542))
+- Fixes empty request body on retry with compression enabled ([#543](https://github.com/opensearch-project/opensearch-go/pull/543))
 
 ### Security
 
--- a/opensearchtransport/opensearchtransport.go
+++ b/opensearchtransport/opensearchtransport.go
@@ -245,7 +245,9 @@ func (c *Client) Perform(req *http.Request) (*http.Response, error) {
 			}
 
 			req.GetBody = func() (io.ReadCloser, error) {
-				return io.NopCloser(buf), nil
+				// We have to return a new reader each time so that retries don't read from an already-consumed body.
+				reader := bytes.NewReader(buf.Bytes())
+				return io.NopCloser(reader), nil
 			}
 			//nolint:errcheck // error is always nil
 			req.Body, _ = req.GetBody()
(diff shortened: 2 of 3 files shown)


dblock (approved):

opensearchtransport/opensearchtransport.go
```diff
@@ -258,8 +260,9 @@ func (c *Client) Perform(req *http.Request) (*http.Response, error) {
 				//nolint:errcheck // ignored as this is only for logging
 				buf.ReadFrom(req.Body)
 				req.GetBody = func() (io.ReadCloser, error) {
-					r := buf
-					return io.NopCloser(&r), nil
+					// Return a new reader each time
+					reader := bytes.NewReader(buf.Bytes())
+					return io.NopCloser(reader), nil
```
dblock, line 265: Are we missing a test for this path?

merlinz01: There is an existing test for the reader reset behavior on non-compressed request body:
https://github.com/opensearch-project/opensearch-go/blob/41fd58f31b62a0d1edb6f7e506640c79b6a6017f/opensearchtransport/opensearchtransport_internal_test.go#L655-L690

dblock: But that test is passing without your code change?

merlinz01: Notice there are two test cases now, one for retries without compression and one (from this PR) for retries with compression.

The code for uncompressed works correctly and passes the relevant test, both with and without my changes.
The old code does a copy of the unread `buf` into `r` so that `buf` as referenced from the closure remains unread. With my changes, it is a bit more obvious what is being done but it functions the same.
So no, if you revert those two lines, no tests will fail. The bug is only in the code for compressed.

The code for uncompressed was from https://github.com/opensearch-project/opensearch-go/commit/fb7bdd15c1e392734922bece65d3c203a4aefb95 and code for compressed was from https://github.com/opensearch-project/opensearch-go/commit/8cf8bd25e2b7f6fefe3b946f20555272bf8c3c47 but the bug was introduced in https://github.com/opensearch-project/opensearch-go/commit/7ea16f7de5fa5d45a0930619868466f5fc430ee2 with pooled gzip compressors.

dblock: Understood! Thank you for being patient with me.

dblock:
@merlinz01 Looks like you need to fix DCO (`git commit -s --amend`), please?

CHANGELOG.md
```diff
@@ -17,6 +17,8 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ### Fixed
 
+- Fixes empty request body on retry with compression enabled. ([#543](https://github.com/opensearch-project/opensearch-go/pull/543))
```
dblock, line 20: Nit: remove period at the end of the sentence, other lines don't have it.

merlinz01: Done.

merlinz01:
DCO successful.

dblock (approved):

dblock:
Thanks! We have lots of other things in https://github.com/opensearch-project/opensearch-go/issues if you have free time :)

Issue #560, opened 2024-06-04
[BUG] MappingFieldResp struct is not parsed correctly.


agnivade:
### What is the bug?
MappingFieldResp struct is not parsed correctly.

### How can one reproduce the bug?

A short snippet

```go
mappingFieldResp, err := client.Indices.Mapping.Field(ctx, &opensearchapi.MappingFieldReq{
		Fields:  []string{"type"},
		Indices: []string{indexName},
	})
	fmt.Println("---------------------")
	io.Copy(os.Stderr, mappingFieldResp.Inspect().Response.Body)
	fmt.Printf("\n%+v\n", mappingFieldResp.Indices)
	fmt.Printf("\n---------------------\n")
```

prints 

```
---------------------
{"channels":{"mappings":{"type":{"full_name":"type","mapping":{"type":{"type":"keyword"}}}}}}
map[]
---------------------
```

### What is the expected behavior?
It should parse the data correctly

### What is your host/environment?
Linux

### Do you have any additional context?
The issue is that there is an additional `Indices` field which is there inside the struct which is not allowing the response to be parsed properly.

Here: https://github.com/opensearch-project/opensearch-go/blob/6836d76494fa993ba7874b2f1762993a5ffaf40f/opensearchapi/api_indices-mapping.go#L63, we need to pass `data.Indices` instead of `data`.

I am not sure how the unit tests are passing. But it seems like a pretty straightforward bug to me.

EDIT: Something like this works:
```go

var mappingFieldResp map[string]struct {
  Mappings json.RawMessage `json:"mappings"`
}
httpResp, err := es.client.Client.Do(ctx, &opensearchapi.MappingFieldReq{
  Fields:  []string{"type"},
  Indices: []string{indexName},
}, &mappingFieldResp)
```

dblock:
Looks like a miss. Want to fix it here?

Also note that we would like to generate the client from spec to catch all this kind of problems (#284), so would be great to check the [spec](https://github.com/opensearch-project/opensearch-api-specification) for correctness for this API.

agnivade:
I can, but currently I have a workaround so it's not urgent for me. I'll see if I can find some time.

PR #572, opened 2024-06-26
Field, Function and test fixes, adding missing struct to node stats

Jakob3xD:
### Description
- Fix FLS role type
- Fix json compare when response is null
- Fix function where the wrong var is given to parse the response
- Add missing node stats cache field 

### Issues Resolved
Closes https://github.com/opensearch-project/opensearch-go/issues/560 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - Adds the `Caches` field to Node stats ([#572](https://github.com/opensearch-project/opensearch-go/pull/572))
 
 ### Changed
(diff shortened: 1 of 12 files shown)


dblock (dismissed):
Want to write some tests for this one? Looks like lots of tests failed.

Jakob3xD:
> Want to write some tests for this one? Looks like lots of tests failed.

Test are failing because of the omitempty tag causing the json compare to omit them and resulting in missing fields when comparing. I am currently thinking about a solution.

Jakob3xD:
I am not totally happy with this solution as I would have liked to not duplicate the structs but there is not lib that compares a struct against a json blob and respects the omitempty tag. Or at least I did not find such a lib.
Therefore duplicating the structs but without omitemtpy is simplest and easiest way of solving this.

dblock (approved):

Issue #284, opened 2023-04-06
[FEATURE] Generate API code from OpenSearch API Spec


VachaShah:
### Is your feature request related to a problem?
Related to https://github.com/opensearch-project/opensearch-clients/issues/19

### What solution would you like?
Re-use the generator that generates API Code that maintains backwards compatibility with the current client APIs and provides a way to add new APIs.

### What alternatives have you considered?
Using OpenAPI generators instead of re-using the existing generator but it makes it difficult to maintain backwards compatibility. 

### Do you have any additional context?
_Add any other context or screenshots about the feature request here._

tannerjones4075:
I would like to help with this issue. Is it still relevant?

VachaShah:
Hi @tannerjones4075, yes absolutely! The first step would be use to the API specs from https://github.com/opensearch-project/opensearch-api-specification/blob/main/OpenSearch.openapi.json and modify the existing non-functional generator in the go client to create code similar to the existing client. The general idea is to replace all the code (module by module or divided in whatever way works) in this repo for the APIs with code generated from the API specs without breaking the users.

Jakob3xD:
I am currently trying to generate sturcts out of the component schemas and I don´t think I can continue at this point.
There are lots of object properties defined which no longer exist in the API and are referencing to types that can not be handled in opensearch.

`component/schemas/_common:ByteSize` is of type `oneOf` pointing to type `number` or `string`. 
Inside golang it is not possible for a field to be both. We would need to set the field to `json.RawMessage` which is just a wrapper for `any`, forcing the user to find out the type of the field on his own.
`_common:ByteSize` is referenced quite some time but when going though the the OpenAPI spec and our current golang code, I noticed that many fields that reference `_common:BysteSize` are no longer returned by the API as they got replaced by the same property suffixed with `_in_bytes` pointing to type number.
As we validate all our golang structs against the API and check for missing fields, I assume that those fields must be very old as they never existed in any Opensearch 1.x o 2.x release. 

For example:

https://github.com/opensearch-project/opensearch-api-specification/blob/9c15bbf1522b1604e2e36f16b53163413f5fe833/spec/schemas/_common.yaml#L982-L986

Should I open an issue for that in the api spec repository? @dblock 
Similar issue for `component/schemas/_common:Percentage`. It is also `oneOf` for `number`  or `string`.
This is used for the cluster health endpoint. Which returns the value as number for yaml and json.


[...]

PR #574, opened 2024-06-27
Add missing SeqNo and PrimaryTerm fields in SearchHit

emtammaru:
### Description
Adds the missing fields `SeqNo` and `PrimaryTerm` to the `SearchHit` struct.  These fields are optionally returned in the response as documented here with the `seq_no_primary_term` param: https://opensearch.org/docs/latest/api-reference/search/

### Issues Resolved
N/A


dblock:
This needs a rebase.

The proposal in #284 might be of interest. Also appreciate if you could take this opportunity to check https://github.com/opensearch-project/opensearch-api-specification for these fields and maybe add a test there.

dblock (approved):

emtammaru:
> Thanks! This needs a rebase.
> 
> The proposal in #284 might be of interest. Also appreciate if you could take this opportunity to check https://github.com/opensearch-project/opensearch-api-specification for these fields and maybe add a test there.

Thank you, I checked the spec and I see the fields under response for https://opensearch-project.github.io/opensearch-api-specification/#/default/search.0. Could you point me to where I could add a test for it in https://github.com/opensearch-project/opensearch-api-specification?

dblock:
I think https://github.com/opensearch-project/opensearch-api-specification/blob/73d552477f8f006f5ff49f451494d150b8378b18/tests/_core/search.yaml would be a logical place. As long as these are mentioned somewhere.

Jakob3xD (approved):

PR #624, opened 2024-09-29
Return HTTP response as second return value from typed API client functions

### LINKED ISSUE #619: [FEATURE] HTTP response as second return value of functions of the `opensearchapi` package
In a project I'm working on, I have some integration tests that are testing some code that relies on the functions of the `opensearchapi` package. However, when it's time to assert that the output is equal to the expectations, it's not possible to write a simple `require.Equal(t, wantResp, resp)` because the response object returned by the client includes the `response *opensearch.Response` field, which is hard to predict.

The solution I propose is to return the HTTP response as second return value, instead of inside the API response object. In this way, `SomeResp` becomes somewhat predictable, as it would no longer contain a field whose value is unknown until the request is actually made. In this way, code like the one below is possible without much effort:

```go
resp, err := s.opensearchClient.SomeFunc(context.Background(), SomeReq{
	// Params
})

want := &elastic_types.SomeResp{
	// Expected values
}

require.NoError(t, err)
require.Equal(t, want, resp)
```

Roughly, the changes would be these ones:

```diff
type SomeResp struct {
-	response *opensearch.Response
}

-func (c *Client) SomeFunc(ctx context.Context, req SomeReq) (*SomeResp, error) {
+func (c *Client) SomeFunc(ctx context.Context, req SomeReq) (*SomeResp, *opensearch.Response, error) {
	var (
		data SomeResp
-		err  error
	)

-	if data.response, err = c.do(ctx, req, &data); err != nil {
+	resp, err := c.do(ctx, req, &data)
+	if err != nil {
-		return &data, err
+		return nil, resp, err
+	}

-	return &data, nil
+	return &data, resp, nil
}
```

An alternative solution might be to compare each field of the response object separately. However, this doesn't scale well when there are a lot of fields and makes the code longer and harder to read and maintain.

As this is a breaking change, the downside is that it might take quite some time to be released. I would also be open to contribute this change myself, if accepted.

ste93cry:
### Description
I'm refactoring the functions of the typed clients to return the HTTP response as second value.

### Issues Resolved
Closes #619


dblock:
Big breaking change :( 

If `data, response, err` was flipped as `data, err, response`, would it be easier to migrate to? Does go allow to specify the first 2 objects and omit the third?

ste93cry:
No, it doesn’t. Even though it’s a breaking change, it’s quite easy to track it and solve it as the project will not compile until the new returned value is either ignored or handled, and find and replace in most cases might be all a user needs to do if he never bothers about the response. Considering that this change is targeting the next major version, where this kind of things are expected to happen when updating, I don’t see big downsides.

dblock:
I think we still want this. I'd want other maintainers to chime into this like @Jakob3xD.

ste93cry:
From my side, I'm still open to finish the changes, but given the amount of time that must be spent on fixing all stuffs I want to be sure first that this is gonna be merged at some point

dblock:
Understandable. All this boilerplate should be generated though, the big ticket that someone should tackle first is #284 and this change becomes trivial.

sean-:
Hi @ste93cry, thanks for keeping this open and apologies for the long silence.

Status: this PR has been overtaken by the v5 codegen work. v5preview already solves the underlying need (#619: getting at the raw `*http.Response`) via the `Inspect()` method on every generated response type:

```
    resp, err := client.Indices.ForceMerge(ctx, req)
    if err != nil { ... }
    httpResp := resp.Inspect().Response  // *opensearch.Response (wraps *http.Response)
```

This pattern is consistent across all 1000+ generated operations in `v5preview/opensearchapi/` and avoids the breaking change to the two-value return signature. Once v5preview moves to top-level `opensearchapi/` it'll ship as the standard way to access the underlying response.

Given the breaking surface here and the v5 alternative, I'd like to close this PR. The v5 `Inspect()` mechanism should give you everything you need.  Let us know on #619 if it doesn't cover your use case and we'll reopen the discussion.

Thanks again for the work and the patience.

Issue #626, opened 2024-10-01
[BUG] Item OnFailure Callbacks Not Executed When Bulk Request Fails


kellen-miller:
### Is your feature request related to a problem?
I have been using the bulk indexing API and noticed that in the event a bulk request fails to reach OpenSearch, it is impossible to know which items were in that bulk request. Even if a failure callback is registered to an item, it is not executed. Only the bulk indexer's `OnError` function is executed, and the items are silently flushed from the worker's buffer.

### What solution would you like?
This results in the sender not knowing if an item successfully made it to Opensearch or not, which can result in data loss. I would like that on a bulk request failure, if an item in the buffer has a failure callback registered to it, it is executed and passed the bulk indexer error. I can open a PR if needed.

Specifically this error check:
https://github.com/opensearch-project/opensearch-go/blob/main/opensearchutil/bulk_indexer.go#L505-L511

Would become:

```go
func (w *worker) flush(ctx context.Context) error {
	// ...
	
	blk, err = w.bi.config.Client.Bulk(ctx, req)
	if err != nil {
		return w.handleBulkError(ctx, fmt.Errorf("flush: %w", err))
	}
	
	// ...
}
```

```go
func (w *worker) handleBulkError(ctx context.Context, err error) error {
	atomic.AddUint64(&w.bi.stats.numFailed, uint64(len(w.items)))

	if w.bi.config.OnError != nil {
		w.bi.config.OnError(ctx, err)
	}

	for i := range w.items {
		if w.items[i].OnFailure != nil {
			w.items[i].OnFailure(ctx, w.items[i], opensearchapi.BulkRespItem{}, err)
		}
	}

	return err
}
```

### What alternatives have you considered?
An external inbox pattern could be used, but that adds a lot of complexity when a simple code change could be used to mitigate this.

### Do you have any additional context?
Implementing this feature would allow developers to handle item-level failures more effectively, especially in applications where data integrity is paramount. It ensures that each item's failure is acknowledged and can be retried or logged as needed.

kellen-miller:
Also unrelated, but this comment is wrong and should probably be removed. It sets the `op` and `info` to the last key and value iterated through in the map, not the first.

https://github.com/opensearch-project/opensearch-go/blob/main/opensearchutil/bulk_indexer.go#L521-L527

dblock:
This looks like a bug. Want to try fixing it? Or even a failing test would be very useful.

kellen-miller:
[...]

PR #627, opened 2024-10-02
Execute item failure callback execution on bulk request error

kellen-miller:
### Description
Execute all bulk item failure callbacks (if they exist) in worker's buffer when a bulk request fails.

### Issues Resolved
Fixes https://github.com/opensearch-project/opensearch-go/issues/626

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - Fix ISM Allocation field types ([#609](https://github.com/opensearch-project/opensearch-go/pull/609))
 - Fix ISM Error Notification types ([#612](https://github.com/opensearch-project/opensearch-go/pull/612))
 - Fix signer receiving drained body on retries ([#620](https://github.com/opensearch-project/opensearch-go/pull/620))
+- Fix Bulk Index Items not executing failure callbacks on bulk request failure ([#626](https://github.com/opensearch-project/opensearch-go/issues/626))
 
 ### Security
 
--- a/opensearchutil/bulk_indexer.go
+++ b/opensearchutil/bulk_indexer.go
@@ -338,7 +338,8 @@ func (w *worker) run() {
 			w.mu.Lock()
 
 			if w.bi.config.DebugLogger != nil {
-				w.bi.config.DebugLogger.Printf("[worker-%03d] Received item [%s:%s]\n", w.id, item.Action, item.DocumentID)
+				w.bi.config.DebugLogger.Printf("[worker-%03d] Received item [%s:%s]\n", w.id, item.Action,
+					item.DocumentID)
 			}
 
 			if err := w.writeMeta(item); err != nil {
@@ -503,11 +504,7 @@ func (w *worker) flush(ctx context.Context) error {
 
 	blk, err = w.bi.config.Client.Bulk(ctx, req)
 	if err != nil {
-		atomic.AddUint64(&w.bi.stats.numFailed, uint64(len(w.items)))
-		if w.bi.config.OnError != nil {
-			w.bi.config.OnError(ctx, fmt.Errorf("flush: %w", err))
-		}
-		return fmt.Errorf("flush: %w", err)
+		return w.handleBulkError(ctx, fmt.Errorf("flush: %w", err))
 	}
 
 	for i, blkItem := range blk.Items {
@@ -520,7 +517,7 @@ func (w *worker) flush(ctx context.Context) error {
 		item = w.items[i]
 		// The OpenSearch bulk response contains an array of maps like this:
 		//   [ { "index": { ... } }, { "create": { ... } }, ... ]
-		// We range over the map, to set the first key and value as "op" and "info".
+		// We range over the map, to set the last key and value as "op" and "info".
 		for k, v := range blkItem {
 			op = k
 			info = v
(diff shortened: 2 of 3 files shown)


kellen-miller:
I ordered the execution of the item failure callback before the execution of the bulk indexers `OnError` callback, but am open to feedback on if that should be reversed. My thought process was that a user of the library may have some code to shutdown/panic their application in the `OnError` callback, which would result in the item `OnFailure` callbacks to not be executed.

dblock (approved):
Looks good! 

Let's take this opportunity and add a few more tests to cover the edge cases (some may exist)?

- fails only some items and makes sure that `failedItems` is correct 
- checks that OnError is called every time, and only once, if there's at least one failure

dblock (changes requested):
Actually looking at the code I see https://github.com/opensearch-project/opensearch-go/pull/627/files#diff-84320628b2df1ee872c513a0e5a15df089c6116a5c2b6e8d4d577102e0bcb6c6R526 that calls `onFailure` below if the response is > 201. 

Something seems fishy ...

kellen-miller:
Correct that second `OnFailure` is called when the entire bulk request doesn't fail. 

But if the whole bulk request errors, for example if the OpenSeach goes down while the request is in flight or the context deadlines, the flush function immediately returns before executing the failure callbacks.

dblock:
> Correct that `OnFailure` below if the response is > 201 is called when the bulk request is successful and an individual item in the request fails, say OpenSearch had an indexing failure with the item.
> 
> But if the whole bulk request errors, for example if OpenSearch goes down while the request is in flight, or the context deadlines, the flush function returns without executing the item failure callbacks and the workers buffer is cleared. So those items are lost and their callbacks are never executed.

Understood. And just to double-check, we don't call those callbacks twice in the successful response but some failed items case? Add a test for this?

kellen-miller:
> Understood. And just to double-check, we don't call those callbacks twice in the successful response but some failed items case? Add a test for this?

Yep the failure callback will only be executed once. If the response is successful (the [error here is nil](https://github.com/kellen-miller/opensearch-go/blob/bugfix/execute-item-fail-callbacks-bulk-error/opensearchutil/bulk_indexer.go#L505)), the code in this PR is never hit. 

My unit test I added has a check to make sure that the number of times the callback is what is expected i.e. nothing is called twice.
https://github.com/kellen-miller/opensearch-go/blob/bugfix/execute-item-fail-callbacks-bulk-error/opensearchutil/bulk_indexer_internal_test.go#L833

dblock:
> > Understood. And just to double-check, we don't call those callbacks twice in the successful response but some failed items case? Add a test for this?
> 
> Yep the failure callback will only be executed once. If the response is successful (the [error here is nil](https://github.com/kellen-miller/opensearch-go/blob/bugfix/execute-item-fail-callbacks-bulk-error/opensearchutil/bulk_indexer.go#L505)), the code in this PR is never hit.
> 
> My unit test I added has a check to make sure that the number of times the failure callback is executed is what is expected i.e. nothing is called twice. https://github.com/kellen-miller/opensearch-go/blob/bugfix/execute-item-fail-callbacks-bulk-error/opensearchutil/bulk_indexer_internal_test.go#L833

I think there's still a test missing for partial failure on some of the items, you want to check that if you have 5 items and 2 failures the callback is only called twice, for example?

kellen-miller:
> I think there's still a test missing for partial failure on some of the items, you want to check that if you have 5 items and 2 failures the callback is only called twice, for example?

I think this unit test earlier in the file already covers that?
https://github.com/kellen-miller/opensearch-go/blob/bugfix/execute-item-fail-callbacks-bulk-error/opensearchutil/bulk_indexer_internal_test.go#L277

Specifically this check here:
https://github.com/kellen-miller/opensearch-go/blob/bugfix/execute-item-fail-callbacks-bulk-error/opensearchutil/bulk_indexer_internal_test.go#L391-L398

I might be misunderstanding though, I'm happy to add more if that's not the test case you're referring too.

dblock:
> I might be misunderstanding though, I'm happy to add more if that's not the use case you're referring too.

First, thank you for hanging in here with me ;)

The old code did not call `item.OnFailure(ctx, item, info, err)` in this scenario at all, but did increment the number of fails (and then that was tested). So those tests with only some item callbacks also should check how many times `item.OnFailure(ctx, item, info, err)` has been invoked and that it was invoked with all the right items.

Without trying, some examples where I think all tests pass but the code is broken: call `item.OnFailure(ctx, **the first item every time**, info, err)`. With first 2/5 items failing call `onFailure` 5 times.

kellen-miller:
> The old code did not call item.OnFailure(ctx, item, info, err) in this scenario at all, but did increment the number of fails (and then that was tested). So those tests with only some item callbacks also should check how many times item.OnFailure(ctx, item, info, err) has been invoked and that it was invoked with all the right items.

> Without trying, some examples where I think all tests pass but the code is broken: call item.OnFailure(ctx, **the first item every time**, info, err). Or, with first 2/5 items failing call onFailure 5 times.

Ah gotcha gotcha, sorry for some much back and forth 😄.  Let me update the other tests to check for this.

dblock:
> Ah gotcha gotcha, sorry for some much back and forth 😄. Let me update the other tests to check for this.

No stress! I enjoy this type of conversations because I am super [OCD about tests](https://github.com/apache/lucene/pull/711#issuecomment-1055853457) :)

dblock (approved):
👏

kellen-miller:
@dblock sorry to bother again! Is there any roadmap or release cadence to this library? Just wondering when the release process with my fix will happen. I can always target `@main` in the mean time, just like to have tagged version set.

dblock:
Open a "release v. Next" issue and we'll take care of it quickly.

PR #635, opened 2024-10-17
Add the Total fields to the SearchResp struct

rjsadow:
### Description
The opensearch Search [reference documentation](https://opensearch.org/docs/latest/api-reference/search/) give some information on the `track_total_hits` parameter but doesn't include the full response body when it's used. `track_total_hits` is useful when you're trying to get a count of all hits in a given index for a given query without using the scroll api or some other traching mechinism. 

This PR adds the return values structure for the `track_total_hits` parameter into the SearchResp structure. 


rjsadow:
I did mean to close it. I was incorrect and I believe the SDK already handles this case.

Issue #653, opened 2025-01-16
[BUG] Highlight results are not returned on search results.


tim-holdaway-clever:
### What is the bug?
When calling `opensearchapi.Client.Search` with a query with a [highlight](https://opensearch.org/docs/latest/search-plugins/searching-data/highlight/) specified, the highlight results coming back in the raw json coming back from the OpenSearch cluster are not returned on the `SearchHit`s in the `SearchResp`.

### How can one reproduce the bug?
Call [opensearchapi.Client.Search()](https://github.com/opensearch-project/opensearch-go/blob/main/opensearchapi/api_search.go#L21) with a `SearchReq` with a body with a `highlight` property specifying highlights for fields that match in they query.

### What is the expected behavior?
The [SearchResp](https://github.com/opensearch-project/opensearch-go/blob/main/opensearchapi/api_search.go#L65) coming back from `Client.Search` should include the highlight results in the [SearchHit](https://github.com/opensearch-project/opensearch-go/blob/main/opensearchapi/api_search.go#L90)s, as they are returned in the raw json api response from the OpenSearch cluster.

### What is your host/environment?
using `opensearch-go@v4.3.0` on mac os 14.7.2 and running in aws lambda.
querying Amazon OpenSearch Service 2.15 cluster.

### Do you have any additional context?
It seems like the `highlight` field is already coming back in the raw json, and just needs to be added to [SearchHit](https://github.com/opensearch-project/opensearch-go/blob/main/opensearchapi/api_search.go#L90-L102) so that it will be populated when the json response is [unmarshalled](https://github.com/opensearch-project/opensearch-go/blob/main/opensearch.go#L270), like this:
```
 type SearchHit struct {
        Index       string                  `json:"_index"`
        ID          string                  `json:"_id"`
        Routing     string                  `json:"_routing"`
        Score       float32                 `json:"_score"`
        Source      json.RawMessage         `json:"_source"`
        Fields      json.RawMessage         `json:"fields"`
        Type        string                  `json:"_type"` // Deprecated field
        Sort        []any                   `json:"sort"`
        Explanation *DocumentExplainDetails `json:"_explanation"`
        SeqNo       *int                    `json:"_seq_no"`
        PrimaryTerm *int                    `json:"_primary_term"`
+       Highlight  map[string][]string     `json:"highlight"`
 }
```
I'm happy to investigate this further and PR it, if it's not already in the works to add this field to the response.

tim-holdaway-clever:
[...]

PR #654, opened 2025-01-16
Add Highlight field to SearchHit struct

tim-holdaway-clever:
### Description
For queries specifying [highlights](https://opensearch.org/docs/latest/search-plugins/searching-data/highlight/), the the `SearchHit`s on the `SearchResp` now include the highlight information returned in the api response.

### Issues Resolved
Closes https://github.com/opensearch-project/opensearch-go/issues/653

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -207,7 +207,7 @@ godoc: ## Display documentation for the package
 	godoc --http=localhost:6060 --play
 
 cluster.build:
-	docker compose --project-directory .ci/opensearch build;
+	docker compose --project-directory .ci/opensearch build --pull;
 
 cluster.start:
 	docker compose --project-directory .ci/opensearch up -d;
--- a/opensearchapi/api_nodes-stats.go
+++ b/opensearchapi/api_nodes-stats.go
@@ -112,6 +112,7 @@ type NodesStats struct {
 	Repositories                   []json.RawMessage                        `json:"repositories"`
 	AdmissionControl               NodesStatsAdmissionControl               `json:"admission_control"`
 	Caches                         NodesStatsCaches                         `json:"caches"`
+	RemoteStore                    NodeStatsRemoteStore                     `json:"remote_store"`
 }
 
 // NodesStatsIndices is a sub type of NodesStats representing Indices information of the node
@@ -728,4 +729,9 @@ type NodesStatsCaches struct {
 		ItemCount   int    `json:"item_count"`
 		StoreName   string `json:"store_name"`
 	} `json:"request_cache"`
+}
+
+// NodeStatsRemoteStore is a sub type of NodesStats
+type NodeStatsRemoteStore struct {
+	LastSuccessfulFetchOfPinnedTimestamps int `json:"last_successful_fetch_of_pinned_timestamps"`
 }
--- a/opensearchapi/api_search-template.go
+++ b/opensearchapi/api_search-template.go
@@ -62,6 +62,7 @@ type SearchTemplateResp struct {
 	Took    int            `json:"took"`
 	Timeout bool           `json:"timed_out"`
 	Shards  ResponseShards `json:"_shards"`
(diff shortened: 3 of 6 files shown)


tim-holdaway-clever:
Failing tests
```
        --- FAIL: TestNodes/Stats/without_request (0.09s)
    --- FAIL: TestSearchTemplate/with_request (0.01s)
```
 appear to also be failing in other recent PRs, for instance in https://github.com/opensearch-project/opensearch-go/actions/runs/12757766943/job/35558719849?pr=652, so seems like are probably not caused by this PR.
 
Failures look like
```
        	Error:      	Should be empty, but was [{"op":"remove","path":"/nodes/W0IruYOMTB2U8Ho0B6ZF4w/remote_store"}]
```
Which seems like maybe the version of OpenSearch cluster that the integration tests are running against may have changed and is now returning some extra unexpected property in the response, or something like that.

dblock:
> Which seems like maybe the version of OpenSearch cluster that the integration tests are running against may have changed and is now returning some extra unexpected property in the response, or something like that.

Appreciate it if you could help fix those first, if you have time.

tim-holdaway-clever:
> Appreciate it if you could help fix those first, if you have time.

I can take a look and see if I can reproduce it locally. Integration tests were all passing in my local run using `make cluster.build cluster.start &&  make test-integ race=true`, so seems like something in the CI build potentially using against a different version of the cluster or configuration or something, especially since they are failing on other recent PRs. 

Do you have any hunches or pointers on where I should look, as I'm not really familiar yet with all the ins-and-outs of this projects testing setup and CI config?

dblock:
> Do you have any hunches or pointers on where I should look, as I'm not really familiar yet with all the ins-and-outs of this projects testing setup and CI config?

I do not. I am mostly just cheerleading in this repo vs. doing real work which is primarily @Jakob3xD and some other contributors more recently :)

opensearchapi/api_search-template.go
```diff
@@ -62,6 +62,7 @@ type SearchTemplateResp struct {
 	Took    int            `json:"took"`
 	Timeout bool           `json:"timed_out"`
 	Shards  ResponseShards `json:"_shards"`
+	Status  int            `json:"status"`
```
tim-holdaway-clever, line 65: 🗒️  Integration tests were failing due to this field that was added to the opensearch 2.18 response. The test was comparing the parsed response to the raw response and failing on the missing field.

opensearchapi/api_nodes-stats.go
```diff
@@ -112,6 +112,7 @@ type NodesStats struct {
 	Repositories                   []json.RawMessage                        `json:"repositories"`
 	AdmissionControl               NodesStatsAdmissionControl               `json:"admission_control"`
 	Caches                         NodesStatsCaches                         `json:"caches"`
+	RemoteStore                    NodeStatsRemoteStore                     `json:"remote_store"`
```
tim-holdaway-clever, line 115: 🗒️  Integration tests were failing due to this field that was added to the opensearch 2.18 response. The test was comparing the parsed response to the raw response and failing on the missing field.
See https://github.com/opensearch-project/OpenSearch/pull/15611 for where this was added to 2.18

Makefile
```diff
@@ -207,7 +207,7 @@ godoc: ## Display documentation for the package
 	godoc --http=localhost:6060 --play
 
 cluster.build:
-	docker compose --project-directory .ci/opensearch build;
+	docker compose --project-directory .ci/opensearch build --pull;
```
tim-holdaway-clever, line 210: 🗒️  This ensures that when running integration tests locally, docker isn't using an older cached version of `opensearch:latest`, and actually gets the most recent one that the CI build will be using.

dblock (approved):

dblock:
Thanks for fixing this!

PR #667, opened 2025-02-12
Lock version of OpenSearch for integration tests.

dblock:
### Description

Lock version of OpenSearch for integration tests. When 2.19 was released it broke CI, which is not desirable. With this change we have to upgrade deliberately like we do for integration tests already.

### DIFF
--- a/.github/workflows/test-integration.yml
+++ b/.github/workflows/test-integration.yml
@@ -4,6 +4,7 @@ on: [push, pull_request]
 
 env:
   GITHUB_ACTIONS: true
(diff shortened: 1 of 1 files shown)


Jakob3xD:
@dblock https://github.com/opensearch-project/opensearch-go/pull/669 fixes the CI issue. Do you still want to lock the version?

dblock:
> @dblock #669 fixes the CI issue. Do you still want to lock the version?

We should lock the version to avoid getting randomized, yes. Want to do it part of #669?

Jakob3xD (approved):

Issue #684, opened 2025-03-18
[FEATURE] Add DataStream field into GET indices response


SlavaUtesinov:
### Is your feature request related to a problem?
I can't get data stream name from get indices response

### What solution would you like?
Addition of DataStream field into [IndicesGetResp](https://github.com/opensearch-project/opensearch-go/blob/v4.4.0/opensearchapi/api_indices-get.go#L39 )

### What alternatives have you considered?
Unmarshall raw HTTP response into custom user defined type

### Do you have any additional context?
```
type IndicesGetResp struct {
	Indices map[string]struct {
                 // DataStream *string `json:"data_stream"` <-- add this field
		Aliases  map[string]struct{} `json:"aliases"`
		Mappings json.RawMessage     `json:"mappings"`
		Settings json.RawMessage     `json:"settings"`
	}
	response *opensearch.Response
}
```

rehanpfmr:
Can this be assigned to me?

andrross:
[Catch All Triage](https://github.com/opensearch-project/.github/blob/main/TRIAGING.md) - [1](https://github.com/andrross)

@rehanpfmr I've assigned this to you. Thanks!

Jakob3xD:
Closed by https://github.com/opensearch-project/opensearch-go/pull/701

PR #689, opened 2025-04-10
[FEATURE] Add DataStream field into GET indices response #684

rehanpfmr:
### Description
Add DataStream field into GET indices response

### Issues Resolved
Closes [https://github.com/opensearch-project/opensearch-go/issues/684]._ 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ## [Unreleased]
+- Adds `new field to the struct to support data streams` ([#684](https://github.com/opensearch-project/opensearch-go/issues/684)) 
 ### Dependencies
 ### Added
 - Adds `InnerHits` field to `SearchResp` ([#672](https://github.com/opensearch-project/opensearch-go/pull/672))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ## [Unreleased]
-- Adds `new field to the struct to support data streams` ([#684](https://github.com/opensearch-project/opensearch-go/issues/684)) 
+- Adds DataStream field to IndicesGetResp struct ([#684](https://github.com/opensearch-project/opensearch-go/issues/684))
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ## [Unreleased]
+- Adds `new field to the struct to support data streams` ([#684](https://github.com/opensearch-project/opensearch-go/issues/684))
```
Jakob3xD, line 6:
```suggestion
- Adds DataStream field to IndicesGetResp struct ([#684](https://github.com/opensearch-project/opensearch-go/issues/684)) 
```

opensearchapi/api_indices-get.go
```diff
@@ -37,6 +37,7 @@ func (r IndicesGetReq) GetRequest() (*http.Request, error) {
 // IndicesGetResp represents the returned struct of the get indices response
 type IndicesGetResp struct {
 	Indices map[string]struct {
+                DataStream *string `json:"data_stream"`
 		Aliases  map[string]struct{} `json:"aliases"`
 		Mappings json.RawMessage     `json:"mappings"`
 		Settings json.RawMessage     `json:"settings"`
```
Jakob3xD, line 43: The linter complains about the spacing.
It should look something like following:
```suggestion
                DataStream *string             `json:"data_stream"`
		Aliases    map[string]struct{} `json:"aliases"`
		Mappings   json.RawMessage     `json:"mappings"`
		Settings   json.RawMessage     `json:"settings"`
```

Jakob3xD (changes requested):
Thanks for the PR.
Can you please adjust the indices test in a way, that a DataStream gets created before running the test so we can validate that the field gets parsed.
To do so, you need to edit the `opensearchapi/api_indices_test.go` file.

The DataStream itself gets already tested in `opensearchapi/api_datastream_test.go`so it should be enough to create it in the top section after the `client.Indices.Delete()` call and by adding a t.Cleanup() call to delete the DataStream.

opensearchapi/api_indices-get.go
rehanpfmr: Sure, i will have the lint fixed. thanks for looking into this

PR #693, opened 2025-04-15
Fixing description to the policy

### LINKED ISSUE #683: [BUG] Can't create policy with empty description
### What is the bug?
I can't create policy with empty description

Create policy with empty description

Policy with empty description created

```
type PolicyBody struct {
	PolicyID          string                   `json:"policy_id,omitempty"`
	Description       string                   `json:"description,omitempty"`
	LastUpdatedTime   int64                    `json:"last_updated_time,omitempty"`
	SchemaVersion     int                      `json:"schema_version,omitempty"`
	ErrorNotification *PolicyErrorNotification `json:"error_notification"`
	DefaultState      string                   `json:"default_state"`
	States            []PolicyState            `json:"states"`
	Template          []Template               `json:"ism_template,omitempty"`
}
```
[Description](https://github.com/opensearch-project/opensearch-go/blob/v4.4.0/plugins/ism/api_policies.go#L72 ) has omitempty tag, that is why empty string value will lead to the absence of this field at json.

rehanpfmr:
### Description
_Describe what this change achieves._

### Issues Resolved
Closes #683 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ## [Unreleased]
+- Fix `Makes the Description field mandatory in policy creation` ([#683](https://github.com/opensearch-project/opensearch-go/pull/683))
 ### Dependencies
 ### Added
 - Adds `InnerHits` field to `SearchResp` ([#672](https://github.com/opensearch-project/opensearch-go/pull/672))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ## [Unreleased]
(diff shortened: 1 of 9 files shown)


Jakob3xD:
Removing the omitempty option results in description being always present. Instead of removing it, we might want to consider changing the type to a pointer string.

Jakob3xD (changes requested):
Removing the omitempty option results in description being always present. Instead of removing it, we might want to consider changing the type to a pointer string.

Jakob3xD:
As there is no activity for this PR, I am closing it.

sean-:
Thanks for reporting this. PR #804 addresses this through the new spec-driven code generator (`osapi/` package). The generated types use `*string` for optional fields, which gives callers explicit control:

- `nil` -> field omitted from JSON (not set)
- `ptr("")` -> field serialized as `"description": ""` (explicitly empty)

See the ISM policy type here: https://github.com/sean-/opensearch-go/blob/3f4e0976ecd91e67f2584ccb28ee8e6fa7d63135/osapi/types_gen.go#L8758

PR #699, opened 2025-05-07
Improve error response when body is an empty string

agnivade:
### Description
We are observing conditions where Opensearch returns with an emtpy body. In that case, the error string returned is just blank which is very confusing to the end user.

To fix this, we add some more context to the error string to make it clear.

### DIFF
--- a/error_test.go
+++ b/error_test.go
@@ -244,7 +244,7 @@ func TestError(t *testing.T) {
 			}
 			assert.True(t, resp.IsError())
 			err := opensearch.ParseError(resp)
-			assert.Contains(t, err.Error(), http.StatusText(http.StatusUnauthorized))
+			assert.True(t, errors.Is(err, opensearch.ErrJSONUnmarshalBody))
 		})
 		t.Run("too many requests", func(t *testing.T) {
 			resp := &opensearch.Response{
@@ -253,7 +253,7 @@ func TestError(t *testing.T) {
 			}
 			assert.True(t, resp.IsError())
 			err := opensearch.ParseError(resp)
-			assert.Contains(t, err.Error(), "429 Too Many Requests /testindex/_bulk")
+			assert.True(t, errors.Is(err, opensearch.ErrJSONUnmarshalBody))
 		})
 	})
 }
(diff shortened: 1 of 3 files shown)


error_test.go
```diff
@@ -244,7 +244,7 @@ func TestError(t *testing.T) {
 			}
 			assert.True(t, resp.IsError())
 			err := opensearch.ParseError(resp)
-			assert.Equal(t, err.Error(), http.StatusText(http.StatusUnauthorized))
+			assert.Contains(t, err.Error(), http.StatusText(http.StatusUnauthorized))
```
Jakob3xD, line 247: Please adjust the tests to use Equal, so we expect a 1 to 1 match and not only part of the string.
```suggestion
			assert.Equal(t, err.Error(), fmt.Sprintf("invalid body: %q", http.StatusText(http.StatusUnauthorized))
```

error_test.go
```diff
@@ -253,7 +253,7 @@ func TestError(t *testing.T) {
 			}
 			assert.True(t, resp.IsError())
 			err := opensearch.ParseError(resp)
-			assert.Equal(t, err.Error(), "429 Too Many Requests /testindex/_bulk")
+			assert.Contains(t, err.Error(), "429 Too Many Requests /testindex/_bulk")
```
Jakob3xD, line 256: Same
```suggestion
			assert.Equal(t, err.Error(), fmt.Sprintf("invalid body: %q", "429 Too Many Requests /testindex/_bulk"))
```

Jakob3xD (changes requested):
Please rebase and verify your signature

error_test.go
agnivade: Checking for equality in error strings is actually an anti-pattern because they are not part of the API, as clearly shown by this failure. If we really want to check for errors, they should be done via errors.Is or errors.As. Do you still want me to change to `assert.Equal`?

Jakob3xD:
I have reevaluated the error handling and the best solution would is to remove the `json.Valid` check, because it gets called in the Unmarshal function. This also reduces the risk of logging secure information that is present in the body.

Could you please remove the check and adjust the tests. It should then be possible to check the error against `ErrJSONUnmarshalBody`. Thank you.

Jakob3xD (approved):

PR #705, opened 2025-05-27
fix(api): alias api creates index when index name is nill

### LINKED ISSUE #650: [BUG] PUT Alias creates index and not an alias
### What is the bug?
When creating aliases it returns successful but actually creates an index instead.

```
resp, err := b.openSearch.Indices.Alias.Put(context.Background(), opensearchapi.AliasPutReq{
		Alias:   alias,
	})
	b.logger.Info("Created alias", slog.Any("response", resp))
	return err
```

An alias is created

osx

rootxrishabh:
### Description
When Alias PUT api is called with an empty index name i.e `""`, the `BuildRequest` function parses characters after the double quote and before the next `/` as hostname for the request i.e `_alias` and request becomes `https://localhost:9200/{aliasname} -H _alias -XPUT` which actually creates an index. 

The fix returns an error when a nil index is passed.

Note -
This is intended to be a temporary fix, however a better fix would be to parse the complete URL i.e instead of creating http request just with path parameters (i.e `/<target>/_alias/<alias-name>`) at `BuildRequest`, it would be better to provide the complete URL (i.e `https://localhost:9200/<target>/_alias/<alias-name>`). This however, requires a redesign of the API system.

### Issues Resolved
#650 

### DIFF
--- a/opensearchapi/api_indices-alias.go
+++ b/opensearchapi/api_indices-alias.go
@@ -9,6 +9,7 @@ package opensearchapi
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"net/http"
 	"strings"
 
@@ -156,6 +157,9 @@ type AliasPutReq struct {
 
 // GetRequest returns the *http.Request that gets executed by the client
 func (r AliasPutReq) GetRequest() (*http.Request, error) {
+	if len(r.Indices) == 0 {
+		return nil, fmt.Errorf("at least one index must be specified for alias operation")
+	}
 	indices := strings.Join(r.Indices, ",")
 
 	var path strings.Builder


Jakob3xD:
I am not sure yet how I think about it. I generally understand that you want to avoid users of the lib from running into errors or wrong behaviours. But I personally don't think that the lib is responsible for this, as there are many more cases where users can run into such errors or wrong behaviours. Moreover, such special error prevention is harder to handle in code generation. What do you think @dblock ?

sean-:
Thanks for digging into this, @rootxrishabh. Status update: #804 (merged 2026-05-28) moved URL-path construction into a generated `internal/path` package. The new `IndicesPutAliasPath.Build()` uses `hasNonEmpty()` to skip empty index slices, so `AliasPutReq{Indices: []string{""}, Alias: "foo"}` now routes to `/_alias/foo` rather than emitting a malformed URL.

Could you re-run your repro from #650 against current `main` and let us know whether the original bug still reproduces? If it's resolved, we'll close this in favor of #804. If there's still a problem case (e.g. with `Indices: nil` specifically, or some other edge), happy to look at it as a targeted fix.

PR #769, opened 2026-01-13
Fix node discovery logic to properly handle dedicated cluster managers

### LINKED ISSUE #765: [BUG] Node discovery incorrectly includes master nodes with remote_cluster_client role
### What is the bug?
The DiscoverNodes() function in opensearchtransport/discovery.go only filters out master/cluster_manager nodes when that is their only role. However, master nodes often have additional roles like remote_cluster_client (enabled by default), causing them to bypass the filter and be added to the connection pool.

1. Configure an OpenSearch cluster with dedicated master nodes that have roles ["master", "remote_cluster_client"] (the default configuration)
2. Create a Go client with DiscoverNodesOnStart: true
3. Observe that master nodes are added to the connection pool and receive client traffic

Master/cluster_manager nodes should be excluded from the connection pool if they do not have a data role, regardless of what other roles they may have.

- OpenSearch version: 2.19.2
- opensearch-go version: v4
- OS: Linux (Kubernetes)

N/A

Current filter logic at discovery.go:75:
if len(node.Roles) == 1 && (node.Roles[0] == "master" || node.Roles[0] == "cluster_manager") {

This misses nodes like:
{"name": "master-1", "roles": ["master", "remote_cluster_client"]}

I have a fix ready that checks for master/cluster_manager role without a data role, which is more future-proof.

sean-:
### Summary

Fixes dedicated cluster manager detection by implementing proper role validation and upstream-compatible node filtering logic. Addresses the core issue where nodes with only cluster management roles were
incorrectly being included in the connection pool for client requests.

### Key Features

- Proper role constants matching OpenSearch server definitions
- `isDedicatedClusterManager()` logic matching Java client `NodeSelector.SKIP_DEDICATED_CLUSTER_MASTERS`
- Comprehensive role validation with master role deprecation warnings
- Warm+search role conflict detection for OpenSearch 3.0+ compatibility
- Remote cluster client role support for cross-cluster operations
- `IncludeDedicatedClusterManagers` configuration option for backward compatibility

### Breaking Change

**BEHAVIOR CHANGE**: Dedicated cluster manager nodes are now excluded from client request routing by default to match upstream Java client behavior and follow best practices.

- **Before**: All discovered nodes (including dedicated cluster managers) received client requests
- **After**: Only nodes with "work" roles (data, ingest, warm, search) receive client requests by default
- **Migration**: Set `IncludeDedicatedClusterManagers: true` in transport config to preserve legacy behavior

Fixes #765

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,11 +6,15 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ### Added
 - Enhanced cluster readiness checking for improved test reliability: `ostest.NewClient()` now includes readiness validation (health + cluster state + nodes info)
+- Configuration option `IncludeDedicatedClusterManagers` for controlling cluster manager node routing ([#765](https://github.com/opensearch-project/opensearch-go/issues/765))
 
 ### Changed
 - Refactor Client struct to use embedded mutex pattern for improved thread safety ([#775](https://github.com/opensearch-project/opensearch-go/pull/775))
 - Refactor metrics struct to use atomic counters for lock-free request/failure tracking ([#776](https://github.com/opensearch-project/opensearch-go/pull/776))
 - Test against Opensearch 2.19.4, 3.1, 3.3, and 3.4 ([#782](https://github.com/opensearch-project/opensearch-go/pull/782))
+- **BREAKING**: Enhanced node discovery to match OpenSearch server behavior ([#765](https://github.com/opensearch-project/opensearch-go/issues/765))
+  - Dedicated cluster manager nodes are now excluded from client request routing by default (best practice)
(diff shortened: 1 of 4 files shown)


Jakob3xD (commented):
In general, it LGTM.
Having the `IncludeDedicatedClusterManagers` option is good, and I like that you added tests for it.

I am missing the `ml` and `coordinating_only` role.
https://docs.opensearch.org/latest/install-and-configure/configuring-opensearch/configuration-system/

I am not sure about the handling of the search role. I am not aware of those antipatterns regarding the search role. Can you validate this and point me to some sources?

Jakob3xD (commented):
Only a small nit.
Rest LGTM
Thank you.

Jakob3xD (approved):

PR #797, opened 2026-03-11
Add Status to TaskGetResp structure

### LINKED ISSUE #788: [FEATURE] Add status to opensearchapi.TasksGetResp
The current TasksGetResp structure includes many, but not all fields from the JSON response.

Specifically there is no access to the `status` field of the task, which looks like:

```json
"tasks": {
        "JzrCxdtFTCO_RaINw8ckNA:54321": {
          ...
          "status": {
            "total": 1000,
            "created": 0,
            "updated": 0,
            "deleted": 0,
            "batches": 1,
            "version_conflicts": 0,
            "noops": 0,
            "retries": {
              "bulk": 0,
              "search": 0
            },
            ...
          },
        ...
        }
      }
```

The status is extremely useful because it tells you what actually happened in the task, e.g. for a delete query, how many documents were actually deleted. Without it, you know the task completed, but that's all - you don't know if it actually achieved its intended purpose.

I would like the status field from the JSON response to be added to the `TasksGetResp` struct.

None.

No

bobbyphilip:
### Description
Add Status to TaskGetResp structure. This is now similar to the java client

### Issues Resolved
Resolves #788


bobbyphilip:
Are there any tests which need to be updated? I only saw the integration test and that basically just checks if the API works, not the response values

opensearchapi/api_tasks-get.go
```diff
@@ -69,6 +69,22 @@ type TasksGetResp struct {
 				ActiveThreads    int `json:"active_threads"`
 			} `json:"thread_info"`
 		} `json:"resource_stats"`
+		Status struct {
```
rmoles-cs, line 72: Nice addition. Would be super valuable to add to all Resp structs IMO!

bobbyphilip: @rmoles-cs  any examples of which other structs are missing this?

VijayanB:
> Are there any tests which need to be updated? I only saw the integration test and that basically just checks if the API works, not the response values

Thank you for contribution. Do you see an opportunity to add unit test or update integration test to test this behavior?

sean-:
Thanks for the contribution @bobbyphilip! The `Status` field is a great addition.

One concern with the current approach: the `status` field in the task response is polymorphic, in that its shape depends on the task type. `reindex`, `delete_by_query`, and `update_by_query` tasks return the struct in this PR, but other task types (`replication`, `resync`, `persistent` tasks) return completely different structures. A fixed struct silently drops data for those cases and can mislead callers with zero-valued fields.

I've opened #803 as an alternative that:

- Uses `json.RawMessage` for the `Status` field (matching the existing pattern for `Headers`, `Aggregations`, `Failures`, etc.)
- Provides a typed `BulkByScrollTaskStatus` struct and `ParseBulkByScrollTaskStatus()` helper for the common reindex/delete_by_query/update_by_query case
- Adds the `Status` field to `TasksListTask` and `TaskCancelInfo` as well
- Includes integration test assertions, a tasks guide, and a compilable sample

This preserves all status data regardless of task type while still giving callers a convenient typed path for the bulk-by-scroll case.  Does this work for you?

bobbyphilip:
> Thanks for the contribution @bobbyphilip! The `Status` field is a great addition.
> 
> One concern with the current approach: the `status` field in the task response is polymorphic, in that its shape depends on the task type. `reindex`, `delete_by_query`, and `update_by_query` tasks return the struct in this PR, but other task types (`replication`, `resync`, `persistent` tasks) return completely different structures. A fixed struct silently drops data for those cases and can mislead callers with zero-valued fields.
> 
> I've opened #803 as an alternative that:
> 
>     * Uses `json.RawMessage` for the `Status` field (matching the existing pattern for `Headers`, `Aggregations`, `Failures`, etc.)
> 
>     * Provides a typed `BulkByScrollTaskStatus` struct and `ParseBulkByScrollTaskStatus()` helper for the common reindex/delete_by_query/update_by_query case
> 
>     * Adds the `Status` field to `TasksListTask` and `TaskCancelInfo` as well
> 
>     * Includes integration test assertions, a tasks guide, and a compilable sample
> 
> 
> This preserves all status data regardless of task type while still giving callers a convenient typed path for the bulk-by-scroll case. Does this work for you?

@sean-   
That looks good.  I will close this PR

PR #801, opened 2026-03-19
Add adaptive max_concurrent_shard_requests for search routing

### LINKED ISSUE #800: [PROPOSAL] Adaptive `max_concurrent_shard_requests` based on congestion control window
### Motivation

OpenSearch's `max_concurrent_shard_requests` parameter controls how many shard-level requests a coordinator node fans out in parallel during a search. The default is **5** — a static constant that does not scale with CPU count, thread pool size, or cluster load ([`SearchRequest.java:649`](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/action/search/SearchRequest.java#L649)).

This default is enforced via a semaphore-based throttle (`PendingExecutions` in `AbstractSearchAsyncAction`) that queues shard requests beyond the limit. In production, we've observed that this is one of the primary bottlenecks constraining coordinator-path throughput:

- With round-robin routing through coordinator nodes, active search threads on data nodes plateau at **~2** despite the search thread pool being sized at `(3 × processors / 2) + 1`.
- After deploying opensearch-go with direct-to-shard routing (bypassing the coordinator hop entirely), active search threads jumped to **~7** — a 3.5× increase in effective utilization.
- The delta is largely attributable to eliminating the `max_concurrent_shard_requests=5` bottleneck and the transport connection pool limits (`transport.connections_per_node.reg=6`).

However, not all search requests can bypass coordination. Multi-index scatter-gather queries, queries without a routing value, and queries where shard placement data is unavailable still go through the coordinator path. For these requests, `max_concurrent_shard_requests=5` remains the bottleneck.

### Proposal

The client should set `max_concurrent_shard_requests` adaptively per-request, derived from the AIMD congestion control window (`cwnd`) already tracked per-connection per-thread-pool in `pool_congestion.go`.

**Value derivation:**

```
max_concurrent_shard_requests = clamp(cwnd, floor=5, ceiling=searchThreadPoolSize)
```

Where:
- `cwnd` is the current congestion window for the selected connection's `search` pool
- `floor=5` matches the OpenSearch default — never more restrictive than stock behavior
- `ceiling` is the search thread pool size: `(3 × allocatedProcessors / 2) + 1`, available via the connection's `allocatedProcessors` field

This makes the two systems coherent: the client's congestion window (how many concurrent requests the client believes the pool can handle) directly informs the server's fan-out limit (how many shard requests the coordinator issues in parallel).

**AIMD behavior carries through:**

The congestion control uses a layered signal hierarchy where queue-based metrics are the primary feedback loop. In practice, cwnd converges to a sustainable level using queuing signals alone — rejections and 429s are a theoretical safety net, not an expected operating condition.

| Signal priority | Congestion signal | cwnd response | Effect on `max_concurrent_shard_requests` |
|---|---|---|---|
| 1 (primary) | **Queue wait time**: `wait_per_completed >= 1ms` (RESIZABLE pools like `search`) | Halves | Fan-out cuts based on queue pressure — fires well before thread pool saturates |
| 2 (fallback) | **Queue saturation**: `queue > 0 && active >= pool_size` (fixed-size pools) | Halves | Equivalent signal for pools that don't report wait-time metrics |
| 3 (no congestion) | Neither signal present, `cwnd < ssthresh` | Doubles (slow start) | Fan-out grows aggressively toward pool capacity |
| 4 (no congestion) | Neither signal present, `cwnd >= ssthresh` | +1 (congestion avoidance) | Fan-out grows linearly, probing for headroom |
| 5 (last resort) | **Rejection** (`delta(rejected) > 0`) | Halves + overloaded flag | Node removed from selection entirely — should not occur in practice because signals 1-2 reduce cwnd before the queue fills to rejection |

The `search` pool is RESIZABLE, so it reports `total_wait_time_in_nanos`. This means signal 1 (wait-time per completed request) is the active feedback mechanism for search. The 1ms threshold triggers multiplicative decrease while the queue is still short — typically hundreds of requests before the queue would fill to its 1,000-entry limit and begin rejecting. By the time cwnd stabilizes, `max_concurrent_shard_requests` reflects what the node can actually sustain without queuing pressure, and 429s / rejections should never be reached.

### Scope

**Apply when:**
- Request matches a search route (pool name `"search"`)
- Routing was NOT shard-exact (i.e., request will be coordinated server-side)
- Caller has not already set `max_concurrent_shard_requests` in the request URL

**Do not apply when:**
- Shard-exact routing was used (coordinator fan-out doesn't happen)
- Non-search requests (bulk, get, index, delete, etc.)
- Caller explicitly set the parameter (respect user override)

### Implementation sketch

1. Add `MaxConcurrentShardRequests int` to `NextHop`.
2. In `indexRouter.Eval()`: after selecting a connection, if the route is `"search"` and shard-exact was not used, compute the value from `conn.poolCwnd("search")` and set it on the `NextHop`.
3. In `Perform()`: if `NextHop.MaxConcurrentShardRequests > 0` and the request URL does not already contain the parameter, append `max_concurrent_shard_requests=N` as a query parameter.
4. Add a routing feature flag (`routingSkipAdaptiveConcurrency`) for opt-out. Default: enabled.
5. Include the computed value in `RouteEvent` / `RouteCandidate` for observability.

### Risks and mitigations

| Risk | Mitigation |
|---|---|
| Over-aggressive fan-out during slow-start | Capped at search thread pool size; AIMD will halve on first congestion signal |
| Interaction with server-side admission control | Server still enforces its own thread pool queue limits and circuit breakers independently |
| Unexpected load increase on clusters tuned for low utilization | Feature flag for opt-out; per-application rollout recommended |
| Stale cwnd after connection failover | cwnd resets to 1 on new connections (slow-start); conservative by default |

sean-:
- **Cluster-wide signal**: After each stats poll cycle, `clusterSearchAIMD` sums `delta(total_wait_time_in_nanos)` and `delta(completed)` across all nodes' search pools, then runs a single AIMD state machine. This correctly models data-node fan-out capacity rather than a single coordinator's local thread pool pressure.
- **Per-node AIMD unchanged**: Connection scoring (which coordinator to pick) still uses per-node cwnd. The cluster aggregate only drives MCSR (how wide to fan out).
- **Hot-shard resilience**: Single overloaded nodes are diluted by healthy peers in the aggregate — MCSR only drops when cluster-wide pressure rises. Hot shards are a data distribution problem handled by the server's shard allocator, not a client fan-out problem.
- **Configurable**: `WithAdaptiveConcurrency(bool)`, `WithAdaptiveConcurrencyLimits(min, max)`, `OPENSEARCH_GO_SHARD_REQUESTS` env var, or `OPENSEARCH_GO_ROUTING_CONFIG=-adaptive_concurrency`
- **Safe defaults**: Clamped to [5, 256], caller-set `max_concurrent_shard_requests` query params are never overwritten, not applied to shard-exact routed requests
- **Zero-alloc hot path**: `poolRouter.Eval` reads one `atomic.Int32` — benchmarks confirm 0 B/op across all router paths
- **Fallback**: Before the first poll cycle completes, falls back to per-node cwnd (existing behavior)
- **singleServerPool fix**: AIMD updates now proceed when discovery collapses to a single connection (e.g. behind a proxy)

Fixes: #800


sean-:
I can't publish the specifics, but I can show you when a team deployed these changes.  There's a clear before and after for when the team toggled the feature on (initially deployed with `OPENSEARCH_GO_ROUTING_CONFIG=false` -> removing this environment variable).  The increase in throughput is welcome, but it's worth noting that the uncorked throughput also comes at the expense of some latency, too.  It's too early to tell if the stand-out tail latency spikes have been permanently resolved or not, but I thought people would benefit from seeing these changes in a production setting (even if out of context).

image
image
image

opensearchtransport/pool_congestion_internal_test.go
```diff
@@ -584,3 +584,323 @@ func TestMaxCwndOrDefault(t *testing.T) {
...
+		var ca clusterSearchAIMD
+		c1 := newConn("n1")
+
+		// Poll 1: establish baseline.
+		ca.update([]nodeSearchSample{makeTestSample(c1, 100, 50000, 13)})
+
+		// Poll 2: completions increased, low wait time -> slow start.
+		ca.update([]nodeSearchSample{makeTestSample(c1, 200, 60000, 13)})
+		cwnd := ca.cwnd.Load()
+		require.Positive(t, cwnd, "cwnd should have grown from slow start")
```
Jakob3xD: The AIMD math is deterministic here: initial cwnd = 0 → `max(cwnd, 1) = 1` → slow start doubles to `min(1×2, 13) = 2` after this poll. `require.Positive` would pass even if the doubling logic were replaced by `cwnd += 1`.

```go
// Should be:
require.Equal(t, int32(2), ca.cwnd.Load(), "slow start should double: 1→2")
```

And on line 638, after poll 3 (`min(2×2, 13) = 4`):
```go
require.Equal(t, int32(4), ca.cwnd.Load(), "slow start should double: 2→4")
```

opensearchtransport/pool_congestion.go
```diff
@@ -262,3 +262,158 @@ func updatePoolCongestion(conn *Connection, threadPools map[string]ThreadPoolSta
...
+func (c *clusterSearchAIMD) update(polled []nodeSearchSample) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	if c.mu.nodes == nil {
+		c.mu.nodes = make(map[*Connection]clusterNodeSearchEpoch, len(polled))
+	}
+
+	// Build the set of connections polled this cycle for stale eviction.
+	polledSet := make(map[*Connection]struct{}, len(polled))
```
Jakob3xD, line 321: `make(map[*Connection]struct{})` allocates on every poll cycle solely for stale-entry eviction. With a typical cluster of 2–10 nodes, a linear O(n²) scan over `polled` and `c.mu.nodes` avoids the allocation entirely. Not the Eval hot path, but the package otherwise pools aggressively.

Jakob3xD (commented):
The adaptive MCSR design is sound: a cluster-wide AIMD window derived from aggregated search thread pool stats, clamped to a configurable [min, max] range and injected transparently as a query parameter. Two items to address before merging.

> **Transparency note:** This review was created with the assistance of Claude (Anthropic) for transparency.

---

### Test gap: `TestClusterSearchAIMD` "slow start doubles cwnd" uses imprecise assertions

The AIMD algorithm is fully deterministic for these inputs. Starting from cwnd = 0, `max(cwnd, 1) = 1`; slow start produces `min(1×2, 13) = 2` after poll 2, and `min(2×2, 13) = 4` after poll 3. The current assertions (`require.Positive` and `require.GreaterOrEqual`) would pass even if the doubling logic were replaced by additive increment. See inline comment.

---

### Minor: `polledSet` heap-allocates on every poll cycle in `clusterSearchAIMD.update`

Every stats poll cycle, `update()` does `make(map[*Connection]struct{}, len(polled))` solely to find stale entries for eviction. With a typical cluster of 2–10 nodes, a simple O(n²) nested loop over `polled` and `c.mu.nodes` would avoid the allocation entirely. This is the background stats goroutine rather than the Eval hot path, so the impact is low — but the package otherwise pools buffers aggressively, so this stands out.

---

### Nit: MCSR skip on shard-exact path is undocumented

When `calcSingleKeyCost` finds candidates, the function returns before the MCSR computation block — intentional (fan-out = 1), but not explained in the production code. A one-line comment on the early-return would prevent readers from wondering whether MCSR was accidentally omitted. (The test at line 867 of `policy_pool_router_internal_test.go` already covers this correctly.)

---

### Bug: `prevWaitTimeNano` resets to 0 when `TotalWaitTimeInNanos` is nil — spurious AIMD delta spike

`clusterSearchAIMD.update` builds `newEpoch` without carrying `prevWaitTimeNano` when `TotalWaitTimeInNanos == nil`:

```go
newEpoch := clusterNodeSearchEpoch{
    prevCompleted: s.stats.Completed,
    hasWaitTime:   epoch.hasWaitTime,
    // prevWaitTimeNano NOT copied
}
if s.stats.TotalWaitTimeInNanos != nil {
    newEpoch.prevWaitTimeNano = *s.stats.TotalWaitTimeInNanos
```

On the next poll where `TotalWaitTimeInNanos` is non-nil, line 363 computes `*TotalWaitTimeInNanos - epoch.prevWaitTimeNano` against 0 (the stale epoch), producing an artificially inflated `totalDeltaWait`. This triggers an unwarranted multiplicative decrease and can permanently suppress MCSR on a cluster that returns mixed nil/non-nil wait-time stats.

Fix: copy `prevWaitTimeNano` unconditionally in the struct literal, or add `newEpoch.prevWaitTimeNano = epoch.prevWaitTimeNano` outside the nil-guard.

https://github.com/opensearch-project/opensearch-go/blob/5e763cb638a3425f6235476194a1ac5b595821f5/opensearchtransport/pool_congestion.go#L339-L364

---

### Bug: bare `debugLogger` reads re-introduce the data race fixed in main at `7d511356`

Main merged commit `7d511356` which wrapped every `debugLogger` access behind an atomic `loadDebugLogger()` accessor to eliminate a data race between the writer (`SetLogger`) and concurrent readers. This PR branched before that fix and introduces ~17 new bare reads (`if debugLogger != nil { debugLogger.Logf(...)`) in `pool_congestion.go` (lines 132, 248, 400, 407) and `pool_multi_server.go` (lines 166, 382, 397, 409, 556, 584). On merge, the race detector will fire on all these sites under concurrent `SetLogger` calls.

Fix: replace all `debugLogger != nil` guards with `loadDebugLogger() != nil` (or the equivalent atomic accessor present in main after `7d511356`).

https://github.com/opensearch-project/opensearch-go/blob/5e763cb638a3425f6235476194a1ac5b595821f5/opensearchtransport/pool_congestion.go#L131-L134

https://github.com/opensearch-project/opensearch-go/blob/5e763cb638a3425f6235476194a1ac5b595821f5/opensearchtransport/pool_multi_server.go#L165-L168

---

### Bug: `getNextActiveConnWithLock()` return-type change conflicts with main at merge

The PR changes the signature from `*Connection` to `(*Connection, bool)` and updates internal callers accordingly. However, main's `multiServerPool` still declares the method as returning `*Connection`. This is a direct compile-time conflict: the PR cannot be merged without a rebase, because both sides modify the same method signature and its call sites.

This is not a pre-existing issue — it is introduced by the PR's refactor of `getNextActiveConnWithLock` to return a second `triggerCapEnforcement` boolean.

https://github.com/opensearch-project/opensearch-go/blob/5e763cb638a3425f6235476194a1ac5b595821f5/opensearchtransport/pool_multi_server.go#L103-L116

---

### Positive observations

- **`clusterMaxCwnd = max(nodeMax)`, not sum**: Correct. MCSR is a coordinator-side semaphore that can fan all sub-requests to one data node, so the ceiling is the largest single node's pool size. The comment at pool_congestion.go:332 explains this clearly.
- **`computeAdaptiveConcurrency` is a clean pure function**: clamp cwnd to [min, max], return 0 when disabled. No side effects, easy to test in isolation.
- **`appendAdaptiveConcurrency` respects caller-set values**: The linear scan of `RawQuery` correctly skips injection when `max_concurrent_shard_requests=` is already present, so user-supplied overrides always win.
- **`adaptiveConcurrencyMinDefault = 5`** matches the OpenSearch server default — the client never reduces concurrency below what an unmodified request would receive.
- **Test coverage** in `adaptive_concurrency_test.go` (462 lines) is thorough: boundary clamping, feature-flag off, custom limits, `appendAdaptiveConcurrency` idempotency, `WithAdaptiveConcurrencyLimits` validation, and `parseShardRequests` with bare-integer and `min:max` forms are all covered.
- **`clusterSearchAIMD` node churn tests** (new node skips first delta, disappeared node evicted) cover the tricky epoch-tracking edge cases well.

opensearchtransport/pool_congestion_internal_test.go
sean-: Good catch.  The algorithm is deterministic here and these assertions were too loose. Fixed: `require.Equal(t, int32(2), ...)` after poll 2 and `require.Equal(t, int32(4), ...)` after poll 3.

opensearchtransport/pool_congestion.go
sean-: Agreed that this package pools aggressively on the hot path, but I'm going to keep the map here. We have production clusters routinely exceeding 100 nodes, and some of our largest deployments are well beyond that. At 100+ nodes, O(n^2) eviction scans become meaningfully worse than a single map allocation on the background stats goroutine (which runs once per poll interval, typically 10-30s). The map is sized exactly via `make(..., len(polled))` so it's a single allocation with no rehashing.

sean-:
Thanks for the review @Jakob3xD.  Responses below:

**Test assertions (line 632)**: Fixed. Replaced `require.Positive`/`require.GreaterOrEqual` with exact `require.Equal` values for the deterministic AIMD progression (0->1->2->4).

**polledSet allocation (line 319)**: Per above, keeping the map. We have production clusters routinely exceeding 100 nodes, and at that scale O(n^2) eviction is worse than a single pre-sized map allocation on a background goroutine that fires once every 10-30s. The map is allocated exactly to `len(polled)` so there's no rehashing overhead.

**MCSR skip on shard-exact path (nit)**: Added a one-line comment at the early-return explaining the intentional skip (fan-out = 1 means no coordinator fan-out to throttle).

**prevWaitTimeNano reset bug (line 339-364)**: Good find.  Fixed by carrying `epoch.prevWaitTimeNano` forward unconditionally in the struct literal. Previously, a nil `TotalWaitTimeInNanos` poll would zero the field, causing an inflated delta on the next non-nil poll.

**Bare debugLogger reads (data race)**: This was already fixed on `main`.  Resolved by rebasing onto current main (which includes 7d511356). All ~8 remaining bare references in the PR's new code have been migrated to the `loadDebugLogger()` atomic accessor.

**getNextActiveConnWithLock signature conflict**: Also resolved cleanly during the rebase: main didn't have the `bool` return value, so the branch's version (which adds it) applied without conflict.

Jakob3xD (approved):

kkhatua:
@sean- would it be possible to add this to the https://github.com/opensearch-project/opensearch-clients and https://github.com/opensearch-project/opensearch-java projects ? The repo is meant to be a template from which other clients can be derived. Currently, a few.. like Go and Java are the only ones that haven't used this repo.

If you feel a server-side implementation removes the need for this with (i.e. https://github.com/opensearch-project/OpenSearch/issues/21420 ), would you be able to take a stab at this ?

PR #802, opened 2026-03-20
fix: prevent double slashes in AliasGetReq path when Indices is empty

### LINKED ISSUE #617: [BUG] Wrong request path computed in `aliasClient` when no indices are specified
### What is the bug?
OpenSearch supports two kind of routes to get an alias: `GET /_alias/alias1` and `GET /index-1/_alias/alias1`. Although similar, the main difference between them is that while the first returns the alias no matter which index it targets, the latter does it only if it's targeting the specified index. To interact with these routes, the `opensearchapi` package provides the `AliasGetReq` request. The problem arises when no indices are specified: in this case, the `GetRequest()` function computes a request path that is equal to `//_alias/alias1`. When passed to `http.NewRequest()`, the double slash at the beginning makes `_alias` be set as host and `alias1` as path, and when the host is replaced in [`setReqURL()`](https://github.com/opensearch-project/opensearch-go/blob/bd5d233600802d031ed57b0eca34683f184327ce/opensearchtransport/opensearchtransport.go#L420), the ending request URL is `http://localhost:9200/alias1` instead of the expected `http://localhost:9200/_alias/alias1`. While the response for the success case is coincidentally compatible, the behavior is different if more than one alias is specified and some do not exist. In that case, the first endpoint returns a structured error while the second returns the information about the existing aliases along with an error message and a status field. This makes it difficult to handle errors as you expect to receive them in a certain format when in reality this is not the case.

```sh
curl --location 'localhost:9200/test-index/_alias/alias1,alias2'
{"error":"alias [alias2] missing","status":404,"test-index":{"aliases":{"alias1":{}}}}

curl --location 'localhost:9200/alias1,alias2'
{"error":{"root_cause":[{"type":"index_not_found_exception","reason":"no such index [alias2]","index":"alias2","resource.id":"alias2","resource.type":"index_or_alias","index_uuid":"_na_"}],"type":"index_not_found_exception","reason":"no such index [alias2]","index":"alias2","resource.id":"alias2","resource.type":"index_or_alias","index_uuid":"_na_"},"status":404}
```

It it enough to write this code:

```go
if _, err := client.Indices.Alias.Get(context.Background(), opensearchapi.AliasGetReq{Alias: []string{"alias1", "alias2"}}); err != nil {
	var opensearchErr *opensearch.StructError
	if errors.As(err, &opensearchErr) {
		// this works as long as the http://localhost:9200/alias1,alias2 endpoint is called,
		// but the expectation is that localhost:9200/_alias/alias1,alias2 gets called instead,
		// which would return the error in a different format
	}
}
```

The request returned by `AliasGetReq.GetRequest()` should have `/_alias/alias1` set as path when no indices are specified.

It used to work fine before version `3.0`. The commit that introduced the bugged code is 9d56cba2f56a4c8d7b8591e011fe4b76e1197194. I'm open to submit a PR to fix the issue once it's triaged.

anzzyspeaksgit:
## Problem
When calling `AliasGetReq` without specifying `Indices` (or providing an empty `Indices` array/slice), the resulting API path incorrectly contains a double slash.
For example, it might produce `//_alias/test-alias` instead of `/_alias/test-alias`. This double slash causes issues for proxies or when interacting with some clusters, such as Aiven where this causes failures (see #617).

## Solution
I modified the `GetRequest` function in `opensearchapi/api_indices-alias.go` to conditionally append the initial `/` and `indices` string only if `indices` is not empty. If `indices` is empty, it appends only `/_alias/` and the aliases name(s), ensuring the URL path is valid (e.g. `/_alias/test-alias`).

## Testing
- Verified via `make test` that all existing tests pass correctly.

Closes #617

### DIFF
--- a/opensearchapi/api_indices-alias.go
+++ b/opensearchapi/api_indices-alias.go
@@ -119,9 +119,13 @@ func (r AliasGetReq) GetRequest() (*http.Request, error) {
 
 	var path strings.Builder
 	path.Grow(9 + len(indices) + len(aliases))
-	path.WriteString("/")
-	path.WriteString(indices)
-	path.WriteString("/_alias/")
+	if indices != "" {
+		path.WriteString("/")
+		path.WriteString(indices)
+		path.WriteString("/_alias/")
+	} else {
+		path.WriteString("/_alias/")
+	}
 	path.WriteString(aliases)
 	return opensearch.BuildRequest(
 		"GET",


sean-:
There's also a fix posted for this in #804 , too.

sean-:
Thanks for the fix, @anzzyspeaksgit. Just to give you a status update: the underlying bug (#617) was fixed by #804 (merged 2026-05-28), which moved URL-path construction into a generated `internal/path` package. The new `IndicesGetAliasPath` builder uses `hasNonEmpty()` and `writeSegments()` helpers that silently skip empty strings, so `AliasGetReq{Alias: []string{"test-alias"}}` now produces `/_alias/test-alias` correctly.

I'd like to close this PR as resolved by #804. Could you verify the current behavior on main matches what you needed? If you find any case still producing double slashes, please open a new issue with a reproducer and we'll dig in.

Thanks again for the contribution.

PR #803, opened 2026-03-20
Add polymorphic Status field to task response structs

### LINKED ISSUE #788: [FEATURE] Add status to opensearchapi.TasksGetResp

sean-:
Add `Status` (`json.RawMessage`) to `TasksGetResp`, `TasksListTask`, and `TaskCancelInfo` to capture task-type-specific status data without silently dropping fields for non-bulk-by-scroll task types.

Provide `BulkByScrollTaskStatus` convenience type and `ParseBulkByScrollTaskStatus` helper for the common `reindex`, `delete_by_query`, and `update_by_query` case.

Add tasks guide, sample, and integration test assertions.

Resolves #788

### DIFF
--- a/_samples/tasks.go
+++ b/_samples/tasks.go
@@ -1,0 +1,143 @@
+// SPDX-License-Identifier: Apache-2.0
+//
+// The OpenSearch Contributors require contributions made to
+// this file be licensed under the Apache-2.0 license or a
+// compatible open source license.
+package main
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"os"
+	"strings"
+	"time"
+
+	"github.com/opensearch-project/opensearch-go/v4"
+	"github.com/opensearch-project/opensearch-go/v4/opensearchapi"
+)
+
+func main() {
+	if err := example(); err != nil {
+		fmt.Println(fmt.Sprintf("Error: %s", err))
+		os.Exit(1)
+	}
+}
+
+func example() error {
+	// Initialize the client with SSL/TLS enabled.
+	client, err := opensearchapi.NewClient(
+		opensearchapi.Config{
+			Client: opensearch.Config{
+				InsecureSkipVerify: true, // For testing only. Use certificate for validation.
+				Addresses:          []string{"https://localhost:9200"},
+				Username:           "admin", // For testing only. Don't store credentials in code.
+				Password:           "myStrongPassword123!",
+			},
+		},
+	)
+	if err != nil {
+		return err
+	}
+
+	ctx := context.Background()
+
+	sourceIndex := "task-source"
+	destIndex := "task-dest"
+
+	// Create source index with test data.
+	_, err = client.Indices.Create(ctx, opensearchapi.IndicesCreateReq{
+		Index: sourceIndex,
+		Body:  strings.NewReader(`{"settings": {"number_of_shards": 1, "number_of_replicas": 0}}`),
+	})
+	if err != nil {
+		return err
+	}
+
+	_, err = client.Index(ctx, opensearchapi.IndexReq{
+		Index: sourceIndex,
+		Body:  strings.NewReader(`{"title": "Test Document", "year": 2024}`),
+		Params: opensearchapi.IndexParams{
+			Refresh: "true",
+		},
+	})
+	if err != nil {
+		return err
+	}
+
+	// Submit an async reindex task.
+	reindexResp, err := client.Reindex(ctx, opensearchapi.ReindexReq{
+		Body: strings.NewReader(fmt.Sprintf(
+			`{"source":{"index":"%s"},"dest":{"index":"%s"}}`,
+			sourceIndex, destIndex,
+		)),
+		Params: opensearchapi.ReindexParams{
+			WaitForCompletion: opensearchapi.ToPointer(false),
+		},
+	})
+	if err != nil {
+		return err
+	}
+	taskID := reindexResp.Task
+	fmt.Printf("Task submitted: %s\n", taskID)
+
+	// Poll for completion.
+	var taskResp *opensearchapi.TasksGetResp
+	for {
+		taskResp, err = client.Tasks.Get(ctx, opensearchapi.TasksGetReq{TaskID: taskID})
+		if err != nil {
+			return err
+		}
+		if taskResp.Completed {
+			break
+		}
+		time.Sleep(500 * time.Millisecond)
+	}
+	fmt.Printf("Task completed: action=%s\n", taskResp.Task.Action)
+
+	// Parse the BulkByScroll status.
+	status, err := opensearchapi.ParseBulkByScrollTaskStatus(taskResp.Task.Status)
+	if err != nil {
+		return err
+	}
+
+	fmt.Printf("Total: %d\n", status.Total)
+	fmt.Printf("Created: %d\n", status.Created)
+	fmt.Printf("Updated: %d\n", status.Updated)
+	fmt.Printf("Deleted: %d\n", status.Deleted)
+	fmt.Printf("Batches: %d\n", status.Batches)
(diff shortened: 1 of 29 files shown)


VijayanB (dismissed):

bobbyphilip (approved):

Jakob3xD:
Why not defining a struct as it is defined by the openapi spec of opensearch?
You can get the spec from this release:
https://github.com/opensearch-project/opensearch-api-specification/releases/tag/main-latest

sean-:
This test failure has been fixed in #821

_samples/tasks.go
```diff
@@ -0,0 +1,143 @@
...
+	"strings"
+	"time"
+
+	"github.com/opensearch-project/opensearch-go/v4"
+	"github.com/opensearch-project/opensearch-go/v4/opensearchapi"
+)
+
+func main() {
+	if err := example(); err != nil {
+		fmt.Println(fmt.Sprintf("Error: %s", err))
```
tender-barbarian: Unnecessary nested `fmt.Println(fmt.Sprintf(...))`. 

```suggestion
        fmt.Printf("Error: %s\n", err)
```

opensearchapi/api_tasks_types.go
```diff
@@ -0,0 +1,264 @@
...
+		return nil, fmt.Errorf("status is nil")
+	}
+
+	var s PersistentTaskStatus
+	if err := json.Unmarshal(status, &s); err != nil {
+		return nil, fmt.Errorf("failed to unmarshal PersistentTaskStatus: %w", err)
+	}
+
+	return &s, nil
+}
```
tender-barbarian: These 4 are near identical with only difference being returned type. How about we use generic instead? We can keep the named wrappers, but have them delegate to the generic.

```suggestion
// ParseBulkByScrollTaskStatus unmarshals a task's raw Status field into a
// BulkByScrollTaskStatus. Use this for reindex, delete_by_query, and
// update_by_query tasks.
//
// Returns an error if status is nil or cannot be unmarshaled.
func ParseBulkByScrollTaskStatus(status json.RawMessage) (*BulkByScrollTaskStatus, error) {
	return parseTaskStatus[BulkByScrollTaskStatus](status)
}

// ParseReplicationTaskStatus unmarshals a task's raw Status field into a
// ReplicationTaskStatus.
//
// Returns an error if status is nil or cannot be unmarshaled.
func ParseReplicationTaskStatus(status json.RawMessage) (*ReplicationTaskStatus, error) {
	return parseTaskStatus[ReplicationTaskStatus](status)
}

// ParseResyncTaskStatus unmarshals a task's raw Status field into a
// ResyncTaskStatus. Use this for primary-replica resync tasks.
//
// Returns an error if status is nil or cannot be unmarshaled.
func ParseResyncTaskStatus(status json.RawMessage) (*ResyncTaskStatus, error) {
	return parseTaskStatus[ResyncTaskStatus](status)
}

// ParsePersistentTaskStatus unmarshals a task's raw Status field into a
// PersistentTaskStatus.
//
// Returns an error if status is nil or cannot be unmarshaled.
func ParsePersistentTaskStatus(status json.RawMessage) (*PersistentTaskStatus, error) {
	return parseTaskStatus[PersistentTaskStatus](status)
}

func parseTaskStatus[T any](raw json.RawMessage) (*T, error) {
	if raw == nil {
		return nil, ErrNilTaskStatus
	}
	var s T
	if err := json.Unmarshal(raw, &s); err != nil {
		return nil, fmt.Errorf("unmarshaling %T: %w", s, err)
	}
	return &s, nil
}
```

Notice I also used sentinel error 

```Go
var ErrNilTaskStatus = errors.New("task status is nil")

File: opensearchapi/api_tasks_types.go [hunk 3]
```diff
@@ -0,0 +1,264 @@
...
+	var status BulkByScrollTaskStatus
+	if err := json.Unmarshal(data, &status); err != nil {
+		return fmt.Errorf("failed to unmarshal BulkByScrollTaskStatus: %w", err)
+	}
+	s.Status = &status
+	return nil
+}
+
+// MarshalJSON implements custom marshaling for BulkByScrollTaskStatusOrException.
+func (s BulkByScrollTaskStatusOrException) MarshalJSON() ([]byte, error) {
```
tender-barbarian, line 156: I suggest to add test for this as well.

_samples/tasks.go
```diff
@@ -0,0 +1,143 @@
...
+	var taskResp *opensearchapi.TasksGetResp
+	for {
+		taskResp, err = client.Tasks.Get(ctx, opensearchapi.TasksGetReq{TaskID: taskID})
+		if err != nil {
+			return err
+		}
+		if taskResp.Completed {
+			break
+		}
+		time.Sleep(500 * time.Millisecond)
```
tender-barbarian: More idiomatic Go pattern for polling would be: 

```Go
ticker := time.NewTicker(500 * time.Millisecond)
defer ticker.Stop()

var taskResp *opensearchapi.TasksGetResp
for {
    // get task
    select {
    case <-ctx.Done():
        return nil, ctx.Err()
    case <-ticker.C:
    }
}

File: _samples/tasks.go [hunk 5]
```diff
@@ -0,0 +1,143 @@
...
+				Username:           "admin", // For testing only. Don't store credentials in code.
+				Password:           "myStrongPassword123!",
+			},
+		},
+	)
+	if err != nil {
+		return err
+	}
+
+	ctx := context.Background()
```
tender-barbarian: Graceful shutdown with timeout

```suggestion
ctx, stop := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGTERM)
defer stop()
ctx, cancel := context.WithTimeout(ctx, 5*time.Minute)
defer cancel()

**Review by tender-barbarian — CHANGES_REQUESTED** [review]
Looks good overall, just some minor nitpicking. 

Changes suggested to the sample code are not strictly necessary since it's just a trivial example, but it's users first impression of the API so should be tight IMHO.

**Comment by tender-barbarian** [comment]
@sean- The Windows test failure is caused by `time.Now().UTC()` in `OnFlushStart`. Calling `.UTC()` strips the monotonic clock, so `time.Since` falls back to wall-clock thus loosing precision. On Windows the wall clock resolution is ~15.6 ms which is too long for this almost instant test. Flush completes within a single tick, so both time reads return the same instant and the duration comes out as 0s. Linux/MacOs have better precision so it doesn't happen there.

Fix: drop the `.UTC()` call, the test only compares start vs end, so monotonic reading is what you want.

```go
return context.WithValue(ctx, contextKey("start"), time.Now())
```

sean-:
#819 skips a known CI flake, and #823 will let us get the detail we need to figure out why this is flaking on >=2.2.0

tender-barbarian (approved):

Jakob3xD:
You can remove 007c53d7a44c7c8ced74e64545b025675fef2c4d from this PR?

Jakob3xD (approved):

PR #809, opened 2026-03-24
Add generic Do[T] function for compile-time pointer enforcement

### LINKED ISSUE #808: [PROPOSAL] Client.Do() accepts any for dataPointer: add generic Do[T] for compile-time pointer safety

`Client.Do()` has this signature:

```go
func (c *Client) Do(ctx context.Context, req Request, dataPointer any) (*Response, error)
```

The `dataPointer` parameter must be a pointer for `json.Unmarshal` to work, but the `any` type cannot enforce this at compile time. Passing a non-pointer value compiles without error and only fails at runtime:

```go
var resp MyResp
// Compiles fine, fails at runtime: json: Unmarshal(non-pointer main.MyResp)
client.Do(ctx, req, resp)
```

This is the same problem `json.Unmarshal` has, but unlike `json.Unmarshal` (which is constrained by Go 1 compatibility), the `opensearch-go` client can use generics (available since Go 1.18, and the module already requires 1.21+).

The bug is subtle because the code looks correct at a glance — the caller simply forgot the `&`. In practice we've seen this cause production incidents where response data silently comes back as zero-valued structs.

### Describe the solution you'd like

Add a generic top-level function alongside the existing method:

```go
// Do is a generic version of Client.Do that enforces dataPointer as a pointer at compile time.
func Do[T any](ctx context.Context, c *Client, req Request, dataPointer *T) (*Response, error) {
    return c.Do(ctx, req, dataPointer)
}
```

The existing `Client.Do()` method signature is **unchanged** — no breaking changes, and the method will not be removed. The generic function is the recommended path forward, and the compiler rejects non-pointer arguments:

```go
var resp MyResp
opensearch.Do(ctx, client, req, &resp)    // OK — *T enforced
opensearch.Do(ctx, client, req, resp)     // compile error
```

`Client.Do()` is marked with a Go `Deprecated` doc annotation to steer callers toward the safer alternative. It remains fully functional and will not be removed; the annotation exists so that `staticcheck` SA1019 and IDE tooling (gopls shows strikethroughs) nudge cross-package callers toward `opensearch.Do[T]()`.

Internally, the unexported `do` method in `opensearchapi`, `plugins/security`, and `plugins/ism` is converted to package-level generic functions, enforcing the same safety across all ~200 internal call sites. Requests that expect no response body (Ping, Delete, etc.) use a separate `doRequest` function with no data parameter, which is cleaner than the current `nil` overloading.

### Additional context

- Go does not allow type parameters on methods, so this must be a top-level function. The ergonomic trade-off (`opensearch.Do(ctx, c, req, &resp)` vs `c.Do(ctx, req, &resp)`) is minor.
- The generic wrapper is trivially inlined by the compiler (single `return` statement, well under the inlining cost budget of 80). Zero runtime overhead.
- All pointer types share the same GC shape, so only one instantiation is generated regardless of how many response types are used.
- The internal refactor splits the overloaded `do(ctx, req, nil)` / `do(ctx, req, &data)` pattern into two functions (`doRequest` / `do[T]`), which makes the error handling paths clearer — typed responses always parse structured errors, no-body requests return status-only errors.
- The `Deprecated` annotation is a nudge, not a removal signal. `Client.Do()` remains functional for callers who need it (e.g., dynamic use cases where the response type isn't known at compile time).

sean-:
Add generic Do[T] function for compile-time pointer enforcement

Client.Do() accepts `any` for its dataPointer parameter, deferring
pointer validation to json.Unmarshal at runtime. Add a top-level
opensearch.Do[T]() function whose *T signature catches non-pointer
arguments at compile time. The generic wrapper is trivially inlined
by the compiler — zero runtime overhead.

Mark Client.Do() with a Deprecated doc annotation to nudge callers
toward the safer alternative via staticcheck SA1019 and IDE tooling.
The method remains fully functional and will not be removed.

Convert the unexported `do` method in opensearchapi, plugins/security,
and plugins/ism to package-level generic functions, enforcing pointer
safety across all ~200 internal call sites. Requests that expect no
response body use a separate doRequest() function.

Fixes: #808

Signed-off-by: Sean Chittenden <sean.chittenden@crowdstrike.com>


opensearchtransport/coverage_async_internal_test.go
Jakob3xD, line 1: Move this to a separate commit.

opensearchtransport/discovery_internal_test.go
Jakob3xD, line 1: Move this to a separate commit.

VijayanB (dismissed):

opensearch.go
```diff
@@ -420,6 +425,12 @@ func (c *Client) Do(ctx context.Context, req Request, dataPointer any) (*Respons
 	return response, nil
 }
 
+// Do is a generic version of [Client.Do] that enforces dataPointer as a pointer at compile time.
+// It delegates to [Client.Do] after the type system has guaranteed *T.
+func Do[T any](ctx context.Context, c *Client, req Request, dataPointer *T) (*Response, error) {
+	return c.Do(ctx, req, dataPointer)
+}
```
tender-barbarian, line 444: Consider what happens if someone tries to use typed nil:

```Go
var resp *StatusResp
_, err := opensearch.Do(ctx, c, req, resp)
```

This fails with `InvalidUnmarshalError`. Once `resp` is widened to `any`, the envelope becomes `(type: *StatusResp, value: <nil>)` — the non-nil type means `dataPointer != nil` evaluates `true` at [opensearch.go:412](https://github.com/opensearch-project/opensearch-go/blob/b7f14b5/opensearch.go#L412), so we enter the branch and hand a nil pointer to `json.Unmarshal`. Arguably, its users fault for not initialising the type... But if we're trying to solve JSON runtime unmarshaling failures, I suggest we also handle typed nil.

My suggestions:

1. Document it. We should mention this case explicitly.
2. Add a guard:

```Go
func Do[T any](ctx context.Context, c *Client, req Request, dataPointer *T) (*Response, error) {
    if dataPointer == nil {
        return c.Do(ctx, req, nil)
    }
    return c.Do(ctx, req, dataPointer)
}
```

It works because `dataPointer` is still a concrete `*T`, not yet widened to `any`, so we catch the typed nil early and just forward it as untyped, it will eventually hit [opensearch.go:412](https://github.com/opensearch-project/opensearch-go/blob/b7f14b5/opensearch.go#L412) and skip unmarshal.

tender-barbarian (changes requested):

opensearch_internal_test.go
```diff
@@ -218,6 +218,35 @@ func TestClientInterfe(t *testing.T) {
...
+	})
+
+	t.Run("Generic Do() nil pointer", func(t *testing.T) {
+		c, err := NewClient(Config{Transport: mockhttp.NewRoundTripFunc(t, defaultRoundTripFunc)})
+		require.NoError(t, err)
+
+		resp, err := Do[struct{}](context.TODO(), c, testReq{}, nil)
+		require.NoError(t, err)
+		require.NotNil(t, resp)
+	})
```
tender-barbarian: I'm not loving this test. Generic `Do` enforces pointer types but not non-nilness and test passes because `resp.Body` is `nil` not because `Do[T]` handles nil pointers. 

`testReq{}` has an empty `Path`, and `defaultRoundTripFunc` at [opensearch_internal_test.go:52](https://github.com/opensearch-project/opensearch-go/blob/b7f14b5/opensearch_internal_test.go#L52) only sets `response.Body` for `"/"` and `"/test"`. So unmarshal is entirely skipped since there is no body at all: [opensearch.go:412](https://github.com/opensearch-project/opensearch-go/blob/b7f14b5/opensearch.go#L412). Realistically, response body is never going to be `nil` in production. So what is this testing really?

tender-barbarian (changes requested):

opensearch.go
sean-: Great catch. You're right that the typed nil slips through the `any` widening and reaches `json.Unmarshal` as a non-nil interface with a nil value.

Added the guard you suggested:

```go
func Do[T any](ctx context.Context, c *Client, req Request, dataPointer *T) (*Response, error) {
    if dataPointer == nil {
        return c.Do(ctx, req, nil)
    }
    return c.Do(ctx, req, dataPointer)
}
```

The doc comment now also calls out the typed nil behavior explicitly.

The three internal `do[T]` functions in `opensearchapi`, `plugins/security`, and `plugins/ism` all delegate to `opensearch.Do`, so the guard covers them transitively.

opensearch_internal_test.go
sean-: Fair point. The empty path meant `defaultRoundTripFunc` returned a nil body, and the test passed because the unmarshal branch was never entered, not because the nil guard worked.

Fixed the test to use `testReq{Path: "/"}` so the response carries a JSON body. Now the test actually proves the typed nil is intercepted before it can reach `json.Unmarshal`.

tender-barbarian (approved):

Jakob3xD:
Could it be an alternative to introduce a 'NoBody' type of type nil, that we use for endpoints that return nothing? So we don't need the additional `doRequest()` function? 

```
  // NoBody is a marker type for do() calls that expect no response body.                                                                                                                      
  type NoBody struct{}                                                                                                                                                                         
                                                                                                                                                                                               
  // noBody is a pre-declared typed nil — passed to do() when unmarshaling                                                                                                                     
  // should be skipped. Type inference resolves T=NoBody; the nil pointer                                                                                                                      
  // triggers the dataPointer == nil guard in opensearch.Do, skipping json.Unmarshal.                                                                                                          
  var noBody *NoBody 
  ```

**Comment by sean-** [comment]
> Could it be an alternative to introduce a 'NoBody' type of type nil, that we use for endpoints that return nothing? So we don't need the additional `doRequest()` function?

Oh, that's a good idea.  Yeah, that removes the `doRequest`/`do[T]` split and yields a single dispatch path through the generic function. Cleaner and harder to misuse.  Can you PTAL?

Thanks for the nudge on this, @Jakob3xD.

File: plugins/security/api.go [hunk 3]
```diff
@@ -61,19 +60,16 @@ func NewClient(config Config) (*Client, error) {
...
+// do calls [opensearch.Do] and checks the response for errors.
+// The generic *T parameter enforces that dataPointer is a pointer at compile time.
+func do[T any](ctx context.Context, c *Client, req opensearch.Request, dataPointer *T) (*opensearch.Response, error) {
+	resp, err := opensearch.Do(ctx, c.Client, req, dataPointer)
 	if err != nil {
 		return nil, err
 	}
 
 	if resp.IsError() {
-		if dataPointer != nil {
```
Jakob3xD, line 72: With removing this check, Endpoints that return only a status code and no body will run into following line. https://github.com/opensearch-project/opensearch-go/blob/main/error.go#L145

I would say, it is expected that some endpoints return no body. Therefore, we might want to adjust the check.

```go
if dataPointer == nil && resp.Body == nil {
    return resp, fmt.Errorf("status: %s", resp.Status())
} else {
    return resp, opensearch.ParseError(resp)
}
```

This also applies for the other plugins and the opensearchapi.

error_test.go
Jakob3xD, line 1: Move this to a separate commit.

plugins/ism/api.go
```diff
@@ -43,19 +42,16 @@ func NewClient(config Config) (*Client, error) {
...
 		return nil, err
 	}
 
 	if resp.IsError() {
-		if dataPointer != nil {
-			return resp, opensearch.ParseError(resp)
-		} else {
-			return resp, fmt.Errorf("status: %s", resp.Status())
-		}
+		return resp, opensearch.ParseError(resp)
```
Jakob3xD, line 58: Same like below.

opensearchapi/opensearchapi.go
```diff
@@ -104,24 +103,26 @@ func NewFromClient(client *opensearch.Client) *Client {
...
 		return nil, err
 	}
 
 	if resp.IsError() {
-		if dataPointer != nil {
-			return resp, opensearch.ParseError(resp)
-		} else {
-			return resp, fmt.Errorf("status: %s", resp.Status())
-		}
+		return resp, opensearch.ParseError(resp)
```
Jakob3xD, line 119: Same as below

Jakob3xD (changes requested):
I would suggest waiting for a go release that allow generics

error_test.go
sean-: Done - split into its own commit in the latest force-push.

plugins/security/api.go
sean-: Good catch. Applied your suggestion across all three packages (`opensearchapi`, `plugins/security`, `plugins/ism`). When `dataPointer == nil && resp.Body == nil`, we now return a status-only error instead of calling `ParseError`. Used `%q` for the status string since it originates from the HTTP response, too.

sean-:
> I would suggest waiting for a go release that allow generics

`go.mod` already requires Go 1.25.9, which has full generics support including generic type aliases. Is there a specific language feature you're thinking of that isn't available yet?

Jakob3xD (approved):

Jakob3xD:
> `go.mod` already requires Go 1.25.9, which has full generics support including generic type aliases. Is there a specific language feature you're thinking of that isn't available yet?

I meant the “Generic Methods for Go” that is planned for GO 1.27. With that we could have just adjusted the Do Functions with a generic one.

However, as we don't try to run this lib on the latest go version, we can merge this one and maybe have a look later on.

PR #782, opened 2026-01-20
bump OpenSearch compatiblity tests to latest releases

sean-:
Update library to support OpenSearch >= 3.1.0

This update adds support for all released versions of OpenSearch, improves test client robustness, and implements necessary API compatibility fixes across OpenSearch versions.

Test Infrastructure Improvements:
- Update `NewClient(t)` to perform cluster readiness validation (health + cluster state + nodes info)
- Add version-aware client configuration supporting both secure/insecure modes with automatic password selection
- Create testify-based `OpenSearchTestSuite` with built-in version detection and health checking
- Implement cluster readiness polling (25 attempts × 5s intervals) to eliminate test flakiness

OpenSearch API Compatibility Enhancements:
- 3.1.0+: Add `TimeInExecutionMillis`, `TimeInExecution` (cluster pending tasks), `PhaseResultsProcessors` (nodes info)
- 3.2.0+: Add `MaxLastIndexRequestTimestamp` (indexing), StartTree query metrics (`StartreeQueryTotal`, `StartreeQueryTimeInMillis`, `StartreeQueryCurrent`), precision fix for `ConcurrentAVGSliceCount`
  (float32 -> float64), `Settings` map (security health)
- 3.3.0+: Add `QueryFailed`, `StartreeQueryFailed` (search stats), `IngestionStatus` (cluster state), `JwksURI` (security config)
- 3.4.0+: Add comprehensive `Warmer` struct with 8 sub-fields (merges section), `StatusCounter` for document status tracking, `Parallelism` (thread pool)

Critical Fix: Cat Indices Field Naming Compatibility:
OpenSearch 3.2.0 introduced inconsistent StarTree field naming (`pri.startree.*`) which was corrected in 3.3.0+ (`pri.search.startree.*`). Implemented dual-field approach with automatic consolidation:
- Stable fields use corrected 3.3.0+ naming as primary
- V32 compatibility fields handle temporary 3.2.0 naming
- `consolidateFields()` method provides transparent fallback logic
- Comprehensive test coverage for all naming scenarios
- All compatibility code marked for removal when 3.2.0 support ends

CI/Build System Enhancements:
- Expand compatibility test matrix from 8 to 30 OpenSearch versions (1.3.20, 2.0.1-2.19.4, 3.0.0-3.4.0, latest)
- Add plugin dependency handling for OpenSearch >= 3.1.0 (flow-framework -> skills -> ml -> anomaly-detection -> security)

Code Quality Improvements:
[...]

PR #810, opened 2026-03-25
Bump OpenSearch version to recent releases

sean-:
Bump versions tested in CI

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -111,6 +111,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ### Changed
 
+- Bump CI and developer guide OpenSearch versions: compatibility matrix to 2.19.5, default integration test version to 3.6.0 ([#810](https://github.com/opensearch-project/opensearch-go/pull/810))
 - Include `_nodes.failures` detail in discovery error messages for diagnosing intermittent CI failures on older OpenSearch versions ([#823](https://github.com/opensearch-project/opensearch-go/pull/823))
 - Test against Opensearch 3.6.0 ([#817](https://github.com/opensearch-project/opensearch-go/pull/817))
 - Consolidate test utilities into two canonical packages: opensearchtransport/testutil (env helpers, polling, version comparison) and opensearchapi/testutil (client-dependent helpers, test suite, JSON comparison)
--- a/opensearchtransport/opensearchtransport.go
+++ b/opensearchtransport/opensearchtransport.go
@@ -932,7 +932,7 @@ func New(cfg Config) (*Client, error) {
 	}
 
 	if cfg.EnableDebugLogger {
-		debugLogger = &debuggingLogger{Output: os.Stdout}
+		storeDebugLogger(&debuggingLogger{Output: os.Stdout})
 	}
 
 	if cfg.EnableMetrics {
@@ -1216,8 +1216,8 @@ func (c *Client) Perform(req *http.Request) (*http.Response, error) {
 				c.metrics.failures.Add(1)
 			}
 
-			if debugLogger != nil {
-				debugLogger.Logf("Request to %s failed: %v\n", conn.URL, err)
+			if dl := loadDebugLogger(); dl != nil {
+				dl.Logf("Request to %s failed: %v\n", conn.URL, err)
 			}
 
 			// Retry on HTTP/2 stream resets (RST_STREAM frames such as REFUSED_STREAM).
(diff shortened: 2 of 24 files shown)


CHANGELOG.md
```diff
@@ -114,7 +114,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - Rename `singleConnectionPool` to `singleServerPool` and `statusConnectionPool` to `multiServerPool` for clarity ([#786](https://github.com/opensearch-project/opensearch-go/pull/786))
 - Refactor Client struct to use embedded mutex pattern for improved thread safety ([#775](https://github.com/opensearch-project/opensearch-go/pull/775))
 - Refactor metrics struct to use atomic counters for lock-free request/failure tracking ([#776](https://github.com/opensearch-project/opensearch-go/pull/776))
-- Test against Opensearch 2.19.4, 3.1, 3.3, and 3.4 ([#782](https://github.com/opensearch-project/opensearch-go/pull/782))
+- Bump CI and developer guide OpenSearch versions: compatibility matrix to 2.19.5, default integration test version to 3.5.0 ([#782](https://github.com/opensearch-project/opensearch-go/pull/782))
```
Jakob3xD: Please don't replace existing entries.

Jakob3xD (changes requested):

CHANGELOG.md
```diff
@@ -110,6 +110,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ### Changed
 
+- Bump CI and developer guide OpenSearch versions: compatibility matrix to 2.19.5, default integration test version to 3.5.0 ([#782](https://github.com/opensearch-project/opensearch-go/pull/782))
```
```suggestion
- Bump CI and developer guide OpenSearch versions: compatibility matrix to 2.19.5, default integration test version to 3.5.0 ([#810](https://github.com/opensearch-project/opensearch-go/pull/810))
```

VijayanB (dismissed):

sean-:
And this CI issue is fixed in #811

rios0rios0 (commented):
Reviewed against issue #813 "quick wins" list. Content LGTM, one unaddressed nit from @Jakob3xD remains.

**What the PR does:** Bumps `OPENSEARCH_VERSION` in the integration workflow from 2.19.4 to 3.6.0, bumps the 2.19.x entry in the compatibility matrix from 2.19.4 to 2.19.5, and updates two `DEVELOPER_GUIDE.md` examples to match.

**Verification:**
- Built `pull/810/head` clean, `go test -race -short` clean.
- CI is green across the full matrix (all 28 integ-test-compat jobs pass).

**One nit** — in the new CHANGELOG line:

```
- Bump CI and developer guide OpenSearch versions: compatibility matrix to 2.19.5, default integration test version to 3.6.0 ([#782](...))
```

the PR reference is `#782`, but this is PR #810. @Jakob3xD called this out in a suggestion block on 2026-03-27 — looks like the number just wasn't updated. One-line fix and this is ready.

Posting as a comment rather than an approval since Jakob's `CHANGES_REQUESTED` still points to a real unresolved issue; once the PR number is fixed I'm happy to convert to an approval.

sean-:
Good catch.  The PR has been updated to reference #810. Kept 3.6.0 since it's been released.

opensearchtransport/opensearchtransport.go
```diff
@@ -1216,8 +1216,8 @@ func (c *Client) Perform(req *http.Request) (*http.Response, error) {
 				c.metrics.failures.Add(1)
 			}
 
-			if debugLogger != nil {
-				debugLogger.Logf("Request to %s failed: %v\n", conn.URL, err)
+			if dl := loadDebugLogger(); dl != nil {
```
linzeyan, line 1219: One small suggestion: we might consider using a no-op logger instead of returning `nil` from `loadDebugLogger()`. This would allow callers to use the logger without having to check for `nil` every time, which can simplify the call sites a bit.

For example:

```go

var noop = &noopLogger{}

type noopLogger struct{}

func (*noopLogger) Log(a ...any) error {
	return nil
}

func (*noopLogger) Logf(format string, a ...any) error {
	return nil
}

func loadDebugLogger() DebuggingLogger {
	if p := debugLoggerPtr.Load(); p != nil {
		return *p
	}
	return noop
}
```

This way, callers can just do:

```go
loadDebugLogger().Logf(...)
```

without additional nil checks.

sean-: I like zerolog's approach as a library. In this case, the goal was narrowly scoped: fix a race condition with the old logging idiom, not replace or introduce a new logger. The loadDebugLogger() pattern with an atomic load plus a nil check is effectively as cheap as it gets (one atomic load, one branch, zero allocation, no argument evaluation).

A broader logging rethink is worth revisiting in the future, though.

Jakob3xD (approved):

PR #812, opened 2026-03-26
Update metrics and error handling docs

sean-:
Update user-facing documentation across two areas:

1. **New: client-side metrics guide** (`guides/metrics.md`). Documents the `Metrics API`, `ConnectionMetric`, `PolicySnapshot`, and `RouterSnapshot` surfaces.

2. **Update: partial-failure call-site recommendation.** User-facing docs (`v5preview/opensearchapi/README.md`, `guides/error_handling.md`) no longer present per-Resp helpers (`BulkItemFailures`, `SearchShardFailures`, `WriteShardFailures`, `MultiSearchItemFailures`, `PartialFailures(mask)`) as the call-site pattern. Helpers remain available without deprecation markers; doc framing changes to:

   - **Hard error:** `if err != nil { return err }` for operations where any failure is reason to stop.
   - **Category-aware:** `for`/`switch` over `opensearchapi.Errors(err)` when partial-error handling lets the application recover from known tolerated failure modes.

   The antipattern discussion covers `errors.As`, `Has`-style helpers, and per-Resp helpers together: all three answer the narrow question "did this category happen?" and silently miss categories added in a future release. The type switch is the only category-aware pattern recommended.

   `opensearchapi/partial_failure_methods.go` file header, `DEVELOPER_GUIDE.md`, and `cmd/osgen/README.md` describe the helpers as engine machinery the dispatch consumes and redirect call-site authors at `opensearchapi.Errors(err)` + `for`/`switch`.


CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ### Added
 
+ Add client-side metrics guide covering Metrics API, ConnectionMetric, PolicySnapshot, and RouterSnapshot
```
```suggestion
- Add client-side metrics guide covering Metrics API, ConnectionMetric, PolicySnapshot, and RouterSnapshot
```

Jakob3xD (changes requested):

sean-:
This has been fixed in other PRs that have been filed and should only be applied after #801 is committed.

sean-:
This CI failure is fixed in #819

rios0rios0 (changes requested):
Reviewed against issue #813 "quick wins" list. The doc content itself is excellent — well-organized, tables for every struct, a working polling example, Prometheus/OTel mapping hints. However, **it's premature to merge** and I think the "quick wins" list may be misleading here.

**Blocker:** Several APIs documented in `guides/metrics.md` do not yet exist on `main`:

- `PolicySnapshot` struct (documented with 13 fields including `WarmupSkips`, `WarmupAccepts`, etc.) — `grep -rn PolicySnapshot` on `main` returns zero hits.
- `ConnectionMetric.MCSR` field (`Adaptive max_concurrent_shard_requests value`) — not present on `main`'s `ConnectionMetric` definition in `opensearchtransport/metrics.go`.

Both appear to be added by #801 ("Add adaptive `max_concurrent_shard_requests` for search routing"), which is still open. As @sean- noted above, this PR "should only be applied after #801 is committed."

**Also:** the failing check `integ-test-compat (true, 2.1.0)` is the exact security-plugin `OptionalDataException` race that #819 fixes. Once #819 lands and this is rebased, that check should go green.

**Recommendation:** hold until #801 merges, then rebase onto `main` (which will include #819's skip guards). At that point the doc will match the shipped API and CI will be clean.

Jakob3xD's earlier `CHANGES_REQUESTED` suggestion on the CHANGELOG line is **already addressed** by the current diff (same wording as his suggestion block).

sean-:
Thanks for the thorough review, @rios0rios0.  This PR's is intended to be intentionally sequenced **after** the following:

| PR | Dependency | Status |
|---|---|---|
| #810 | Bump OpenSearch version to recent releases (bug fixes) | Open |
| #801 | Introduces `PolicySnapshot`, `ConnectionMetric.MCSR`, and the adaptive `max_concurrent_shard_requests` routing that this doc describes | Open |
| #819 | Fixes the `integ-test-compat (true, 2.1.0)` `OptionalDataException` race causing the CI failure here | Open |

The plan is:
1. **#810** and **#819** merge first (bug fixes, CI fix)
2. **#801** merges (adds the APIs this doc references)
3. **This PR** rebases onto `main` at which point the documented structs/fields will exist and CI should be green

I filed this early so the documentation could be reviewed in parallel with the implementation PRs, but agree it should not merge until its dependencies land.

@Jakob3xD: the CHANGELOG entry already reflects your suggested wording. Let me know if there's anything else you'd like adjusted.

guides/metrics.md
```diff
@@ -0,0 +1,254 @@
...
+| `ConnectionMetric.EstLoad`   | Gauge (per-node)   | `opensearch_client_node_est_load{node="..."}`   |
+| `ConnectionMetric.MCSR`      | Gauge (per-node)   | `opensearch_client_node_mcsr{node="..."}`       |
+| `PolicySnapshot.ActiveCount` | Gauge (per-policy) | `opensearch_client_policy_active{policy="..."}` |
+| `IndexRouterState.FanOut`    | Gauge (per-index)  | `opensearch_client_index_fan_out{index="..."}`  |
+
+---
+
+## Observer API
+
+For event-driven observability (as opposed to polling), implement the `ConnectionObserver` interface and pass it via `opensearchtransport.WithObserver()`. The observer receives callbacks for connection lifecycle events (promote, demote, overload), routing decisions, health checks, and shard map invalidations. See the [routing guide](routing.md) for details on observer events.
```
Jakob3xD: `opensearchtransport.WithObserver()` does not exist in this module — there is no exported `WithObserver` function anywhere in the repo (the only match is the test name `TestDocRouterEval_ShardExactWithObserver`). The observer is wired via the `Observer` field on `opensearch.Config` (`Observer opensearchtransport.ConnectionObserver`). Code following this guide as written won't compile.

Suggest: "...implement the `ConnectionObserver` interface and set it on the `Observer` field of `opensearch.Config`."

guides/metrics.md
```diff
@@ -0,0 +1,254 @@
...
+    // Router cache (when routing is active)
+    if m.Router != nil {
+        for _, idx := range m.Router.Indexes {
+            log.Printf("  index %s: fan_out=%d shard_nodes=%d rate=%.1f",
+                idx.Name, idx.FanOut, idx.ShardNodes, idx.RequestRate)
+        }
+    }
+}
+
+func safeFloat(f *float64) float64 {
```
Jakob3xD: This `func safeFloat` is declared *after* statements inside the same code block (the `ticker := ...` / `for range ticker.C { ... }` loop above). It won't compile as shown: a named function declaration can't appear inside a function body, and the preceding statements can't appear at package scope. Hoist `safeFloat` into its own top-level snippet (or define it before the loop).

guides/metrics.md
```diff
@@ -0,0 +1,254 @@
...
+
+The `Metrics()` method lives on `opensearch.Client`. It returns an `opensearchtransport.Metrics` struct and an error -- non-nil when metrics are disabled or a snapshot callback fails.
+
+---
+
+## Metrics Struct
+
+The top-level `Metrics` struct contains aggregate counters, per-connection details, per-policy pool snapshots, and router cache state.
+
+### Request Counters
```
Jakob3xD, line 35: Convention nit: these four aggregate tables (Request Counters, Connection Pool State, Connection Lifecycle Counters, Health Check Counters) list JSON tag names under a column titled `Field`, but the later tables (ConnectionMetric, PolicySnapshot, RouterSnapshot, ...) put the Go field name under `Field` plus a separate `JSON` column — and the polling example and Prometheus table refer to these by their Go names (`m.Requests`, `m.LiveConnections`). Within one doc "Field" then means two different things. Consider giving these four the same `Field` (Go name) + `JSON` (tag) split so a reader can map `m.LiveConnections` to the `live_connections` row.

guides/error_handling.md
```diff
@@ -208,32 +208,43 @@ if err != nil {
...
-`r.PartialFailures(mask errmask.ErrorMask) []error` returns every wrapper category that fired and was not suppressed by `mask`. Useful for recreating the dispatch's mask-gated behavior at the call site -- pass `errmask.Empty` to see every category, or any narrower mask to suppress specific ones:
+**Inspect categories with a `for`/`switch`** -- when partial error handling is appropriate. Partial error handling lets the client and its application recover from known failure modes they can tolerate (e.g. continue serving a search with a few failed shards, or retry only the bulk items the server rejected) instead of failing the whole operation. The `default` arm catches transport / HTTP / decode errors and any partial-failure category added in a future release:
 
 ```go
-for _, sub := range resp.PartialFailures(errmask.Empty) {
-    // same shape as opensearchapi.Errors(err)
+resp, err := client.MSearch(ctx, req)
+for _, sub := range opensearchapi.Errors(err) {
+    switch e := sub.(type) {
+    case *opensearchapi.PartialBulkError:
```
Jakob3xD: This switch is anchored to `client.MSearch`, but `*opensearchapi.PartialBulkError` (this line) and `*opensearchapi.ShardFailureError` (a few lines below) can never be returned by MSearch. Per this doc's own Error Type Reference table, `*PartialBulkError` is returned only by `Bulk`/`BulkStream` and `*ShardFailureError` only by `Index`/`Document.Create`/`Document.Delete`/`Update`. MSearch yields only `*PartialSearchError` / `*MultiSearchItemError` (the sibling "Inline iteration" example below uses exactly those two). Drop the two impossible cases, or de-anchor the example from `client.MSearch`.

guides/error_handling.md
```diff
@@ -208,32 +208,43 @@ if err != nil {
...
+    return err
 }
-
-resp2, _ := client.MSearch(ctx, req)
-if e := resp2.SearchShardFailures(); e != nil { /* ... */ }
-if e := resp2.MultiSearchItemFailures(); e != nil { /* ... */ }
-
-resp3, _ := client.Index(ctx, req)
-if e := resp3.WriteShardFailures(); e != nil { /* ... */ }
+// resp is fully populated; partial failures (if any) are folded into err.
```
Jakob3xD: `partial failures (if any) are folded into err` is not true under the v4 default. This guide uses the v4 import and sets no `Config.Errors`, so the default mask is `errmask.All` — on a partial failure `err` is `nil` and the failure is *not* folded into `err` (cf. this file's "Manual Partial Failure Checking" note: "When the relevant wrapper bits are masked (the v4 default)..."). Suggest qualifying: "...folded into `err` when the wrapper bits are unmasked (the v5preview default, or v4 with `Config.Errors: errmask.New()`)." (The identical comment in the v5preview README is correct, since v5 defaults to `errmask.Empty`.)

v5preview/opensearchapi/README.md
```diff
@@ -264,25 +264,42 @@ for _, sub := range opensearchapi.Errors(err) {
...
-if e := resp2.MultiSearchItemFailures(); e != nil { /* ... */ }
+**Inspect categories with a `for`/`switch`** -- when partial error handling is appropriate. Partial error handling lets the client and its application recover from known failure modes they can tolerate (e.g. continue serving a search with a few failed shards, or retry only the bulk items that the server rejected) instead of failing the whole operation. The `default` arm catches transport / HTTP / decode errors and any partial-failure category added in a future release:
 
-resp3, _ := client.Index(ctx, req)
-if e := resp3.WriteShardFailures(); e != nil { /* ... */ }
+```go
+resp, err := client.MSearch(ctx, req)
+for _, sub := range opensearchapi.Errors(err) {
+    switch e := sub.(type) {
+    case *opensearchapi.PartialBulkError:
```
Jakob3xD: Same dead-case issue as the `guides/error_handling.md` example: this MSearch switch includes `*opensearchapi.PartialBulkError` (this line) and `*opensearchapi.ShardFailureError` (below), which MSearch never returns — only `*PartialSearchError` / `*MultiSearchItemError`. This contradicts the Error Type Reference table above. Drop the two cases or de-anchor the example from `client.MSearch`.

(Note: the pre-existing "Inspecting errors" example earlier in this file has the same two impossible cases — worth fixing together.)

CHANGELOG.md
```diff
@@ -98,10 +99,10 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
...
-  - Per-Resp helper methods (`BulkItemFailures`, `SearchShardFailures`, `WriteShardFailures`, `MultiSearchItemFailures`) plus `PartialFailures(mask)` aggregator for focused inspection at the call site
-  - `opensearchapi.Errors(err) []error` package-level helper that flattens single- and multi-wrapper errors into a uniform slice for `switch` dispatch
+  - `opensearchapi.Errors(err) []error` package-level helper that flattens single- and multi-wrapper errors into a uniform slice; recommended call-site pattern is a `for`/`switch` over the result (not `errors.As` against a specific type)
   - Helper functions: `IsPartialFailure`, `ToleratePartialFailures`, `RequireSuccessRate` for threshold-based error tolerance
   - Operation constants: `OperationIndex`, `OperationCreate`, `OperationUpdate`, `OperationDelete`
+  - Per-Resp helper methods (`BulkItemFailures`, `SearchShardFailures`, `WriteShardFailures`, `MultiSearchItemFailures`, `PartialFailures(mask)`) exist on the response types as engine machinery for the dispatch; new code should prefer the `for`/`switch` pattern over `opensearchapi.Errors(err)` for forward compatibility
```
Jakob3xD: This reads backwards: "prefer the `for`/`switch` pattern over `opensearchapi.Errors(err)`" parses as "use a for/switch *instead of* `Errors(err)`", but the recommended pattern *is* a `for`/`switch` over the result of `opensearchapi.Errors(err)` — the thing to avoid is the per-Resp helper methods. Cf. `cmd/osgen/README.md`: "a `for`/`switch` over `opensearchapi.Errors(err)`, not the per-Resp helpers". Suggest: "...prefer a `for`/`switch` over `opensearchapi.Errors(err)` rather than the per-Resp helpers, for forward compatibility."

Jakob3xD (commented):
### Code review

Reviewed the docs changes, plus a cross-check against the still-open follow-up items from #844 (they live in the same files this PR edits). Seven issues are posted as inline comments above. Pre-existing and minor items are listed below.

Context: the earlier "blocked on #801" concern no longer applies — #801 (merged 2026-05-21) and #844 (merged 2026-06-04) landed the documented metrics + error-mask API on `main`, and CI on this head is green (the #819 `OptionalDataException` race fix merged 2026-04-27).

#### Open documentation follow-ups from #844 (the "D-series") — for the separate cleanup PR

These are all documentation items. They sit on lines this PR does not modify (D2's file isn't even part of this PR), so they can't be inline-anchored:

**D1** — `custom, _ := opensearchtransport.NewMuxRouter()` — `NewMuxRouter()` returns a single value (`Router`), so the two-value assignment won't compile:

https://github.com/opensearch-project/opensearch-go/blob/d87bb6edc545805762cedf13e001c23808221027/v5preview/opensearchapi/README.md#L353-L355

**D2** — the optional forward-compatible `replace` directive in MIGRATING.md points at a package *inside* the v4 module, not a standalone module. A versioned `replace ... => .../v4/v5preview/opensearchapi v4.7.0` won't resolve because that path has no `go.mod` of its own:

https://github.com/opensearch-project/opensearch-go/blob/d87bb6edc545805762cedf13e001c23808221027/v5preview/opensearchapi/MIGRATING.md#L29-L43

**D3** — v5preview `ErrorCause.Reason` is a `*string`; formatting it with `%s` prints a pointer address rather than the reason text. Needs a nil-checked deref (`*item.Error.Reason`):

https://github.com/opensearch-project/opensearch-go/blob/d87bb6edc545805762cedf13e001c23808221027/guides/error_handling.md#L94-L97

https://github.com/opensearch-project/opensearch-go/blob/d87bb6edc545805762cedf13e001c23808221027/guides/error_handling.md#L291-L293

**D5** — the pre-existing MSearch switch here carries the same impossible `*PartialBulkError` / `*ShardFailureError` cases flagged in the inline comments (this PR reintroduces the pattern in the new "Recommended pattern" blocks, so they are worth fixing together):

https://github.com/opensearch-project/opensearch-go/blob/d87bb6edc545805762cedf13e001c23808221027/v5preview/opensearchapi/README.md#L240-L244

#### Minor / optional

- `guides/metrics.md`: a few field descriptions assert mechanisms not evident from the struct definitions — "Marked overloaded by stats poller" (L85), `Estimated load: inFlight / cwnd` (L138), RTTBucket "lower is closer" (L100). Consider softening or verifying against the implementation.
- `CHANGELOG.md` L9: the metrics-guide bullet has no `(#…)` backlink, unlike its neighbours.
- `UPGRADING.md` L38: the intro says callers "can match … with `errors.As`", which sits in tension with the new paragraph that discourages `errors.As`.

guides/metrics.md
sean-: You're right, there is no `WithObserver()` in the module - the observer is wired through the `Observer` field on `opensearch.Config`. That section as written would not compile.

Fixed in the latest push: the paragraph now reads "...implement the `opensearchtransport.ConnectionObserver` interface and set it on the `Observer` field of `opensearch.Config`." Thanks for catching this before it landed in the docs.

guides/metrics.md
sean-: Right - a named function declaration inside the same fenced block as the `for range ticker.C { ... }` loop is invalid Go and the snippet wouldn't paste-and-run.

Fixed by hoisting `safeFloat` to the top of the same code block, ahead of `ticker := ...`. The example is now copy-paste runnable. Thanks.

guides/metrics.md
sean-: Agreed - mixing JSON tag names under "Field" in the four aggregate tables and Go names under "Field" in the per-struct tables makes "Field" mean two different things in the same doc, and breaks the link to `m.LiveConnections` in the polling example and the Prometheus mapping table.

Fixed: the four aggregate tables (Request Counters, Connection Pool State, Connection Lifecycle Counters, Health Check Counters) now use the same `Field` (Go name) + `JSON` (tag) split as `ConnectionMetric` / `PolicySnapshot` / `RouterSnapshot`. A reader can map `m.LiveConnections` straight to the `live_connections` row now. Thanks.

guides/error_handling.md
sean-: You're right - per the Error Type Reference table in this same file, MSearch can only ever return `*PartialSearchError` and `*MultiSearchItemError`. The `*PartialBulkError` and `*ShardFailureError` cases are dead and contradict the table above.

Fixed by dropping the two impossible cases from the example so it matches the reference table. The sibling "Inline iteration" example below was already correct on this point.

guides/error_handling.md
sean-: Good catch. You're right - this guide uses the v4 import and doesn't set `Config.Errors`, so the default mask is `errmask.All` and the partial failure is *not* folded into `err`. As written the comment is only true under v5preview's `errmask.Empty` default, or v4 with an explicit unmask.

Fixed by qualifying the comment:

```go
// resp is fully populated; partial failures (if any) are folded into err
// when the wrapper bits are unmasked (the v5preview default, or v4 with
// Config.Errors: errmask.New()).
```

The identical line in the v5preview README is left as-is, since v5 defaults to errmask.Empty and the bare claim is correct there. Thanks for catching this.

v5preview/opensearchapi/README.md
sean-: Same fix as the `guides/error_handling.md` thread - dropped `*PartialBulkError` and `*ShardFailureError` from this MSearch switch so it matches the Error Type Reference table.

I also fixed the pre-existing "Inspecting errors with `opensearchapi.Errors`" example earlier in this file (the same two impossible cases were there too) so both MSearch examples in this README agree with the reference table. Thanks for flagging both.

CHANGELOG.md
sean-: You're right, the wording reads as "use a `for`/`switch` *instead of* `Errors(err)`", which is the opposite of what the dispatch story actually is - the recommended pattern *is* a `for`/`switch` over `opensearchapi.Errors(err)`, and the thing to avoid is the per-Resp helpers.

Reworded to match the framing in `cmd/osgen/README.md`:

- Per-Resp helper methods (BulkItemFailures, ...) exist on the response types as engine machinery for the dispatch; new code should prefer a for/switch over opensearchapi.Errors(err) rather than the per-Resp helpers, for forward compatibility

Thanks - this one would have aged badly.

sean-:
Thanks for the careful cross-check, @Jakob3xD - especially walking the #844 follow-ups against the lines this PR edits. Latest push addresses the inline threads plus everything in your D-series and minor lists, in this PR rather than punting to a follow-up.

### #844 follow-ups

- **D1** (`v5preview/opensearchapi/README.md`) - `opensearchtransport.NewMuxRouter()` returns a single `Router`, so the two-value assignment is now `custom := opensearchtransport.NewMuxRouter()`.
- **D2** (`v5preview/opensearchapi/MIGRATING.md`) - removed the "Forward-compatible `replace` directive" section. The cited path has no `go.mod` of its own and the directive cannot resolve, so the section was actively misleading; better to drop it than to leave hallucinated guidance in the docs.
- **D3** (`guides/error_handling.md`) - `v5preview` `ErrorCause.Reason` is `*string`, so the v5preview bulk example now nil-checks and derefs the reason. The version-agnostic `handleMSearchError` example switched to logging `item.Status` instead, which is portable across both v4 and v5preview without per-version branching.
- **D5** - already addressed in the prior push as part of the inline `*PartialBulkError` / `*ShardFailureError` cleanup; both MSearch examples in `v5preview/opensearchapi/README.md` now agree with the Error Type Reference table.

### Minor list

- **`guides/metrics.md` field descriptions** - softened the three flagged claims so descriptions describe field meaning rather than how the field gets set: `IsOverloaded` is "Whether the connection is currently flagged overloaded", `RTTBucket`  is "Quantized RTT tier (smaller values indicate lower observed RTT)", `EstLoad` is "Estimated per-connection load". The `inFlight / cwnd` mechanism narrative now lives only in `guides/routing.md` where the implementation context is.
- **`CHANGELOG.md` L9 backlink** - added `([#812](...))` to the metrics-guide bullet for parity with its neighbours.
- **`UPGRADING.md` L37 `errors.As` tension** - reworded the intro so the body's type-switch recommendation is the only call-site pattern advertised. The intro no longer names `errors.As` as the entry point.

Jakob3xD (approved):

PR #804, opened 2026-03-20
Fix double-slash URL path bug across 74 GetRequest methods

sean-:
## Summary

When optional path segments (`Indices`, `Index`, `Repo`, `TaskID`, etc.) are empty, `GetRequest()` methods produced a double-slash (e.g. `//_alias/myalias`) which `http.NewRequest` misparses per [RFC 3986](https://www.rfc-editor.org/rfc/rfc3986#section-3.2), treating the next segment as the URL authority. After `setReqURL` overwrites the host, the path loses its API prefix entirely, causing requests to hit the wrong endpoint.

**Observed symptoms:**
- **#650**: `AliasPutReq` with empty `Indices` creates an index instead of an alias
- **#617**: `AliasGetReq` with empty `Indices` returns wrong error format (hits index lookup instead of alias lookup)

**Root cause:** 74 `GetRequest()` methods across `opensearchapi`, `plugins/security`, and `plugins/ism` unconditionally wrote `"/" + variable` into the URL path without guarding against empty values.

**Fix:** Extract `opensearch.BuildPath` helper that skips empty segments, preventing the double-slash class of bug. All 74 vulnerable methods converted to use it. Benchmarked at 1 alloc / ~21ns vs `path.Join`'s 3 allocs / ~63ns.

Fixes: #617, #650

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #815, opened 2026-03-27
Add configurable shard cost multipliers via env var and RouterOption

### LINKED ISSUE #814: Add configurable shard cost multipliers for connection scoring

The shard cost multipliers used in connection scoring (`costPreferred=1.0`, `costAlternate=2.0`, `costRelocating=8.0`, `costInitializing=16.0`, `costUnknown=32.0`) are compile-time constants. Operators who need to tune read/write routing bias.  For example, to equalize primary and replica preference, or to adjust penalties for relocating shards (as required by different workloads), library users must fork and rebuild the client.

Runtime-configurable shard cost multipliers via:

1. **Environment variable** `OPENSEARCH_GO_SHARD_COST`: evaluated once at client initialization.
2. **`WithShardCosts(spec)` RouterOption**: programmatic equivalent.
3. **`ShardCostConfig` field** on `Config`: passthrough for the `RouterOption`.

Precedence: env var > `RouterOption` > compile-time defaults. Any value ≤ 0 falls back to the compile-time default for that slot.

**Format:**

- **Bare numeric**: `1.5`: sets preferred and alternate to 1.5 in both read and write tables.
- **Abstract key=value**: `preferred=1.0,alternate=1.0`: role-aware keys that map to the correct shard type per table (preferred = replica for reads, primary for writes).
- **Prefixed key=value**: `r:replica=1.0,w:primary=0.5`: concrete shard type names scoped to a specific table.

Examples:
```bash
# Disable role preference (treat primary and replica equally)
OPENSEARCH_GO_SHARD_COST=preferred=1.0,alternate=1.0

# Tune read replicas and write primaries explicitly
OPENSEARCH_GO_SHARD_COST=r:replica=1.0,w:primary=0.5,alternate=1
```

### Additional context

This also consolidates all shard cost types, constants, tables, and methods into `shard_cost_config.go` for source code locality (previously spread across `policy_index_router.go`), introduces a `ShardCostConfigError` typed error for structured parse failure reporting, and extracts server-side thread pool name strings into constants referencing `ThreadPool.Names` in the OpenSearch server source.

sean-:
Introduce `OPENSEARCH_GO_SHARD_COST` environment variable, `WithShardCosts()` `RouterOption`, and `ShardCostConfig` passthrough field to override shard cost multipliers at runtime for connection scoring tuning.

Consolidate all shard cost types, constants, and tables into `shard_cost_config.go` for source code locality. Add `ShardCostConfigError` typed error for structured parse failure reporting.

Fixes: #814


sean-:
This test failure is effectively a flake and fixed in https://github.com/opensearch-project/opensearch-go/pull/821.  Steady-state clusters likely will never encounter this failure mode.

opensearchtransport/opensearchtransport.go
```diff
@@ -334,6 +334,17 @@ type Config struct {
...
+	// ShardCostConfig overrides shard cost multipliers used for connection scoring.
+	// Passed through to the router via [WithShardCosts] when constructing the
+	// default router. Format: bare numeric (sets r:base, the primary read cost
+	// at idle) or comma-separated key=value items. Dynamic keys are prefixed
+	// with "r:" (r:base, r:amplify, r:exponent) and control the read-primary
+	// cost curve. Static keys (unknown, relocating, initializing, replica,
+	// write_primary, write_replica) override shard state costs in the lookup
+	// tables.
+	// Can also be set via OPENSEARCH_GO_SHARD_COST (env var takes precedence).
+	ShardCostConfig string
```
Jakob3xD, line 355: This field is never read in `New()`. The transport assigns `cfg.Router` at line 812 but never references `cfg.ShardCostConfig` anywhere in `New()`. As a result, setting this field has no effect — the value is stored in the struct and immediately forgotten. See the overall review comment for the suggested fix.

opensearchtransport/shard_cost_config.go
```diff
@@ -0,0 +1,454 @@
...
+	// Key=value parsing.
+	for item := range strings.SplitSeq(spec, ",") {
+		item = strings.TrimSpace(item)
+		if item == "" {
+			continue
+		}
+
+		key, valStr, hasEq := strings.Cut(item, "=")
+		if !hasEq {
+			return shardCostConfig{}, &ShardCostConfigError{Key: key, Reason: "missing value"}
```
Jakob3xD: All three error-path returns in this function (`return shardCostConfig{}, ...` at lines 395, 400, and 432) return the zero value instead of `cfg`, which was already initialised with the compile-time defaults (`shardCostForReads` / `shardCostForWrites`) a few lines above. The caller in `NewDefaultPolicy` logs `"using defaults"` on error but actually receives all-zero tables, making every node score `0.0` and losing all shard cost differentiation. Change to `return cfg, &ShardCostConfigError{...}` on each error path.

Jakob3xD (commented):
The shard cost configuration architecture, pooled-allocation refactor, and multi-key routing are well-designed. Four issues need attention before merging.

> **Transparency note:** This review was created with the assistance of Claude (Anthropic) for transparency.

---

### Critical: `Config.ShardCostConfig` is never applied — the field has no effect

`opensearchtransport.Config.ShardCostConfig` (and `opensearch.Config.ShardCostConfig`) is defined with a doc comment saying it is "passed through to the router via `WithShardCosts`", but `New()` only assigns `cfg.Router` to `client.router` and never reads `cfg.ShardCostConfig`:

```go
// opensearchtransport/opensearchtransport.go:812
router: cfg.Router,  // ShardCostConfig is never consulted here
```

A user who sets `opensearch.Config{ShardCostConfig: "r:base=0.9"}` is silently ignored. The only working path today is constructing the router manually via `NewDefaultRouter(WithShardCosts("..."))`.

**Suggested fix** — in `New()`, when no explicit router is provided:
```go
if cfg.Router == nil && cfg.ShardCostConfig != "" {
    client.router = NewDefaultRouter(WithShardCosts(cfg.ShardCostConfig))
} else {
    client.router = cfg.Router
}
```

---

### Critical: Error paths in `parseShardCostConfig` return all-zero tables, not defaults

All three error-return statements in `parseShardCostConfig` (lines 395, 400, 432) return `shardCostConfig{}` (zero value) instead of `cfg`, which was already initialised with compile-time defaults on lines 366–369:

```go
cfg := shardCostConfig{
    reads:  shardCostForReads,  // value copy of defaults
    writes: shardCostForWrites, // value copy of defaults
}
```

Because the caller in `NewDefaultPolicy` always uses the returned value:

```go
costCfg, costErr := parseShardCostConfig(shardCostSpec)
if costErr != nil {
    dl.Logf("... using defaults: %v", costErr)  // logs "using defaults"
}
// costCfg is all-zero here — NOT the defaults
wrapWithRouter(..., &costCfg.reads, ...)
```

an invalid config string causes `calcConnDefaultScore` to return `rtt * util * 0.0 = 0.0` for every node, silently eliminating all shard cost differentiation while the log claims defaults are in use.

**Fix:** change each error return to `return cfg, &ShardCostConfigError{...}` so the already-populated defaults are preserved.

---

### Critical: Curve parameters accept `NaN`, `Inf`, and negative values without validation

The key=value parser for `r:base`, `r:amplify`, and `r:exponent` (lines 405–410) uses `strconv.ParseFloat` but applies no range validation:

```go
case "r:exponent":
    curveExponent = v   // v could be NaN, +Inf, -Inf, or negative
```

The bare-float path explicitly guards with `if v > 0`, but the key=value path does not. Invalid inputs propagate into `newReadScoreFunc` and from there into `math.Pow(writeUtil, exponent)`:

- `r:exponent=NaN` → `math.Pow(writeUtil, NaN) = NaN` → every connection scores `NaN`, routing breaks entirely
- `r:exponent=-2` with `writeUtil=0` → `math.Pow(0, -2) = +Inf` → all primary candidates score `+Inf`
- `r:base=-5` or `r:amplify=-1` → negative effective costs, preference inverted

**Fix:** after parsing each curve key, validate: `!math.IsNaN(v) && !math.IsInf(v, 0) && v >= 0`. For `r:exponent` specifically, also enforce `v >= 0`.

---

### Warning: Invalid env var configuration is completely silent in production

`NewDefaultPolicy` (router.go:638–640) only logs a parse error if a debug logger is active:

```go
if costErr != nil {
    if dl := loadDebugLogger(); dl != nil {
        dl.Logf("OPENSEARCH_GO_SHARD_COST: invalid config %q, using defaults: %v
", ...)
    }
    // no fallback log path
}
```

In production — where a debug logger is rarely configured — a completely invalid `OPENSEARCH_GO_SHARD_COST` value produces zero diagnostic output. Combined with the zero-value bug above, the operator sees correct-looking connections while shard cost routing is entirely non-functional.

At minimum, this should log to the standard logger (or `log.Printf`) unconditionally when the env var is set but unparseable.

---

### Minor: Bare-numeric example comment is misleading

The `envShardCost` godoc example says:
```
OPENSEARCH_GO_SHARD_COST=3.0  // r:amplify=3.0 (steeper curve)
```
But the format description directly above it says *"bare numeric (sets r:base)"*, and the parsing code confirms `curveBase = v`. A `r:base` of 3.0 makes primaries 3× more expensive than replicas at idle — the opposite of a steeper write-shedding curve. Should be corrected to show a representative `r:base` use case (e.g. `1.0` to equalize at idle, as the next example does).

---

### Positive observations

- The pooled-buffer approach (`pooledConns`, `pooledFloats`, `pooledNodeSet`) in `calcSingleKeyCost` / `calcMultiKeyCost` is clean and consistent with the allocation discipline established elsewhere in this package.
- `shardCostConfig` consolidating static tables + dynamic `connScoreFunc` in one parsed struct is good design — one value flows through the call stack instead of three.
- Test coverage in `shard_cost_config_internal_test.go` (429 lines) is thorough: empty spec, bare numeric, all key=value paths, zero-value clamping, and all error cases are exercised.
- `strings.SplitSeq` (Go 1.24 range-over-func) is fine given `go 1.24.0` in `go.mod`.
- The `IndexRouter` shard-exact path intentionally passes `scoreFunc=nil` — correct, since `IndexRouter` is a non-scored fallback policy; all scored routing goes through `poolRouter`.

opensearchtransport/opensearchtransport.go
sean-: Fixed. `New()` now constructs a default router when `cfg.Router == nil && cfg.ShardCostConfig != ""`:

```go
router := cfg.Router
if router == nil && cfg.ShardCostConfig != "" {
    var err error
    router, err = NewDefaultRouter(WithShardCosts(cfg.ShardCostConfig))
    if err != nil {
        cancel()
        return nil, fmt.Errorf("shard cost config: %w", err)
    }
}
```

This also ties into the error-surfacing fix below (invalid config now fails New() rather than being silently ignored).

opensearchtransport/shard_cost_config.go
sean-: Fixed with a stronger change: `parseShardCostConfig` now returns `(*shardCostConfig, error)` and returns `nil` on error. This makes the failure case unambiguous at the call site.  Callers can now `nil`-check rather than comparing against a zero-value struct that could be confused with a valid all-zeros config.

Additionally, `NewDefaultPolicy` and `NewDefaultRouter` now propagate the error upstream (see below), so the "log and continue with defaults" code path is removed entirely.

sean-:
Thanks for the review. All five items addressed in the fixup commits.

1) `Config.ShardCostConfig` wired up : `New()` now creates a default router when the field is set and no explicit router is provided. (See inline reply.)

2) Error paths fixed: `parseShardCostConfig` returns `(*shardCostConfig, error)` with `nil` on error, making the failure case unambiguous. (See inline reply.)

3) Curve param validation: `r:base`, `r:amplify`, and `r:exponent` now reject NaN, Inf, and negative values:

```go
case "r:base", "r:amplify", "r:exponent":
    if math.IsNaN(v) || math.IsInf(v, 0) || v < 0 {
        return nil, &ShardCostConfigError{
            Key:    key,
            Reason: "value must be a finite non-negative number",
        }
    }
```

4) Silent production failure: `NewDefaultPolicy` and `NewDefaultRouter` now return `(Policy, error)` and `(Router, error)` respectively. Parse errors propagate to the caller instead of being swallowed with a debug log. This is a breaking API change, but the functions are new in this PR so there are no existing consumers to migrate.

5) Misleading doc example: Fixed the `envShardCost` doc comment to show `r:base=1.0` (equal with replicas at idle) instead of the incorrect `r:amplify=3.0` annotation.

Jakob3xD (commented):
### Code review

Found 6 issue(s) — 4 posted as inline comments.

**Guide examples will not compile after API signature change**

`NewDefaultRouter()` and `NewDefaultPolicy()` now return `(Router, error)`. Multiple guide examples still use the old single-return form — anyone following them gets a compile error:

- `guides/routing.md` lines 12, 1498, 1835, 1974
- `guides/bulk.md` lines 56, 357, 398
- `guides/node_discovery_and_roles.md` line 203

Each needs `router, err := opensearchtransport.NewDefaultRouter(...)` with error handling.

https://github.com/opensearch-project/opensearch-go/blob/9e2bbd85f74cf805c173b2e3938b4d1a7e3f85e3/guides/routing.md#L12-L13

opensearchtransport/shard_cost_config.go
```diff
@@ -0,0 +1,486 @@
...
+
+	spec = strings.TrimSpace(spec)
+	if spec == "" {
+		cfg.scoreFunc = newReadScoreFunc(curveBase, curveAmplify, curveExponent)
+		return &cfg, nil
+	}
+
+	// Try bare numeric first: sets r:base (primary read cost at idle).
+	if v, parseErr := strconv.ParseFloat(spec, 64); parseErr == nil {
+		if v > 0 {
```
Jakob3xD: `+Inf` passes the `v > 0` guard: `strconv.ParseFloat("Inf", 64)` succeeds and `+Inf > 0` is true, so `curveBase = +Inf`. In `newReadScoreFunc`, any node with `primaryPct > 0` then scores `+Inf`, silently routing all reads to replica-only nodes. The key=value path correctly rejects this with `math.IsInf(v, 0)` at line 429. Fix: `if v > 0 && !math.IsInf(v, 0) && !math.IsNaN(v)`.

opensearchtransport/shard_routing.go
```diff
@@ -6,66 +6,249 @@
...
+		return &s
+	},
+}
+
+func acquireScoredConns() *[]scoredConn {
+	return scoredConnsPool.Get().(*[]scoredConn) //nolint:forcetypeassert // pool only stores *[]scoredConn
+}
+
+func releaseScoredConns(s *[]scoredConn) {
+	*s = (*s)[:0]
```
Jakob3xD, line 207: `releaseScoredConns` truncates the length but does not call `clear((*s)[:cap(*s)])` before `Put`. `scoredConn` contains a `*Connection` pointer; without clearing, the pooled backing array retains those references across GC cycles, keeping `Connection` objects alive longer than necessary. `putConnSlice` and `pooledConns.Release` both call `clear()` before Put — this function should do the same.

Fix: add `clear((*s)[:cap(*s)])` before `*s = (*s)[:0]`.

opensearchtransport/policy_index_router.go
```diff
@@ -600,17 +468,28 @@ func connScoreSelect(
...
+			primaryPct = calcNodePrimaryPct(nodeInfo)
 		default:
-			// Cluster lookup: no shard data. All nodes get costUnknown;
-			// equal cost cancels out, so selection is by RTT + congestion.
-			scores[i] = calcConnScore(c, costUnknown, poolName, poolInfoReady)
+			sc = costUnknown
+		}
+
+		if extraCost != nil {
+			sc += extraCost[i]
```
Jakob3xD: `extraCost[i]` (miss penalty for nodes not holding the routing key) is added to `sc` before `scoreFunc` receives it. In `newReadScoreFunc`, the blend is `cost = (1-primaryPct)*shardCost + primaryPct*dynamicCost`. When `primaryPct == 1.0`, `shardCost` — including the miss penalty — is dropped entirely and replaced by `dynamicCost`. A node holding none of the routing keys but hosting all-primary shards (dynamic cost ≈ 0.95 at idle) is incorrectly preferred over a replica-only node holding all keys (static cost = 1.0).

Fix: apply `extraCost` to the final score after `scoreFunc`, not inside `sc` before it.

opensearchtransport/opensearchtransport.go
```diff
@@ -747,6 +758,16 @@ func New(cfg Config) (*Client, error) {
 	}
 	ctx, cancel := context.WithCancel(parent)
 
+	router := cfg.Router
+	if router == nil && cfg.ShardCostConfig != "" {
```
Jakob3xD: Two silent-ignore issues at this condition:

1. **`OPENSEARCH_GO_SHARD_COST` env var has no effect** when `cfg.ShardCostConfig` is empty (the default). `NewDefaultRouter` — which reads the env var — is never called unless `cfg.ShardCostConfig != ""`. The field comment says "env var takes precedence", implying standalone env var support. Fix: also create the default router when `os.Getenv(envShardCost) != ""`.

2. **`cfg.ShardCostConfig` is silently discarded** when `cfg.Router != nil`. No error is returned and the field godoc does not mention this mutual-exclusion. Either document it explicitly or return an error when both are set.

opensearchtransport/shard_cost_config.go
sean-: Fixed in 04197335. The bare numeric path now also guards against `+Inf` and `NaN`:

```go
if v > 0 && !math.IsInf(v, 0) && !math.IsNaN(v) {
    curveBase = v
}
```

This makes both the bare numeric and `key=value` paths consistent. Test cases for bare `+Inf` uses defaults and bare `NaN` uses defaults were added to `shard_cost_config_internal_test.go`, too.

opensearchtransport/shard_routing.go
sean-: Fixed in 0de15704. `releaseScoredConns` now clears the backing array before returning the slice to the pool, matching the pattern used by `putConnSlice` and `pooledConns.Release`:

```go
func releaseScoredConns(s *[]scoredConn) {
    clear((*s)[:cap(*s)])
    *s = (*s)[:0]
    scoredConnsPool.Put(s)
}
```

This drops the retained *Connection references so they can be collected on the next GC cycle.

opensearchtransport/policy_index_router.go
sean-: Fixed in 0de15704. The miss penalty is now applied to the final score after scoreFunc, so the primary blend formula can no longer drop it:

```go
if scoreFunc != nil {
    scores[i] = scoreFunc(c, sc, primaryPct, poolName, poolInfoReady)
} else {
    scores[i] = calcConnDefaultScore(c, sc, poolName, poolInfoReady)
}

if extraCost != nil {
    scores[i] += extraCost[i]
}
```

A replica-only node holding all the routing keys (static cost = 1.0) now correctly scores below an all-primary node holding none of them (0.95 + miss_penalty), regardless of primaryPct.

opensearchtransport/opensearchtransport.go
sean-: Both addressed by reframing the trigger model. Two related changes:

1. Router-creation trigger is now `OPENSEARCH_GO_ROUTER`, not `ShardCostConfig` (54797cf8)

In v4 the `DefaultRouter` is *off* by default. Setting `OPENSEARCH_GO_ROUTER=true` is now the way to enable auto-construction without code changes; in v5 the default flips, and the env var is removed in v6. `OPENSEARCH_GO_SHARD_COST` is read internally during router construction:

```go
router := cfg.Router
if router == nil {
    if val, ok := os.LookupEnv(envRouter); ok && val != "" {
        if b, err := strconv.ParseBool(val); err == nil && b {
            var opts []RouterOption
            if cfg.ShardCostConfig != "" {
                opts = append(opts, WithShardCosts(cfg.ShardCostConfig))
            }
            router, err = NewDefaultRouter(opts...)
            if err != nil {
                cancel()
                return nil, fmt.Errorf("shard cost config: %w", err)
            }
        }
    }
}
```

2. Mutual exclusion documented

The ShardCostConfig godoc now spells out the precedence rules explicitly:

```
//  Consumed only when a router is being constructed:
//
// - When Config.Router is set, this field is ignored (the user's router was
// already constructed with its own configuration).
// - When Config.Router is nil and OPENSEARCH_GO_ROUTER=true triggers
// auto-construction, this field is applied via [WithShardCosts].
// - Otherwise (no router), this field has no effect.
//
// When a router is constructed, OPENSEARCH_GO_SHARD_COST takes precedence over
// this field for the same-named parameters.
```

Allowing post-construction injection via a `SetShardCosts` method on `Router` requires a wider refactor (i.e. atomic swap of `*shardCostMultiplier` pointers, `scoreFunc` indirection in `poolRouter`). Deferred for now as the doc-only resolution is sufficient for v4.

sean-:
Fixed in 54797cf8 and 2c21d57a. All guide examples now use the two-return form with error handling:

```go
router, err := opensearchtransport.NewDefaultRouter()
if err != nil {
    log.Fatal(err)
}
```

opensearchtransport/pool_buffers.go
```diff
@@ -0,0 +1,104 @@
+// SPDX-License-Identifier: Apache-2.0
```
VijayanB, line 1: nit: can you add dedicated test for this file?

sean-: Done.  Added `opensearchtransport/pool_buffers_internal_test.go` covering acquire length/cap-grow paths, zero-value safety on `Slice`/`Len`/`Release`, the pointer-clearing behavior in `pooledConns.Release`.

opensearchtransport/shard_cost_config.go
```diff
@@ -0,0 +1,486 @@
...
+
+	spec = strings.TrimSpace(spec)
+	if spec == "" {
+		cfg.scoreFunc = newReadScoreFunc(curveBase, curveAmplify, curveExponent)
+		return &cfg, nil
+	}
+
+	// Try bare numeric first: sets r:base (primary read cost at idle).
+	if v, parseErr := strconv.ParseFloat(spec, 64); parseErr == nil {
+		if v > 0 && !math.IsInf(v, 0) && !math.IsNaN(v) {
```
Jakob3xD: The bare-numeric path silently accepts `v <= 0` (e.g. `OPENSEARCH_GO_SHARD_COST=-1.0` or `=0`): `v > 0` is false, so `curveBase` stays at the default 0.95, and the function returns `nil` error. The key=value path (`r:base=-1.0`) correctly returns a `ShardCostConfigError` with `shardCostReasonNotFinite`.

The previous fix (commit 04197335) added `+Inf`/`NaN` guards but left negative and zero values as a silent fallback. Suggest returning an error for the same cases the key=value path rejects:
```go
if v <= 0 || math.IsInf(v, 0) || math.IsNaN(v) {
    return nil, &ShardCostConfigError{Reason: shardCostReasonNotFinite}
}
curveBase = v
```

Jakob3xD (commented):
### Code review

Found 5 issue(s) — all posted as inline comments.

opensearchtransport/opensearchtransport.go
```diff
@@ -747,6 +766,24 @@ func New(cfg Config) (*Client, error) {
...
+			if b, err := strconv.ParseBool(val); err == nil && b {
+				var opts []RouterOption
+				if cfg.ShardCostConfig != "" {
+					opts = append(opts, WithShardCosts(cfg.ShardCostConfig))
+				}
+				var newErr error
+				router, newErr = NewDefaultRouter(opts...)
+				if newErr != nil {
+					cancel()
+					return nil, fmt.Errorf("shard cost config: %w", newErr)
```
Jakob3xD: `ShardCostConfigError.Error()` (shard_cost_config.go:55) already returns a string prefixed with `"shard cost config: "`. Wrapping it with `fmt.Errorf("shard cost config: %w", newErr)` produces a doubled prefix:

```
shard cost config: shard cost config "r:base=-1": value must be a finite non-negative number
```

Suggest either returning `newErr` directly, or using a non-overlapping wrapper like `fmt.Errorf("auto-constructing router: %w", newErr)`.

CHANGELOG.md
```diff
@@ -8,6 +8,11 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
...
+- Add dynamic read cost scoring: primary shard cost scales with write-pool utilization via `connScoreFunc`, preferring primaries at idle and shedding reads to replicas under write load
+- Add `OPENSEARCH_GO_SHARD_COST` environment variable and `WithShardCosts()` router option with `r:base`/`r:amplify`/`r:exponent` curve keys and static cost overrides
+- Add `ShardCostConfig` field to `Config` struct for programmatic shard cost override passthrough
+- Add `OPENSEARCH_GO_ROUTER` environment variable to enable the DefaultRouter without code changes; set to `true` to opt in (off by default in v4, on by default in v5, removed in v6) ([#816](https://github.com/opensearch-project/opensearch-go/issues/816))
+- Add `primary_terms_map` and `split_shards_metadata` fields to ClusterState index metadata for OpenSearch >=3.6.0 compatibility
```
Jakob3xD: Duplicate entry — identical to line 9: "Add `primary_terms_map` and `split_shards_metadata` fields to ClusterState index metadata for OpenSearch >=3.6.0 compatibility". One of the two should be removed.

guides/routing.md
```diff
@@ -1877,6 +1943,71 @@ OPENSEARCH_GO_DISCOVERY_CONFIG=-routing_num_shards,-node_stats
...
+if err != nil {
+    log.Fatal(err)
+}
+```
+
+Or via the top-level client config (passthrough):
+
+```go
+client, _ := opensearch.NewClient(opensearch.Config{
+    ShardCostConfig: "r:base=0.90,r:amplify=2.5",
```
Jakob3xD, line 2008: The "passthrough" example shows `Config{ShardCostConfig: "r:base=0.90,r:amplify=2.5"}` as a standalone snippet, but `ShardCostConfig` is only consumed when `OPENSEARCH_GO_ROUTER=true` triggers auto-construction of the default router (opensearchtransport.go, env var check). Without that env var set, this field is silently ignored.

Suggest adding a note, e.g.:
```go
// Requires OPENSEARCH_GO_ROUTER=true (or Config.Router set) — otherwise ShardCostConfig has no effect.
client, _ := opensearch.NewClient(opensearch.Config{
    ShardCostConfig: "r:base=0.90,r:amplify=2.5",
})
```

opensearchtransport/shard_cost_config.go
```diff
@@ -0,0 +1,486 @@
...
+	shardCostKeyWritePrimary = "write_primary"
+	shardCostKeyWriteReplica = "write_replica"
+)
+
+// ShardCostConfigError is returned when [parseShardCostConfig] encounters
+// an invalid shard cost configuration string.
+type ShardCostConfigError struct {
+	// Key is the config key that caused the error (may be empty for bare values).
+	Key string
+	// Reason classifies the failure: shardCostReasonUnknownKey, shardCostReasonInvalid, or shardCostReasonMissing.
```
Jakob3xD: The doc comment lists only three possible `Reason` values (`shardCostReasonUnknownKey`, `shardCostReasonInvalid`, `shardCostReasonMissing`) but a fourth — `shardCostReasonNotFinite` — is actively returned at line 432. Suggest adding it to the list.

opensearchtransport/shard_cost_config.go
sean-: Fixed in f044ee16. The bare-numeric path now rejects the same set as the `key=value` path:

```go
if v <= 0 || math.IsInf(v, 0) || math.IsNaN(v) {
    return nil, &ShardCostConfigError{Reason: shardCostReasonNotFinite}
}
curveBase = v
```

`OPENSEARCH_GO_SHARD_COST=0`, `=-1.0`, `=+Inf`, and `=NaN` now return `ShardCostConfigError{Reason: shardCostReasonNotFinite}` instead of silently falling back to the 0.95 default.  Since there is no backwards compatibility behavior for this, I would rather someone wire up the library so the process won't start if someone misconfigures this env var because if someone misconfigure it and it silently uses the default, that could be worse in an incident.

opensearchtransport/opensearchtransport.go
sean-: Fixed in f044ee16. `New()` now returns `newErr` directly without wrapping, since `ShardCostConfigError.Error()` already prefixes with `"shard cost config: "`:

```go
router, newErr = NewDefaultRouter(opts...)
if newErr != nil {
    cancel()
    return nil, newErr
}
```

The error message is now `shard cost config "r:base=-1": value must be a finite non-negative number` (single prefix). Callers can still type-assert with `errors.As(err, &scErr)` to recover the structured `ShardCostConfigError`.

CHANGELOG.md
sean-: Fixed in f044ee16. The duplicate entry on line 15 was removed; the entry on line 9 remains. Net change is `-1` line.  The hazards of branch management and repo maintenance are real.  Thank you.

guides/routing.md
sean-: Fixed in f044ee16. The passthrough example now states the requirement inline:

> Or via the top-level client config (passthrough). Requires `OPENSEARCH_GO_ROUTER=true` (or `Config.Router` set) to take effect; otherwise `ShardCostConfig` is silently ignored:
>
> ```go
> client, _ := opensearch.NewClient(opensearch.Config{
>     ShardCostConfig: "r:base=0.90,r:amplify=2.5",
> })
> ```

This matches the precedence rules already documented on the `ShardCostConfig` field godoc.

opensearchtransport/shard_cost_config.go
sean-: Fixed in f044ee16:

```go
// Reason classifies the failure: shardCostReasonUnknownKey, shardCostReasonInvalid, shardCostReasonMissing, or shardCostReasonNotFinite.
Reason string
```

All four reason constants are now listed.

Jakob3xD (commented):
### Code review

Found 3 issue(s):
`DiscoverNodesOnStart bool→*bool` (commit `22615c5b`) is a breaking API change — existing `Config{DiscoverNodesOnStart: true}` fails to compile. No `### Changed` entry exists in [Unreleased] for this, even though other breaking changes in the same block (e.g. `context.Context` parameter additions) are documented.

https://github.com/opensearch-project/opensearch-go/blob/f044ee16057d6a32218ac8b968f4ddf66bd25643/CHANGELOG.md#L136-L160

opensearchtransport/doc.go
```diff
@@ -196,6 +200,23 @@ logging globally without code changes. When enabled, debug output is written to
...
+# Enabling the Router via Environment Variable
+
+In v4 the DefaultRouter is off by default. Set the OPENSEARCH_GO_ROUTER environment
+variable to enable scored, role-aware request routing without code changes:
+
+	OPENSEARCH_GO_ROUTER=true
+
+The value is parsed with strconv.ParseBool (accepts "true", "1", "t", etc.). When set
+to a truthy value and no programmatic Config.Router is provided, [New] automatically
+creates a [NewDefaultRouter] and sets DiscoverNodesOnStart to true so topology is
```
Jakob3xD: `DiscoverNodesOnStart` is a field of `opensearch.Config`, not `opensearchtransport.Config`. The transport layer has no such field — it is set exclusively by `opensearch.NewClient()`. Users calling `opensearchtransport.New()` directly with `OPENSEARCH_GO_ROUTER=true` get a router but not auto-discovery on start, contradicting this doc.

opensearch.go
```diff
@@ -322,7 +328,18 @@ func NewClient(cfg Config) (*Client, error) {
...
 	}
 
-	if cfg.DiscoverNodesOnStart {
+	// When the caller did not set DiscoverNodesOnStart and no programmatic
+	// Router was provided, inherit the value from the OPENSEARCH_GO_ROUTER
+	// env var so that router enablement also implies on-start discovery.
+	if cfg.DiscoverNodesOnStart == nil && cfg.Router == nil {
+		if val, ok := os.LookupEnv(envRouter); ok && val != "" {
+			if b, err := strconv.ParseBool(val); err == nil {
+				cfg.DiscoverNodesOnStart = &b
```
Jakob3xD: `strconv.ParseBool` succeeds for both `\"true\"` and `\"false\"`, so `OPENSEARCH_GO_ROUTER=false` sets `cfg.DiscoverNodesOnStart = &false` instead of leaving it `nil`. The field's documented semantics (`nil = auto`) are violated — code checking `cfg.DiscoverNodesOnStart == nil` to detect \"user did not set this\" will get the wrong answer. The transport layer guards the same env var with `err == nil && b` (acting only on `true`). Suggest the same guard here:
```go
if b, err := strconv.ParseBool(val); err == nil && b {
    cfg.DiscoverNodesOnStart = &b
}
```

opensearchtransport/doc.go
sean-: Fixed in 87da5b6c. The doc no longer claims `opensearchtransport.New` touches `DiscoverNodesOnStart` (it can't: the field doesn't exist at the transport layer - I'm pretty sure it used to, but was removed due to an import cycle). The section now states what `New` actually does (constructs a `NewDefaultRouter`) and notes separately that `opensearch.NewClient` observes the same env var to enable on-start discovery, which is where that behavior actually lives:

> The value is parsed with strconv.ParseBool (accepts "true", "1", "t", etc.). When
> set to a truthy value and no programmatic Config.Router is provided, [New]
> automatically creates a [NewDefaultRouter]. Programmatic Config.Router always
> takes precedence.
>
> The opensearch package observes the same variable independently: when
> OPENSEARCH_GO_ROUTER is truthy and opensearch.Config.DiscoverNodesOnStart has
> not been explicitly set, opensearch.NewClient enables on-start topology
> discovery so the router starts with fresh node information. opensearchtransport
> itself has no DiscoverNodesOnStart field; callers constructing the transport
> directly must trigger discovery themselves if desired.

Callers using `opensearchtransport.New` directly now get an accurate description of what the env var does at that layer.

opensearch.go
sean-: Fixed in `3b01c2f8`. Adopted your suggested guard so only an explicit truthy value flips `DiscoverNodesOnStart`; any other value (including `false`, unparseable, or empty) leaves it `nil` per the documented semantics:

```go
if cfg.DiscoverNodesOnStart == nil && cfg.Router == nil {
    if val, ok := os.LookupEnv(envRouter); ok && val != "" {
        if b, err := strconv.ParseBool(val); err == nil && b {
            cfg.DiscoverNodesOnStart = &b
        }
    }
}
```

This now matches the guard already used at the transport layer (`opensearchtransport.go:771`), so both env var observers agree on the "act only on truthy" rule. `OPENSEARCH_GO_ROUTER=false` correctly leaves `cfg.DiscoverNodesOnStart` as `nil`, and code checking `cfg.DiscoverNodesOnStart == nil` to detect "user did not set this" gets the right answer.

opensearchtransport/policy_misc_internal_test.go
```diff
@@ -77,3 +77,51 @@ func TestWithShardExactRouting(t *testing.T) {
...
+
+	tests := []struct {
+		name    string
+		initial string // pre-existing shardCostConfig value
+		spec    string // argument to WithShardCosts
+		want    string // expected shardCostConfig after applying
+	}{
+		{
+			name: "stores spec in config",
+			spec: "preferred=1.0,alternate=1.0",
```
Jakob3xD: `WithShardCosts` stores the string verbatim in `routerConfig.shardCostConfig` without parsing, so this test only validates field assignment — not parser correctness. The keys `preferred`, `alternate`, `r:replica`, and `w:primary` are unknown to `parseShardCostConfig` and would produce a `ShardCostConfigError` if passed to `NewDefaultRouter`. Same applies to lines 102–103 and 107–115.

Consider replacing with valid keys (`r:base`, `r:amplify`, `r:exponent`, `replica`, `write_primary`, `write_replica`) so the tests exercise the round-trip through the parser.

opensearchtransport/shard_cost_config.go
```diff
@@ -0,0 +1,487 @@
...
+	// Read cost constants.
+	costReadReplica      = 1.0  // preferred -- lock-free Lucene snapshot
+	costReadPrimary      = 2.0  // static fallback; overridden by dynamic scoring
+	costReadRelocating   = 8.0  // shard moving, may require proxy hop
+	costReadInitializing = 16.0 // shard not yet ready to serve
+
+	// Write cost constants. Writes always target the primary shard;
+	// any other node must coordinate to the primary, which is avoidable work.
+	costWritePrimary      = 1.0  // preferred -- write lands directly on primary
+	costWriteReplica      = 8.0  // must coordinate to primary -- unnecessary proxy
```
Jakob3xD, line 132: `costWriteReplica` was raised from `2.0` to `8.0` (4×) in this PR, but CHANGELOG.md has no mention of this change. For users enabling the DefaultRouter this silently quadruples the routing penalty for writes to replica nodes — a behavioral breaking change in the scoring model. Please add a CHANGELOG entry.

opensearchtransport/policy_index_router.go
```diff
@@ -292,9 +171,9 @@ func (p *IndexRouter) Eval(_ context.Context, req *http.Request) (NextHop, error
 			return NextHop{}, nil
 		}
 
-		scoreBuf := acquireScoreSlice(len(conns))
-		best := connScoreSelect(conns, nil, nil, p.shardCosts, "", loadPoolInfoReady(p.config.poolInfoReady), *scoreBuf)
-		releaseScoreSlice(scoreBuf)
+		scores := acquireFloats(len(conns))
+		best := connScoreSelect(conns, nil, nil, p.shardCosts, "", loadPoolInfoReady(p.config.poolInfoReady), scores.Slice(), nil, nil)
```
Jakob3xD: `connScoreSelect` is called with `scoreFunc: nil` here and at all other call sites in this file (lines 211, 268). `IndexRouter` has no `scoreFunc` field, so the dynamic read-cost function (`newReadScoreFunc`) introduced by this PR is never applied on the index-routing path. `DocRouter` has the same gap (`policy_doc_router.go` lines 156, 210).

If omitting dynamic scoring here is intentional (e.g. consistency semantics take priority), please document why — otherwise this looks like incomplete wiring of the PR's core feature.

guides/routing.md
```diff
@@ -1877,6 +1943,71 @@ OPENSEARCH_GO_DISCOVERY_CONFIG=-routing_num_shards,-node_stats
...
+```go
+router, err := opensearchtransport.NewDefaultRouter(
+    opensearchtransport.WithShardCosts("r:base=0.90,r:amplify=2.5"),
+)
+if err != nil {
+    log.Fatal(err)
+}
+```
+
+Or via the top-level client config (passthrough). Requires `OPENSEARCH_GO_ROUTER=true` (or `Config.Router` set) to take effect; otherwise `ShardCostConfig` is silently ignored:
```
Jakob3xD: This passthrough example is missing a prerequisite: `opensearch.Config.ShardCostConfig` is silently ignored unless `OPENSEARCH_GO_ROUTER=true` is set (so the router is auto-constructed). If `Config.Router` is set explicitly, `ShardCostConfig` is also ignored — the router was already built without it.

Suggested addition: "Note: this field is only applied when the DefaultRouter is auto-constructed via `OPENSEARCH_GO_ROUTER=true`. It has no effect when `Config.Router` is set, or when no router is configured."

Jakob3xD (commented):
### Code review

Found 5 issue(s) — 4 as inline comments, 1 below.

---

**`opensearchtransport/policy_override_env.go` (line 24) — psIsEnabled resolution table is misleading [score: 100]**

The resolution priority table at line 24 claims `psEnvEnabled → true (operator says on)`, but `psIsEnabled()` (line 44) never reads the `psEnvEnabled` bit — it only checks `psEnvDisabled` and `psEnabled`. The function's own doc comment at line 42 confirms this: "psEnvEnabled is recorded for observability but does not alter resolution."

Setting `OPENSEARCH_GO_POLICY_*=true` has no routing effect. The table entry `psEnvEnabled → true` should be removed or annotated as observability-only.

https://github.com/opensearch-project/opensearch-go/blob/3b01c2f866579830044352133df1a06130e999f7/opensearchtransport/policy_override_env.go#L23-L49

---

opensearchtransport/policy_misc_internal_test.go
sean-: Fixed during the adversarial-review rebase. The test now uses keys that round-trip cleanly through `parseShardCostConfig`:

- `replica` -> `shardCostKeyReplica`
- `write_primary` -> `shardCostKeyWritePrimary`
- `r:base` / `r:amplify` curve keys

See the `Add configurable shard cost multipliers via env var and RouterOption` commit for additional  details.

opensearchtransport/shard_cost_config.go
sean-: Same answer as the `costWriteReplica` change: the `bool` -> `*bool` field-type swap is wholly within `[Unreleased]`. The original `bool` field is itself part of the unreleased v4 line — no released version of the client ever exposed the `bool` form, so there's no consumer-facing API delta between releases for a CHANGELOG entry to describe.

opensearchtransport/policy_index_router.go
sean-: Fixed.  I think some work got lost along the way doing rebases and cherry-picking commits and maintaining PRs.  `IndexRouter` and `DocRouter` have been updated to carry `scoreFunc connScoreFunc` and `poolName string` fields, threaded through every `connScoreSelect` call site (3 in `IndexRouter`, 2 in `DocRouter`). `IndexRouter`'s rendezvous-fallback path also computes adaptive MCSR (`computeAdaptiveConcurrency`) when `poolName == poolSearch`, matching `poolRouter.Eval`.

The constructors became externally usable in the process:

```go
// public, options-based, returns error if shard-cost spec invalid
func NewIndexRouter(opts ...RouterOption) (*IndexRouter, error)
func NewDocRouter(opts ...RouterOption) (*DocRouter, error)
```

Both honor the same `WithShardCosts`, `WithMinFanOut`, `WithMaxFanOut`, `WithIndexFanOut`, `WithFanOutPerRequest`, `WithDecayFactor`, `WithIdleEvictionTTL`, `WithShardExactRouting`, `WithAdaptiveConcurrency`, and `WithAdaptiveConcurrencyLimits` options as `NewDefaultRouter`, plus the same env-var precedence rules. The previous unexported `indexSlotCacheConfig` parameter type made the constructors externally unusable — that's now an internal helper (`newIndexRouter` / `newDocRouter`) for tests.

Folded into the `Add dynamic read cost scoring with write-pool-aware primary shedding` commit, alongside three shared helpers (`buildStandaloneRouterConfig`, `indexSlotCacheConfigFromRouter`, plus the `wrapWithRouter` plumbing) so `NewDefaultPolicy`, `NewIndexRouter`, and `NewDocRouter` use the same option/env precedence path.

guides/routing.md
sean-: Fixed. The passthrough example now states the precedence inline:

> Or via the top-level client config (passthrough). Only applied when the DefaultRouter is auto-constructed via `OPENSEARCH_GO_ROUTER=true`; the field is silently ignored when
`Config.Router` is set (the router was already built without it) or when no router is configured

Matches the precedence rules already documented on the `ShardCostConfig` field godoc in `opensearchtransport/opensearchtransport.go`. Folded into the `feat(transport): add OPENSEARCH_GO_ROUTER env var` commit.

sean-:
> The resolution priority table at line 24 claims `psEnvEnabled → true (operator says on)`, but `psIsEnabled()` (line 44) never reads the `psEnvEnabled` bit — it only checks `psEnvDisabled` and `psEnabled`. The function's own doc comment at line 42 confirms this: "psEnvEnabled is recorded for observability but does not alter resolution."

Fixed. The resolution-priority table no longer claims `psEnvEnabled -> true`; it now lists only the bits `psIsEnabled` actually consults (`psEnvDisabled`, `psEnabled`, default-false), and a separate paragraph below the table explains `psEnvEnabled` in the affirmative — it's set by `OPENSEARCH_GO_POLICY_*=true` so observers can attribute the active state to "operator opted in" vs. "default behavior".

`psIsEnabled`'s own godoc was also tightened to repeat the observability note in the same affirmative phrasing. Folded into the `chore(transport): clarify psIsEnabled doc and apply goimports drift` commit.

sean-:
Before pushing, I ran an adversarial commit-by-commit review of the branch to proactively surface other issues from the fidelity loss as these PRs have been rebased and maintained across repos, and folded those findings back into the original commits via `git commit --fixup` and `git rebase -i --autosquash`.

Beyond the per-comment fixes above, the rebase also addressed:

- `ShardCostConfigError` API surface: `Reason` + `Detail` collapsed into a single `Message` field; tests use `require.ErrorContains` against the public `Error()` instead of struct-field equality.
- `shardCostReason*` constants renamed to `shardCostMessage*` to match the new field name. The long unknown-key error string was extracted into `shardCostValidKeysHint` to keep the line under the linter's 140-char limit.
- `OPENSEARCH_GO_*` env-var names consolidated into a new `internal/envvars` package shared between `opensearch` and `opensearchtransport`. Avoids the import cycle that previously forced two parallel constants for `OPENSEARCH_GO_ROUTER`.
- New table-driven test for `OPENSEARCH_GO_ROUTER` auto-construction covering env-unset, `=true`, `=1`, `=false`, garbage, programmatic-router-precedence, and invalid-`ShardCostConfig`-returns-error paths.
- `pooledNodeSet` moved from `shard_routing.go` to `pool_buffers.go` so all pooled-buffer types live together; `acquireNodeSet`'s godoc tightened.
- `getConnSlice` (append-style) and `acquireConns` (indexed-style) co-located in `pool_buffers.go` with a header comment explaining the access-pattern split, addressing what could otherwise read as a half-finished refactor.

Build clean, all unit tests pass, integration tests pass for the non-flaky portion of the matrix.

---

The CI failures on `integ-test-compat (false, 1.3.20)` and `integ-test-compat (false, 2.3.0)` are the same `TestShardExactRouting_FullPipeline_Integration` flake I noted earlier in [#issuecomment-4254379251](https://github.com/opensearch-project/opensearch-go/pull/815#issuecomment-4254379251). The signature is unchanged: a partial pass/fail split within a single run, where some routing values resolve shard-exact and others race with discovery.

```
PASS routing=user42, order-12345, 0, tenant:acme, deterministic-key
FAIL routing=hello, abc, 1, session_9999
```

The end-to-end readiness probe added in [`7b43d507`](https://github.com/opensearch-project/opensearch-go/commit/7b43d507b5338368fd6d76a192c9dc048ad52b8f) verifies a single canary routing value resolves before the test proceeds, but does not gate on every shard's mapping converging. On older OpenSearch versions (1.3.20, 2.3.0) the discovery cycle following the canary probe occasionally clears partial shard-map state for other shards, producing the mixed pass/fail.

**Permanent fix lands in #804** ([gh-650](https://github.com/opensearch-project/opensearch-go/pull/804)). Three commits on that branch address the root causes:

- `Add version guards and http.Status constants to shard routing tests` (skips shard-routing tests on OpenSearch < 2.0 where `routing_num_shards` is unavailable from `_cluster/state/metadata`)
- `Add internal/test/readiness package for layered FSM-based test gating` (per-node FSM packing `LayerHTTP` -> `LayerClusterJoin` -> `LayerStatsReady` into one atomic, replacing the structural-only readiness check)
- `Replace WaitForAllNodesReady inline polling with readiness.Wait` (migrates `TestShardExactRouting_FullPipeline_Integration` and friends to the new layered gate)

#804 is still rebasing so SHAs are subject to change; the commit subjects above will survive the rebase.

opensearchtransport/router.go
```diff
@@ -671,7 +699,25 @@ func NewDefaultPolicy(opts ...RouterOption) Policy {
...
+		return cfg, nil, fmt.Errorf("shard cost config %q: %w", shardCostSpec, costErr)
+	}
+
+	return cfg, costCfg, nil
+}
+
+// indexSlotCacheConfigFromRouter projects a parsed [routerConfig] into the
+// matching [indexSlotCacheConfig] used by [newIndexSlotCache].
+func indexSlotCacheConfigFromRouter(cfg routerConfig) indexSlotCacheConfig {
+	return indexSlotCacheConfig{
```
Jakob3xD, line 743: `ShardCostConfigError.Error()` (shard_cost_config.go:55) already returns a string prefixed with `"shard cost config: "`. Wrapping it here with `fmt.Errorf("shard cost config %q: %w", ...)` produces a doubled prefix:

```
shard cost config "r:base=-1": shard cost config: value must be a finite non-negative number for key "r:base"
```

The same doubled-prefix issue was fixed at the `New()` call site in `opensearchtransport.go` (now returns `newErr` directly), but this call in `NewDefaultPolicy` still wraps. Fix: use a non-overlapping wrapper that adds context without repeating the type prefix:
```go
return nil, fmt.Errorf("building default policy: %w", costErr)
```

opensearchtransport/shard_cost_config.go
```diff
@@ -0,0 +1,496 @@
...
+		}
+
+		key, valStr, hasEq := strings.Cut(item, "=")
+		if !hasEq {
+			return nil, &ShardCostConfigError{Key: key, Message: shardCostMessageMissing}
+		}
+
+		v, parseErr := strconv.ParseFloat(valStr, 64)
+		if parseErr != nil {
+			return nil, &ShardCostConfigError{Key: key, Message: shardCostMessageInvalid, Err: parseErr}
```
Jakob3xD, line 430: The bare-numeric path (line 403) correctly rejects `v <= 0`:
```go
if v <= 0 || math.IsInf(v, 0) || math.IsNaN(v) {
```
But this key=value guard uses `v < 0`, allowing `r:base=0` to pass. Since the doc comment at line 80 says bare form *sets r:base*, the two paths must validate identically.

`r:base=0` produces `dynamicCost = 0 + amplify*(writeUtil^exponent)`, which evaluates to `0.0` at idle — any primary-carrying node scores zero and is permanently preferred regardless of load, silently defeating the cost model.

Fix: change `v < 0` to `v <= 0` for the `r:base` case (line 430 condition), matching the bare-numeric guard. `r:amplify=0` (flat curve) and `r:exponent=0` may be intentional, so consider whether the zero guard should apply only to `r:base` or all three keys.

opensearchtransport/policy_misc_internal_test.go
```diff
@@ -77,3 +77,51 @@ func TestWithShardExactRouting(t *testing.T) {
...
+
+	tests := []struct {
+		name    string
+		initial string // pre-existing shardCostSpec value
+		spec    string // argument to WithShardCosts
+		want    string // expected shardCostSpec after applying
+	}{
+		{
+			name: "stores spec in config",
+			spec: "replica=1.0,write_primary=0.5",
```
Jakob3xD: `WithShardCosts` stores the spec string verbatim without parsing, so these test cases only validate field assignment — not parser correctness. The keys `preferred`, `alternate`, `r:replica`, and `w:primary` are not recognized by `parseShardCostConfig` (valid keys: `r:base`, `r:amplify`, `r:exponent`, `replica`, `write_primary`, `write_replica`, `unknown`, `relocating`, `initializing`) and would produce a `ShardCostConfigError` if passed to `NewDefaultRouter`.

The same issue applies to lines 102–103 and 107–115. Suggest replacing with valid specs so the tests also verify the parse round-trip:
```go
{name: "stores spec in config", spec: "r:base=0.9,r:amplify=2.5", want: "r:base=0.9,r:amplify=2.5"},
```

Jakob3xD (commented):
### Code review

Found 3 issue(s) — all posted as inline comments.

**Fixed since last review (20/22):** `Config.ShardCostConfig` applied in `New()`, error paths return defaults, NaN/Inf/negative validation (bare-numeric and key=value), env var error surfaced unconditionally, guide examples updated for two-return API, `releaseScoredConns` clears pointers before pool Put, `extraCost` applied after `scoreFunc`, env var standalone support, duplicate CHANGELOG entry, doc comment completeness, `DiscoverNodesOnStart` false-pointer bug, `psIsEnabled` table, `costWriteReplica` CHANGELOG entry.

opensearchtransport/router.go
sean-: Fixed in 824c6855. The wrap no longer doubles the prefix that `ShardCostConfigError.Error()` already produces:

```go
costCfg, costErr := parseShardCostConfig(shardCostSpec)
if costErr != nil {
    return cfg, nil, fmt.Errorf("parsing %q: %w", shardCostSpec, costErr)
}
```

Output is now single-prefixed:

```
parsing "r:base=-1": shard cost config: value must be a finite non-negative number for key "r:base"
```

Kept `%q` so the spec is still in the message for bare-numeric specs (where `ShardCostConfigError.Key` is empty and the spec value is the only context the user has). `errors.As(err, &scErr)` continues to recover the structured error.

opensearchtransport/shard_cost_config.go
sean-: Fixed in 824c6855. The `r:base` validator in the key=value path now matches the bare-numeric path's rejection set:

```go
case shardCostKeyBase, shardCostKeyAmplify, shardCostKeyExponent:
    // r:amplify=0 disables dynamic amplification (static-only mode).
    // r:exponent=0 holds dynamic cost at max regardless of utilization.
    // r:base must be strictly positive: r:base=0 makes primaries free at
    // idle, and the bare-numeric form (which sets r:base) rejects v<=0,
    // so the key=value form must match.
    if math.IsNaN(v) || math.IsInf(v, 0) || v < 0 || (key == shardCostKeyBase && v == 0) {
        return nil, &ShardCostConfigError{
            Key:     key,
            Message: shardCostMessageNotFinite,
        }
    }
```

Scoped the zero-rejection to `r:base` only, as you suggested. `r:amplify=0` (flat curve, static-only mode) and `r:exponent=0` are still accepted since they have well-defined semantics. Added an `r:base` zero rejected case to `TestParseShardCostConfig` to enforce the current behavior.

opensearchtransport/policy_misc_internal_test.go
sean-: Fixed in 824c6855. The keys in this test were already updated to valid forms in an earlier round, but you're right that the underlying gap remained. Upgraded the test to verify three things end-to-end:

```go
// 1. Field-assignment contract (original).
require.Equal(t, tt.want, cfg.shardCostSpec)

// 2. Each requested override lands in the right static-table slot;
//    slots not listed retain compile-time defaults.
parsed, err := parseShardCostConfig(cfg.shardCostSpec)
require.NoError(t, err)
for i := range wantReads {
    require.InDelta(t, wantReads[i], parsed.reads[i], epsilon, "reads[%d]", i)
}
for i := range wantWrites {
    require.InDelta(t, wantWrites[i], parsed.writes[i], epsilon, "writes[%d]", i)
}

// 3. Curve params are threaded into scoreFunc. At primaryPct=1.0 on an
//    idle connection, effectiveCost = base + amplify*(0^exponent) = base,
//    so score = rtt * base -- isolating curveBase as observable.
score := parsed.scoreFunc(conn, 1.0, 1.0, "", true)
require.InDelta(t, rtt*tt.wantBase, score, 0.001)
```

`r:base=0.9` now fails the test if any future regression drops the value between `WithShardCosts` and the dynamic-scoring closure. `InDelta` with `1e-9` covers the static slots (parser writes the IEEE-754 round-trip of the input float); `0.001` covers the score path where rtt-bucketing introduces its own quantization.

opensearchtransport/shard_cost_config.go
```diff
@@ -0,0 +1,496 @@
...
+			case shardCostKeyBase:
+				curveBase = v
+			case shardCostKeyAmplify:
+				curveAmplify = v
+			case shardCostKeyExponent:
+				curveExponent = v
+			}
+
+		// Static costs: shared (both tables).
+		case shardCostKeyUnknown:
```
Jakob3xD, line 456: Static cost keys (`unknown`, `relocating`, `initializing`, `replica`, `write_primary`, `write_replica`) have no NaN/`+Inf` guard before the table assignment. The dynamic curve keys (`r:base`, `r:amplify`, `r:exponent`) have an explicit `math.IsNaN(v) || math.IsInf(v, 0)` check at line 440 — the static path in the same function does not.

The post-parse clamp at lines 483–490 uses `cfg.reads[i] <= 0`, but `NaN <= 0` evaluates to `false` in IEEE 754, so NaN passes through unclamped. `+Inf <= 0` is also false. Result: `OPENSEARCH_GO_SHARD_COST=unknown=NaN` silently stores NaN in the cost table, and `rtt * utilization * NaN = NaN` corrupts all routing scores without error.

Fix: extend the clamp to cover non-finite values:
```go
for i := range cfg.reads {
    if cfg.reads[i] <= 0 || math.IsNaN(cfg.reads[i]) || math.IsInf(cfg.reads[i], 0) {
        cfg.reads[i] = shardCostForReads[i]
    }
}
```
Or add the same `math.IsNaN(v) || math.IsInf(v, 0)` guard before each static key assignment, consistent with the dynamic key path.

Jakob3xD (commented):
### Code review

Found 9 issue(s) — 1 as inline comment, 8 in body.

**1. `guides/routing.md` — `OPENSEARCH_GO_ROUTER` missing from Feature Environment Variables table**

The reference table at lines 1894–1900 lists `OPENSEARCH_GO_ROUTING_CONFIG`, `OPENSEARCH_GO_SHARD_REQUESTS`, `OPENSEARCH_GO_SHARD_COST`, `OPENSEARCH_GO_DISCOVERY_CONFIG`, and `OPENSEARCH_GO_FALLBACK`, but omits `OPENSEARCH_GO_ROUTER` — the primary opt-in env var for the DefaultRouter introduced by this PR. A user reading only the reference table will not find it.

https://github.com/opensearch-project/opensearch-go/blob/db224d80dd4925c739c28ad7e1e2b78ca29a8710/guides/routing.md#L1894-L1900

**2. `USER_GUIDE.md` — env var quick-reference table missing three routing variables**

Lines 504–515 list `OPENSEARCH_GO_POLICY_*`, `OPENSEARCH_GO_STANDBY_*`, etc., but are missing `OPENSEARCH_GO_ROUTER`, `OPENSEARCH_GO_SHARD_REQUESTS`, and `OPENSEARCH_GO_SHARD_COST` — all introduced or extended by this PR.

https://github.com/opensearch-project/opensearch-go/blob/db224d80dd4925c739c28ad7e1e2b78ca29a8710/USER_GUIDE.md#L504-L515

**3. Static cost key NaN/`+Inf` validation — see inline comment (`shard_cost_config.go:456`).**

**4. `policy_pool_router.go:46` — dead `poolRouter.decay` field.** `decay` is stored at construction but never read at scoring time; the actual decay runs inside `indexSlotCache`. Should be removed.

https://github.com/opensearch-project/opensearch-go/blob/db224d80dd4925c739c28ad7e1e2b78ca29a8710/opensearchtransport/policy_pool_router.go#L44-L48

**5. `policy_index_router.go:31–75` — dead `scoreSlicePool`.** `scoreSlicePool`, `acquireScoreSlice`, and `releaseScoreSlice` are defined with an `init()` and tests but have zero production call sites after `pool_buffers.go` was introduced. Should be removed.

https://github.com/opensearch-project/opensearch-go/blob/db224d80dd4925c739c28ad7e1e2b78ca29a8710/opensearchtransport/policy_index_router.go#L31-L75

**6. `testutil/debug.go:55` — hardcoded `"OPENSEARCH_GO_DEBUG"` string.** Should use `envvars.Debug` after the `ad21ce02` consolidation. The package is in the same module and the import is permitted.

https://github.com/opensearch-project/opensearch-go/blob/db224d80dd4925c739c28ad7e1e2b78ca29a8710/opensearchtransport/testutil/debug.go#L53-L58

**7. `testutil/debug.go:56` — `IsDebugEnabled` empty-string semantics diverge from `envvars.Truthy`.** `OPENSEARCH_GO_DEBUG=` (set but empty) returns `true` from `IsDebugEnabled` but `false` from `envvars.Truthy` — same env var, opposite behavior between the test helper and the runtime logger. Inconsistency introduced by this PR adding `envvars.Truthy` without unifying the test helper.

https://github.com/opensearch-project/opensearch-go/blob/db224d80dd4925c739c28ad7e1e2b78ca29a8710/opensearchtransport/testutil/debug.go#L53-L58

**8. `policy_override_env.go:42–45` — `psIsEnabled` docstring mentions `psEnvEnabled` despite the function never reading it.** The sentence was copied from the const-group comment and is misleading in the context of a function that ignores that bit.

https://github.com/opensearch-project/opensearch-go/blob/db224d80dd4925c739c28ad7e1e2b78ca29a8710/opensearchtransport/policy_override_env.go#L41-L51

**9. `WithDecayFactor` / `WithFanOutPerRequest` out-of-range values silently swallowed.** Invalid values are clamped with no `cfg.errs` entry and no log, inconsistent with `WithAdaptiveConcurrencyLimits` which does record a structured error.

https://github.com/opensearch-project/opensearch-go/blob/db224d80dd4925c739c28ad7e1e2b78ca29a8710/opensearchtransport/router.go#L564-L577

---

This is the final review pass. All 22 previously reported issues are confirmed fixed. Once the issues above are addressed, any new findings can be handled in a follow-up PR.

opensearchtransport/shard_cost_config.go
sean-: Fixed. Static-key parsing now rejects `NaN`/`+Inf` upfront with a `ShardCostConfigError` (`shardCostMessageNotFinite`), parallels the dynamic-curve validation handling. The post-parse `<= 0` clamp is left in place as a second line of defense, so an out-of-range static value can't silently propagate even if a future code path bypasses the per-key validation.

```go
case shardCostKeyUnknown:
    if math.IsNaN(v) || math.IsInf(v, 0) {
        return nil, &ShardCostConfigError{Key: key, Message: shardCostMessageNotFinite}
    }
    cfg.reads[shardCostUnknown] = v
    cfg.writes[shardCostUnknown] = v
```

(Same guard added to `relocating`, `initializing`, `replica`, `write_primary`, `write_replica`.)

sean-:
Thanks for the final review. All 9 items from [#pullrequestreview-4369853736](https://github.com/opensearch-project/opensearch-go/pull/815#pullrequestreview-4369853736) addressed in fixup commits.

**1. ✅ `guides/routing.md`** Feature Environment Variables table: added `OPENSEARCH_GO_ROUTER` row.

**2. ✅ `USER_GUIDE.md`** env var quick-reference table: added `OPENSEARCH_GO_ROUTER`, `OPENSEARCH_GO_SHARD_REQUESTS`, and `OPENSEARCH_GO_SHARD_COST` rows.

**3. ✅ Static cost key NaN/+Inf validation**: addressed inline at [discussion_r3308823592 (https://github.com/opensearch-project/opensearch-go/pull/815#discussion_r3308823592). Each static key now rejects non-finite values with a typed `ShardCostConfigError`; the post-parse `<= 0` clamp remains as defense in depth.

**4. ✅ Dead `poolRouter.decay` field**: removed. The decay value flows directly from `routerConfig.decay` -> `indexSlotCacheConfig.decayFactor` -> `slot.requestDecay.increment/decay` via the cache; the wrapper never needed its own copy. `wrapWithRouter`'s `decay` parameter dropped accordingly; all call sites and tests updated.

**5. ✅ Dead code** `scoreSlicePool` + `acquireScoreSlice` + `releaseScoreSlice` + their `init()`: removed. `pool_buffers.go`'s `acquireFloats(n)` is the live replacement; all production callers (`policy_index_router.go:199,237,296`, `policy_doc_router.go:179,235`) already use it. `TestConnScoreSelectLargePool` switched to `acquireFloats(...).Slice() / .Release()`; the pool-internals tests (`TestAcquireReleaseScoreSlice`, `TestScoreSlicePoolRatchetsUp`, `TestScoreSlicePoolDiscardsUndersized`, `BenchmarkAcquireReleaseScoreSlice`) deleted.

**6. ✅ `testutil/debug.go`** hardcoded `"OPENSEARCH_GO_DEBUG"` string: replaced with `envvars.Debug` constant.

**7.  ✅ `IsDebugEnabled`** empty-string semantics divergence from `envvars.Truthy`: unified. The implementation is now a one-liner: `return envvars.Truthy(envvars.Debug)`. The `OPENSEARCH_GO_DEBUG=` (empty) -> `true` quirk is gone; test helper and runtime logger now agree. `TestIsDebugEnabled` table updated: empty string returns `false` to match.

**8. ✅ `psIsEnabled` docstring**: rewritten. The function-level godoc now states what the function does and points readers to `parsePolicyOverrides` and the `policyState` const block for details. The actual bit-resolution logic is documented inline at the relevant branches inside the function body, removing the stale `psEnvEnabled` reference.

**9. ✅ `WithDecayFactor` / `WithFanOutPerRequest` silent clamps: both options now validate inputs and append to `routerConfig.errs` on out-of-range values, matching the `WithAdaptiveConcurrencyLimits` pattern. NaN, Inf, and out-of-range values are rejected:

```go
func WithDecayFactor(d float64) RouterOption {
    return func(c *routerConfig) {
        if math.IsNaN(d) || math.IsInf(d, 0) || d <= 0 || d >= 1 {
            c.errs = append(c.errs, fmt.Errorf(
                "WithDecayFactor(%v): must satisfy 0 < d < 1", d))
            return
        }
        c.decay = d
    }
}
```

The errors are logged at router build time via `loadDebugLogger()`, so an operator sees their bad config instead of having it silently overwritten by the `newIndexSlotCache` default-substitution clamp at `index_routing_cache.go:183-187`.

Jakob3xD (approved):

Issue #813, opened 2026-03-27
Call for New Maintainers — Help Wanted


Jakob3xD:
I want to use this issue to be fully transparent with the community about the current state of maintenance for this repository, and to formally call for new maintainers to step up. I believe openness here is more respectful to contributors and users than silently letting things stagnate.

---

### The Maintenance Situation

**We are critically low on active maintainers.**

At this point, the number of people with merge rights and active responsibility over this repository is very small and not all of them are reachable.

For an open-source library that serves as the primary Go client for OpenSearch, this is a serious problem. Users and organizations depending on opensearch-go deserve a project that is actively maintained, reviewed with care, and steered with a technical vision.

---

### The Volume and Complexity of Incoming Contributions

This repository has been seeing a significant number of substantial pull requests, in particular from @sean-, who has been contributing major features and architectural changes. I want to be clear: **these contributions are valuable and appreciated**.

However, reviewing large, complex PRs properly is not a trivial task. It requires:
- A thorough understanding of the codebase and its design decisions
- Time to test, reason about edge cases, and think about long-term implications
- The confidence to approve or request changes with a clear rationale

A PR like [#787](https://github.com/opensearch-project/opensearch-go/pull/787) is a good illustration of this challenge. It is a wide-ranging change that needs careful review.

---

### My Personal Limitations as a Maintainer

I want to be transparent about my constraints and why we need more maintainers.

**Time constraints.** This repository is a side task for me. It is not part of my main job or primary responsibilities. My capacity to review PRs, triage issues, and engage with contributors is limited.

**I am not a software developer by profession.** I do not have a clear technical vision for where this library should go architecturally and what guidelines we should use to enforce it. Specially when it comes to complex connection handling and multi process handling.

**My original motivation was pragmatic and narrow.** I came to this project because I needed a working Go client to manage OpenSearch clusters with my own internal tooling. In the longer term, I also have an interest in using it as a foundation for work around the [OpenSearch Kubernetes Operator](https://github.com/opensearch-project/opensearch-k8s-operator). These are legitimate use cases, but they are not a substitute for the broader vision and technical expertise this library needs in a maintainer.

**I am going on holiday.** Starting soon, I will be away for several weeks and effectively unavailable. I don't want to leave the project in a complete void without at least raising this issue first.

---

### What This Project Needs

I'd love to see one or more new maintainers step in who bring:

- **Technical depth in Go** and ideally experience with OpenSearch
[...]

PR #819, opened 2026-04-14
Skip shard routing tests on OpenSearch < 2.2.0 with security

sean-:
OpenSearch < 2.2.0 with the security plugin throws java.io.OptionalDataException on shard-routed requests due to non-thread-safe HashSet/HashMap in User serialization. Fixed in 2.2.0 by opensearch-project/security#1970 (50a94b47).

Widen the existing 2.1.0-only skip to cover all versions below 2.2.0.


rios0rios0 (approved):
Reviewed against issue #813 "quick wins" list. LGTM.

**What the PR does:** Adds `testutil.SkipIfVersion(t, "<", "2.2.0", ...)` guards behind `testutil.IsSecure(t)` to nine integration tests in `opensearchtransport/`. On OpenSearch <2.2.0 with the security plugin, a non-thread-safe `User` serialization (fixed upstream in opensearch-project/security#1970) produces `OptionalDataException` on inter-node traffic, which was causing intermittent false CI failures.

**Verification:**
- Checked out `pull/819/head`, `go build ./...` clean, `go test -race -short ./...` passes.
- `go test -tags=integration -run=NONE ./opensearchtransport/...` compiles cleanly (tag-gated tests typecheck).
- CI is green across the full compatibility matrix (1.3.20 through 3.6.0, both `true`/`false` security).

**Scope is narrow and surgical** — no production code changed, only test gating. Skip reason string is explicit, linked upstream security PR is referenced in every skip comment.

I'd suggest merging this **before #812**, since the doc PR's one failing `integ-test-compat (true, 2.1.0)` check is the exact race this guards.

Jakob3xD (approved):

PR #787, opened 2026-01-28
improve opensearchtransport

sean-:
improve `opensearchtransport/`

Fixes: #786 

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #820, opened 2026-04-15
Fix flaky health check retry

sean-:
## Summary

- **Fix flaky `TestDefaultHealthCheck_RetryAfterMaxRetry`**: Replace channel-based synchronization and wall-clock `maxRetryClusterHealth=50ms` with `ctx.Done()` signaling, eliminating races where httptest round-trip latency exceeds the retry interval. Apply the same fix to `TestDefaultHealthCheck_DisabledClusterHealthProbing` and `TestHealthCheckRequestModifier`. Introduced in #787.

- **Migrate `clusterHealthState` into `connLifecycle` bits**: Replace the separate `atomic.Int64` field on `Connection` with two lifecycle bits (`lcClusterHealthProbed`, `lcClusterHealthAvailable`) in the existing `connState` packed word. Delete `clusterHealthProbeState` type; add `hasClusterHealth()`, `clusterHealthPending()`, `clusterHealthUnavailable()` methods on `Connection`. Widen `conn.mu` critical sections to cover both data writes and lifecycle transitions atomically.

### DIFF
--- a/opensearchtransport/cluster_health.go
+++ b/opensearchtransport/cluster_health.go
@@ -447,7 +447,7 @@ func (c *Client) snapshotClusterHealthConnections() []*Connection {
 	p.mu.RLock()
 	snapshot := make([]*Connection, 0, len(p.mu.ready))
 	for _, conn := range p.mu.ready {
-		if conn.loadClusterHealthState().HasClusterHealth() {
+		if conn.hasClusterHealth() {
 			snapshot = append(snapshot, conn)
 		}
 	}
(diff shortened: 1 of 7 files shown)


opensearchtransport/cluster_health.go
```diff
@@ -573,7 +531,10 @@ func (c *Client) refreshClusterHealth(conn *Connection) {
 
 	switch {
 	case statusCode == http.StatusOK && health != nil:
+		conn.mu.Lock()
```
VijayanB, line 534: won't this add any performance bottlenecks under high concurrency?

sean-: The lock is acquired once every 30s-5min depending on the cluster size, holds for a pointer swap + CAS (nanoseconds), runs on a single background goroutine off the request path, and a single-node cluster skip this entirely.  Compared to everything else involved in a request, this is negligible.

VijayanB (approved):

PR #822, opened 2026-04-15
Add AddressResolver to rewrite discovered node URLs before pool entry

sean-:
When nodes are discovered via /_nodes/http, their publish_address may not be reachable from the client — e.g., containers exposing internal IPs, or environments using sidecar proxies on different ports. Without a hook, users must disable discovery or hand-manage the address list.

AddressResolverFunc is a per-node callback invoked during discovery that can rewrite, keep, or drop a node's URL. A four-case return protocol handles partial failures gracefully: (nil, error) drops the node, (nil, nil) keeps the default, (*url.URL, error) keeps the returned URL with a logged warning, and (*url.URL, nil) rewrites. Only when every resolver call fails does the client return ErrAllResolversFailed.

The built-in handler runs resolvers concurrently (limited by MaxAddressResolvers). For full control over concurrency, failure policy, or retry logic, AddressResolverRunnerFunc replaces the built-in handler entirely — it receives the node list and the per-node resolver (pre-instrumented with metrics) and returns the resolved set.

Three atomic counters (AddressResolverCalls, AddressResolverRewrites, AddressResolverErrors) and an OnAddressRewrite observer callback provide observability.


opensearchtransport/discovery.go
```diff
@@ -1073,6 +1103,248 @@ func (c *Client) getNodesInfo(ctx context.Context) ([]nodeInfo, error) {
...
+		if r.node.url != nil {
+			out = append(out, r.node)
+		}
+	}
+
+	if len(out) == 0 && len(errs) > 0 {
+		return nil, fmt.Errorf("%w: %w", ErrAllResolversFailed, errors.Join(errs...))
+	}
+
+	return out, nil
```
tender-barbarian, line 1366: In case there would be empty results and zero errors (e.g. context cancelled before any goroutine runs), function returns `(empty, nil)`. `nodeDiscovery` then builds an empty connection pool, wiping the existing one.

opensearchtransport/discovery.go
```diff
@@ -1073,6 +1103,248 @@ func (c *Client) getNodesInfo(ctx context.Context) ([]nodeInfo, error) {
...
+				if obs != nil {
+					obs.OnAddressRewrite(AddressRewriteEvent{
+						ID:           node.ID,
+						Name:         node.Name,
+						Roles:        node.Roles,
+						OriginalURL:  u.String(),
+						RewrittenURL: resolved.String(),
+						Timestamp:    time.Now().UTC(),
+					})
+				}
```
tender-barbarian: This repeated at ` runAddressResolverRunner()`. Might be worth a helper?

opensearchtransport/discovery.go
```diff
@@ -1073,6 +1103,248 @@ func (c *Client) getNodesInfo(ctx context.Context) ([]nodeInfo, error) {
...
+			}
+
+			info := NodeInfo{
+				ID:             p.node.ID,
+				Name:           p.node.Name,
+				Roles:          p.node.Roles,
+				Attributes:     p.node.Attributes,
+				PublishAddress: p.node.HTTP.PublishAddress,
+				URL:            p.defaultURL,
+			}
```
tender-barbarian, line 1315: `NodeInfo` doc says it's immutable but `Roles`, `Attributes` and `URL` are not copied. A resolver doing `node.URL.Path = "/x"` or `node.Attributes["k"] = v` will mutate the underlying `nodeInfo`.

```suggestion
			defaultURL := *p.defaultURL
			info := NodeInfo{
				ID:             p.node.ID,
				Name:           p.node.Name,
				Roles:          slices.Clone(p.node.Roles),
				Attributes:     maps.Clone(p.node.Attributes),
				PublishAddress: p.node.HTTP.PublishAddress,
				URL:            &defaultURL,
			}
```

I also suggest to add a test for this case.

opensearchtransport/discovery.go
```diff
@@ -1073,6 +1103,248 @@ func (c *Client) getNodesInfo(ctx context.Context) ([]nodeInfo, error) {
...
+	byID := make(map[string]discoveryPendingNode, len(pending))
+	for i, p := range pending {
+		nodeInfos[i] = NodeInfo{
+			ID:             p.node.ID,
+			Name:           p.node.Name,
+			Roles:          p.node.Roles,
+			Attributes:     p.node.Attributes,
+			PublishAddress: p.node.HTTP.PublishAddress,
+			URL:            p.defaultURL,
+		}
```
tender-barbarian, line 1458:
```suggestion
		defaultURL := *p.defaultURL
		nodeInfos[i] = NodeInfo{
			ID:             p.node.ID,
			Name:           p.node.Name,
			Roles:          slices.Clone(p.node.Roles),
			Attributes:     maps.Clone(p.node.Attributes),
			PublishAddress: p.node.HTTP.PublishAddress,
			URL:            &defaultURL,
		}

**Review by tender-barbarian — CHANGES_REQUESTED** [review]
I agree with overall approach, but the implementation is fairly complex so I probably missed something.

File: opensearchtransport/discovery.go (hunk 1 above)
Comment (sean-) [reply opensearchtransport/discovery.go:1366]: Good catch! Fixed by adding a guard after the error-collection loop: `if len(out) == 0 && len(pending) > 0`, we return `ctx.Err()` (or `ErrAllResolversFailed` if context is still live). Updated the test to assert the new behavior.

File: opensearchtransport/discovery.go (hunk 2 above)
Comment (sean-) [reply opensearchtransport/discovery.go]: Agreed. Extracted into `applyRewrite()` on `*Client`, used by both paths.

File: opensearchtransport/discovery.go (hunk 3 above)
Comment (sean-) [reply opensearchtransport/discovery.go:1315]: `NodeInfo` is documented as "a read-only snapshot" -- mutating its fields is undefined behavior. Adding `slices.Clone` or `maps.Clone` on every discovery cycle for every node is allocation overhead to guard against a contract violation. Because this is an "advanced" feature, I'd prefer to keep the contract and add a clarifying sentence to the doc if needed.

File: opensearchtransport/discovery.go (hunk 4 above)
Comment (sean-) [reply opensearchtransport/discovery.go:1458]: Same reasoning as above

File: opensearchtransport/discovery.go (hunk 3 above)
Comment (tender-barbarian) [reply opensearchtransport/discovery.go:1315]: Fair enough about alloc overhead. Can we fix the doc then? It says "mutating its fields has no effect on the client" but it does have an effect on the client, unless I misunderstood something ?

Comment (sean-) [reply opensearchtransport/discovery.go:1315]: Good catch, you are right: the old doc was misleading. I've updated the `NodeInfo` godoc to make this explicit:

> **WARNING:** NodeInfo shares memory with the client's internal node state. Resolvers MUST treat all fields as read-only. Mutating reference-type fields (Roles, Attributes, URL) corrupts internal state and the damage persists on the live Connection until the node is removed from the pool through rediscovery (identity change or cluster removal). The client does not defensively copy these fields for performance reasons.

If someone has a use-case where they need to inject temporary routing, roles, attributes, etc., that are used for evaluation, but need to reset, I could be persuaded to come up with a more nuanced way to inject a user-modified version that gets reset every request.

**Review by tender-barbarian — APPROVED** [review]

File: opensearchtransport/discovery.go [hunk 5]
```diff
@@ -1204,6 +1234,237 @@ func (c *Client) getNodesInfo(ctx context.Context) ([]nodeInfo, error) {
...
+		if r.node.url != nil {
+			out = append(out, r.node)
+		}
+	}
+
+	if len(out) == 0 && len(errs) > 0 {
+		return nil, fmt.Errorf("%w: %w", ErrAllResolversFailed, errors.Join(errs...))
+	}
+
+	if len(out) == 0 && len(pending) > 0 {
```
Jakob3xD, line 1359: The `cf60b842` guard covers the all-empty cancel case, but the **partial** cancel case slips through. On context cancellation mid-fan-out, `sem.Acquire` fails and the launch loop `break`s, leaving the undispatched nodes as zero-value `results[i]` (`url == nil`, `err == nil`). They are silently skipped in the collection loop. If at least one node already resolved, `out` is non-empty, so neither this guard (`len(out) == 0`) nor the all-failed guard fires, and the function returns `(partialList, nil)`.

`doDiscoverNodes` has no `ctx.Err()` check after `getNodesInfo`, so that partial list flows into `updateConnectionPool`, which treats the dropped nodes as departed from the cluster and evicts them (`router.DiscoveryUpdate` / `OnDiscoveryRemove`). A caller invoking the public `DiscoverNodes(ctx)` with a deadline that fires mid-resolution can thus silently evict healthy nodes. (The dispatched-but-cancelled goroutines also return `ctx.Err()` into `errs`, which is then swallowed by `return out, nil`.)

Suggested fix: prefer a single `if ctx.Err() != nil { return nil, ctx.Err() }` after `getNodesInfo` in `doDiscoverNodes` — it covers both resolver paths and any future one. Alternatively, extend this guard to fire whenever `ctx.Err() != nil` regardless of `len(out)`. `runAddressResolverRunner` has the same gap.

Jakob3xD (commented):
### Code review

Reviewed the AddressResolver feature on head `d2e08c4b`. The design is solid and the prior review thread is addressed — the empty-pool-on-full-cancel guard (`cf60b842`), the `applyRewrite` de-duplication, and the documented `NodeInfo` read-only aliasing tradeoff all check out, and the default (no-resolver) discovery path is byte-identical to before. Not re-raising those.

One issue is posted as an inline comment (partial context-cancel can silently evict healthy nodes). Minor notes below.

#### Minor

1. **Runner receives a nil `resolve`** when `AddressResolverRunner` is set but `AddressResolver` is not — `newInstrumentedResolver` returns nil and that nil is passed to the runner. A test asserts this, so it's intended, but `AddressResolverRunnerFunc`'s godoc doesn't warn that `resolve` may be nil, so a runner that calls it unguarded panics. Worth a one-line godoc note (or pass a no-op resolver).

https://github.com/opensearch-project/opensearch-go/blob/d2e08c4b5c238a71a9a196b933b3974a073571b9/opensearchtransport/discovery.go#L1434-L1438

2. **No validation of resolver-returned URLs.** A resolver returning a non-nil but malformed `*url.URL` (empty `Host`/`Scheme`, e.g. `&url.URL{}`) flows through `applyRewrite` into the connection pool and surfaces as confusing HTTP errors later rather than a resolver-time diagnostic. Consider rejecting non-nil URLs with empty `Scheme`/`Host`, or documenting that callers must return fully-formed URLs.

https://github.com/opensearch-project/opensearch-go/blob/d2e08c4b5c238a71a9a196b933b3974a073571b9/opensearchtransport/discovery.go#L1362-L1385

3. **Runner path doesn't dedupe by node ID.** `runAddressResolverRunner` appends every `ResolvedAddress` with a non-nil URL; if a runner returns the same `Node.ID` twice, two `*Connection` for one node enter the pool (the built-in path is immune via positional `results[i]`). Dedupe by ID, or document "each node at most once".

https://github.com/opensearch-project/opensearch-go/blob/d2e08c4b5c238a71a9a196b933b3974a073571b9/opensearchtransport/discovery.go#L1446-L1459

4. **Doc nits.** `opensearch.go` `MaxAddressResolvers` says `1 = serial (no goroutines spawned)`, but goroutines are still spawned — they're just serialized by a weight-1 semaphore. And the `ErrAllResolversFailed` var doc scopes it to the built-in `(nil, error)` case, while `runAddressResolverRunner` also returns it on an empty result set.

https://github.com/opensearch-project/opensearch-go/blob/d2e08c4b5c238a71a9a196b933b3974a073571b9/opensearch.go#L236-L243

5. **Consistency:** `AddressRewriteEvent.Roles` is assigned `node.Roles` directly, whereas `ConnectionEvent` copies via `Roles.toSlice()`. Given the accepted read-only-aliasing tradeoff for `NodeInfo` this may be deliberate, but an observer retaining the event sees a live slice; a copy would match the other event types.

https://github.com/opensearch-project/opensearch-go/blob/d2e08c4b5c238a71a9a196b933b3974a073571b9/opensearchtransport/discovery.go#L1373-L1380

6. `ConnectionObserver` gains a 15th method (`OnAddressRewrite`); `BaseConnectionObserver` provides a no-op so embedders are safe, but direct implementers break at compile time. Worth a CHANGELOG breaking-change note for the minor release.

opensearchtransport/discovery.go
sean-: Good catch -- you're right that the partial-cancel case slips through. Took your suggested approach: a single `ctx.Err()` check in `doDiscoverNodes` right after `getNodesInfo`. Bails out before the partial node list reaches `updateConnectionPool`, covers both the built-in handler and the `AddressResolverRunner` path, and any future resolution path picks it up for free.

New table-driven test `TestDiscoverNodes_PartialCancelDoesNotEvict` exercises the three relevant shapes -- pre-cancelled context through the built-in handler, mid-flight cancel through the built-in handler (forces a partial `results[i]`), and runner-emitted partial result -- each seeds a populated pool first and asserts the cancelled cycle leaves it intact.

sean-:
Thanks for the review, @Jakob3xD. much appreciated. All six items addressed.

A few notes on what changed in code vs. docs:

- **Partial context-cancel eviction**: single `ctx.Err()` guard in `doDiscoverNodes` right after `getNodesInfo`, per your suggested approach. Covers both resolver paths and any future one. Test details in the inline reply.
- **`AddressResolverRunnerFunc` nil `resolve`**: godoc now states explicitly that `resolve` is nil when `AddressResolverRunner` is configured without a paired `AddressResolver`, and that runners delegating to it must guard the nil case.
- **URL validation**: a non-nil `*url.URL` with empty `Scheme` or `Host` is now treated as a resolver error -- default URL kept, `AddressResolverErrors` incremented, debug log emitted. Documented in `AddressResolverFunc`'s return-semantics block. Avoids letting `&url.URL{}` reach the connection pool and surface later as a confusing HTTP error. New row in the protocol table-driven test.
- **Dedupe in `runAddressResolverRunner`**: tracks a `seen` set keyed by `Node.ID`; first non-nil URL wins, subsequent entries are dropped with a debug log. Documented in `AddressResolverRunnerFunc`'s godoc. New table-driven sub-test `runner_duplicate_node_IDs_are_deduped` covers four shapes: exact duplicate per node, duplicate with a different URL (verifies first-wins), triplicate collapsing to one, and a nil-URL-then-duplicates sequence (verifies nil-drop and dedupe stack correctly).
- **Doc nits**: `MaxAddressResolvers=1` is now described as "serialized (weight-1 semaphore)" rather than "no goroutines spawned", which was misleading. `ErrAllResolversFailed` doc now names both return paths (built-in `(nil, error)` and runner empty-result).
- **Roles consistency**: `AddressRewriteEvent.Roles` now uses `slices.Clone(node.Roles)` to match `ConnectionEvent`'s `c.Roles.toSlice()`. The `NodeInfo` aliasing tradeoff stays as documented for the resolver-callback surface; the observer-event surface is now consistent across all event types.

Jakob3xD (approved):
Re-reviewed on a0d7f60e. All seven items from the previous review are addressed, and I verified the substantive fixes in the code rather than just the commit messages:

- Partial context-cancel eviction (H1): the new `ctx.Err()` guard in `doDiscoverNodes` right after `getNodesInfo`, before any `updateConnectionPool` call, correctly covers the partial-cancel case for both the built-in handler and the `AddressResolverRunner` path. This is the clean single-site fix.
- URL validation: a non-nil `*url.URL` with empty Scheme/Host is now treated as a resolver error in `applyRewrite` (default URL kept, no malformed URL reaches the pool).
- Runner dedupe: first-non-nil-URL-per-ID wins, foreign IDs not in the discovered set are skipped.
- `AddressRewriteEvent.Roles` now uses `slices.Clone`, matching the other event types.
- nil-`resolve` godoc warning, the `MaxAddressResolvers`/`ErrAllResolversFailed` doc nits, and the `ConnectionObserver` BREAKING changelog note are all in.

LGTM.

One nice-to-have, non-blocking: when a resolver returns both a malformed URL and a non-nil error, `addressResolverErrors` is incremented twice -- once at the resolver call site (`err != nil`) and once in `applyRewrite` (empty Scheme/Host). A doubly-degenerate case; only the metric is affected, behavior is correct. Could skip the `applyRewrite` increment when the resolver already reported an error, or leave it as-is.

PR #823, opened 2026-04-16
Include _nodes.failures in discovery error messages

sean-:
The discovery guard added in b8ba2e29 rejects `/_nodes/http` responses where `successful=0`, but the error message only reported the counts (`total`, `successful`, `failed`) without the server's failure reasons.

CI intermittently fails on OpenSearch 2.0.1 (secure) with all three nodes reported as failed during the `/_nodes` internal transport fan-out. We suspect this is the `OptionalDataException` race in the security plugin's User serialization (opensearch-project/security#1970, fixed in 2.2.0), but cannot confirm without seeing the actual failure details.

Surface the `_nodes.failures` JSON array in both the returned error and the debug log so the next CI failure reveals the root cause.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -111,6 +111,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ### Changed
 
+- Include `_nodes.failures` detail in discovery error messages for diagnosing intermittent CI failures on older OpenSearch versions ([#823](https://github.com/opensearch-project/opensearch-go/pull/823))
 - Test against Opensearch 3.6.0 ([#817](https://github.com/opensearch-project/opensearch-go/pull/817))
 - Consolidate test utilities into two canonical packages: opensearchtransport/testutil (env helpers, polling, version comparison) and opensearchapi/testutil (client-dependent helpers, test suite, JSON comparison)
 - Rename `singleConnectionPool` to `singleServerPool` and `statusConnectionPool` to `multiServerPool` for clarity ([#786](https://github.com/opensearch-project/opensearch-go/pull/786))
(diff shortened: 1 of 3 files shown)


sean-:
This is needed to figure out why https://github.com/opensearch-project/opensearch-go/actions/runs/24528045333/job/71703532786 failed.  I suspect we just need to more broadly apply a `SkipIfVersion()` guard for more tests.

rios0rios0 (approved):
Reviewed against issue #813 "quick wins" list. LGTM.

**What the PR does:** Promotes the `_nodes.failures` raw JSON from debug-only logging into the error returned by `getNodesInfo` when `successful == 0`. Adds `(*_NodesMeta).formatFailures()` that marshals the `[]json.RawMessage` to a compact JSON string (returns `"[]"` when empty; returns a descriptive placeholder on marshal error — no panic).

**Verification:**
- Built `pull/823/head` clean, `go test -race` clean.
- New `TestNodesMeta_formatFailures` covers nil, empty, single, multiple, and malformed-JSON cases — all pass.
- Existing `TestGetNodesInfoNodesMeta/all_nodes_failed` was tightened to assert the new `failures=[...]` substring.
- `errors.Is(err, errDiscoveryEmpty)` is preserved — wrapping order unchanged, existing callers that key off that sentinel continue to work.

**Small observation** (non-blocking): if a real cluster returns a lot of large failure objects, the full JSON will land in the error message. For the stated purpose (diagnosing intermittent CI failures) that's the whole point, and Go's `%w`-wrapped errors are lazy-stringified, so no runtime cost unless something actually calls `.Error()`. No concern from me.

Jakob3xD (approved):

PR #824, opened 2026-04-23
Fix bulk indexer silently creating duplicate documents for IDs with <, >, &

sean-:
Go's json.Marshal HTML-escapes `<`, `>`, and `&` to `\u003c`, `\u003e`, `\u0026` by default. OpenSearch's bulk API stores the raw `_id` value without un-escaping, so a document ID like `prefix|<root_account>|suffix` was persisted as `prefix|\u003croot_account\u003e|suffix`. This causes:

1. Duplicate documents when upserting against pre-existing docs that were indexed with the literal `_id`.
2. Documents unreachable by their expected `_id` on read-by-ID paths.
3. Downstream services receiving escaped `_id` values from search results and aggregations.

Present since commit `3da59092` (`2021-08-25`) across `v1`, `v2`, `v3`, and `v4`.

Replace `json.Marshal` with `json.NewEncoder` + `SetEscapeHTML(false)` in `writeMeta` and `JSONReader`. Replace per-worker `aux []byte` with a `sync.Pool`-backed `*bytes.Buffer` (configurable cap via `MetaBufferPoolMaxBytes`, default `32 KiB`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -145,6 +145,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ### Fixed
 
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -145,6 +146,8 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ### Fixed
 
+- Fix bulk indexer HTML-escaping `_id` and `routing` values containing `<`, `>`, or `&` characters, causing OpenSearch to store escaped values (e.g., `<root_account>` instead of `<root_account>`), leading to duplicate documents, unreachable data on read-by-ID paths, and potential shard routing mismatches. Present since the `json.Marshal` migration in 2021 (commit `3da59092`). Replace `json.Marshal` with `json.NewEncoder` + `SetEscapeHTML(false)` in `opensearchutil.worker.writeMeta`
```
VijayanB: i am confused, why we have two entries? can this be just one entry and add link to PR?

sean-: Yes. Give me about 15 minutes and I'll push something up when I get back online.

sean-: Updated

VijayanB (approved):

PR #830, opened 2026-04-30
Cancel old pool context on multi-to-single demotion

sean-:
When discovery temporarily sees a single node (rolling restart, network partition), demoteConnectionPoolWithLock replaces the multiServerPool with a singleServerPool. Resurrection goroutines from the old pool capture the pool pointer and loop on its context. Because the pool previously shared the Client's context directly, those goroutines never exit until the entire client is closed.

Give each multiServerPool its own derived context. Cancel it on demotion so orphaned resurrection goroutines exit via <-ctx.Done().

Ref: opensearch-project/opensearch-go#811


VijayanB (dismissed):

opensearchtransport/opensearchtransport.go
```diff
@@ -2315,6 +2317,11 @@ func (c *Client) promoteConnectionPoolWithLock(readyConnections, deadConnections
 func (c *Client) demoteConnectionPoolWithLock() *singleServerPool {
 	switch currentPool := c.mu.connectionPool.(type) {
 	case *multiServerPool:
+		// Cancel the old pool's context to clean up stale background goroutines.
+		if currentPool.cancel != nil {
```
Jakob3xD, line 2341: This != nil guard exists because the multiServerPool built inline in New() (~line 886) stores ctx: ctx but never sets cancel. If that initial pool is demoted (e.g. discovery temporarily sees one node after a rolling restart), this check short-circuits and the pool's resurrection goroutines keep looping until the entire client is closed — the original bug.

The fix is to store cancel on the initial pool too. The cleanest path is to call `newMultiServerPoolFromClientWithLock` from `New()` instead of the inline literal, since that function already does the right thing. At minimum the literal should carry the cancel already in scope:
```
pool := &multiServerPool{
     name:   "client",
     ctx:    ctx,
     cancel: cancel, // cancel is already in scope from the context.WithCancel on line 743
     ...
 }
```
Once that gap is plugged the != nil guard becomes dead code (context.WithCancel always returns a non-nil func), though keeping it as a defensive safety net is harmless.

opensearchtransport/opensearchtransport_internal_test.go
```diff
@@ -1450,6 +1450,40 @@ func TestConnectionPoolPromotion(t *testing.T) {
 		require.Nil(t, singlePool.connection, "Should handle empty pool gracefully")
 	})
```
Jakob3xD, line 1473: This sub-test only exercises the `newMultiServerPoolFromClientWithLock` path. The initial pool created by `New()` is a separate code path and should have its own sub-test. The following currently fails, confirming the gap:
```
t.Run("demoteConnectionPoolWithLock cancels initial New() pool context", func(t *testing.T) {
    t.Parallel()
    u1, _ := url.Parse("http://node1:9200")
    u2, _ := url.Parse("http://node2:9200")
    client, err := New(Config{URLs: []*url.URL{u1, u2}})
    require.NoError(t, err)

    client.mu.RLock()
    pool, ok := client.mu.connectionPool.(*multiServerPool)
    client.mu.RUnlock()
    require.True(t, ok, "expected initial multiServerPool")

    poolCtx := pool.ctx

    select {
    case <-poolCtx.Done():
        t.Fatal("pool context should not be cancelled before demotion")
    default:
    }

    client.mu.Lock()
    client.demoteConnectionPoolWithLock()
    client.mu.Unlock()

    select {
    case <-poolCtx.Done():
    default:
        t.Fatal("pool context should be cancelled after demotion")
    }
})
```

Jakob3xD (commented):
For transparency.
As I am not familiar with context and pooling, the review was created with the help of Claud.

opensearchtransport/opensearchtransport.go
sean-: Good catch: you're right that the inline pool literal in `New()` was storing `ctx` without a corresponding `cancel`, making this guard a silent no-op for the initial pool.

Fixed by giving each inline `multiServerPool` its own derived child context:

```go
poolCtx, poolCancel := context.WithCancel(ctx)
pool := &multiServerPool{
    name:   "client",
    ctx:    poolCtx,
    cancel: poolCancel,
    ...
}
```

This matches what `newMultiServerPoolFromClientWithLock()` already does: demotion cancels only the pool's context without touching the client's root context, so `Close()` cascade still works correctly.

The same gap existed on the seed-fallback pool; fixed both. Kept the `!= nil` guard as a defensive safety net.

opensearchtransport/opensearchtransport_internal_test.go
sean-: Added. The test you provided now passes and included it alongside the existing `newMultiServerPoolFromClientWithLock` test so both code paths have explicit coverage.

CHANGELOG.md
```diff
@@ -154,16 +153,14 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
...
+- Fix bulk indexer HTML-escaping `_id` and `routing` values containing `<`, `>`, or `&` characters, causing OpenSearch to store escaped values (e.g., `\u003croot_account\u003e` stored instead of `<root_account>`), leading to duplicate documents, unreachable data on read-by-ID paths, and potential shard routing mismatches. Present since the `json.Marshal` migration in 2021 (commit `3da59092`). Replace `json.Marshal` with `json.NewEncoder` + `SetEscapeHTML(false)` in `opensearchutil.worker.writeMeta` and `opensearchutil.JSONReader`; replace per-worker `aux []byte` with `sync.Pool`-backed `*bytes.Buffer`; add table-driven test coverage for `writeMeta` edge cases ([#824](https://github.com/opensearch-project/opensearch-go/pull/824))
 - Fix pool replacement orphaning resurrection goroutines during node discovery, causing connections to become permanently dead with no active health checker ([#786](https://github.com/opensearch-project/opensearch-go/pull/786))
+- Fix multi-to-single pool demotion leaking resurrection goroutines by giving each `multiServerPool` its own derived context and cancelling it on demotion ([#811](https://github.com/opensearch-project/opensearch-go/issues/811))
```
```suggestion
- Fix multi-to-single pool demotion leaking resurrection goroutines by giving each `multiServerPool` its own derived context and cancelling it on demotion ([#830](https://github.com/opensearch-project/opensearch-go/pull/830))
```

sean-: Done. Updated to `#830` PR link and rebased on latest `main`.

opensearchtransport/pool_multi_server.go
```diff
@@ -29,7 +29,8 @@ type multiServerPool struct {
 	// resurrection, RTT probes). Derived from Client.ctx during pool construction
 	// so that Client.Close() cancels all background goroutines.
 	//nolint:containedctx // Long-lived context required for background worker lifecycle
-	ctx context.Context
+	ctx    context.Context
+	cancel context.CancelFunc
```
Jakob3xD, line 35: The comment above (lines 28–30) now only describes the transitive cancellation path via `Client.Close()`, but the PR adds a second mechanism: explicit per-pool cancellation via `demoteConnectionPoolWithLock()`. The description is now incomplete and could mislead future readers into thinking `Close()` is the only cleanup path.

Suggest adding a note, e.g.:
```
// ctx is the lifecycle context for async pool operations (health checks,
// resurrection, RTT probes). Derived from Client.ctx during pool construction
// so that Client.Close() cancels all background goroutines. Each pool also
// holds a cancel func for early cleanup on demotion (multi→single); see
// demoteConnectionPoolWithLock().
```

opensearchtransport/opensearchtransport.go
```diff
@@ -841,9 +841,11 @@ func New(cfg Config) (*Client, error) {
 			seedConns[i] = conn
 		}
 
+		seedCtx, seedCancel := context.WithCancel(ctx)
```
Jakob3xD: The `seedFallbackPool` has no demotion path — it lives in `c.seedFallbackPool` and is never assigned to `c.mu.connectionPool`. Since `demoteConnectionPoolWithLock()` only operates on `c.mu.connectionPool`, `seedCancel` is never called explicitly. The seed pool's goroutines are cleaned up transitively when `c.cancelFunc()` in `Close()` cancels the root context.

The fix correctly removed the original bug (no `cancel` stored at all on the initial pools), but the `context.WithCancel` here creates an unused cancel func. Using `ctx` directly for the seed pool would avoid the allocation and make the lifecycle asymmetry explicit:
```go
seedPool := &multiServerPool{
    name: "seed-fallback",
    ctx:  ctx, // cleaned up transitively via c.cancelFunc() in Close()
    ...
}
```

Jakob3xD (commented):
### Code review

Found 2 issue(s) — all posted as inline comments.

opensearchtransport/pool_multi_server.go
sean-: Thank you.  The doc was stale after the PR added the second cancellation path. Updated the comment block to mention both:

```go
// ctx is the lifecycle context for async pool operations (health checks,
// resurrection, RTT probes). Derived from Client.ctx during pool construction
// so that Client.Close() cancels all background goroutines. cancel is the
// pool's own cancel func for early cleanup on multi-to-single demotion; see
// [Client.demoteConnectionPoolWithLock].
```

opensearchtransport/opensearchtransport.go
sean-: Good call.  `seedFallbackPool` is held in `c.seedFallbackPool` and is never assigned to `c.mu.connectionPool`, so `demoteConnectionPoolWithLock()` cannot reach it and `seedCancel` is
unreachable. Cleanup occurs transitively via `c.cancelFunc()` in `Close()`.

Updated to pass `ctx` directly and removed the unused `context.WithCancel`. Added a comment noting the lifecycle asymmetry:

```go
// seed-fallback has no demotion path (it lives in c.seedFallbackPool, not
// c.mu.connectionPool), so it does not need its own cancel func; its
// goroutines are cleaned up transitively when Client.Close() cancels the
// root context.
seedPool := &multiServerPool{
    name: "seed-fallback",
    ctx:  ctx,
    ...
}
```

The main `client` pool retains its own `cancel` because it is subject to demotion.

Jakob3xD (approved):

PR #833, opened 2026-05-16
ci: Add Docker image pre-pull with retry for resilience

ssunno:
## Why

CI may fail when Docker Hub is unstable. Add a setting to retry for up to 5 minutes instead of failing immediately. 
> this is also a problem for other PRs, it is worth proceeding with a quick merge instead of waiting for a large PR.

This was also covered in #804

## What changed

- apply pre-pull
  `opensearchproject/opensearch:<matrix-version>` with a 30-attempt × 10s
  retry loop before `make cluster.build`

### DIFF
--- a/.github/workflows/test-compatibility.yml
+++ b/.github/workflows/test-compatibility.yml
@@ -55,6 +55,15 @@ jobs:
           sudo sysctl -w fs.file-max=262144
           sudo sysctl -w vm.max_map_count=262144
 
+      - name: Pre-pull OpenSearch image
+        run: |
+          echo "Pre-pulling opensearchproject/opensearch:${{ matrix.entry.opensearch_version }}..."
+          for attempt in $(seq 30); do
+            docker pull opensearchproject/opensearch:${{ matrix.entry.opensearch_version }} && break
+            echo "Pull attempt ${attempt}/30 failed, retrying in 10s..."
+            sleep 10
+          done
+
       - name: Launch OpenSearch cluster
         run: |
           export OPENSEARCH_VERSION=${{ matrix.entry.opensearch_version }}
--- a/.github/workflows/test-compatibility.yml
+++ b/.github/workflows/test-compatibility.yml
@@ -55,15 +55,6 @@ jobs:
           sudo sysctl -w fs.file-max=262144
           sudo sysctl -w vm.max_map_count=262144
 
-      - name: Pre-pull OpenSearch image
-        run: |
-          echo "Pre-pulling opensearchproject/opensearch:${{ matrix.entry.opensearch_version }}..."
-          for attempt in $(seq 30); do
-            docker pull opensearchproject/opensearch:${{ matrix.entry.opensearch_version }} && break
-            echo "Pull attempt ${attempt}/30 failed, retrying in 10s..."
-            sleep 10
-          done
-
       - name: Launch OpenSearch cluster
         run: |
           export OPENSEARCH_VERSION=${{ matrix.entry.opensearch_version }}
--- a/Makefile
+++ b/Makefile
@@ -336,6 +336,12 @@ cluster.docker-build:
 		fi \
 	))
 	@echo "Building OpenSearch $(OPENSEARCH_VERSION) with role: $(manager_role), secure: $(SECURE_INTEGRATION)"
+	@echo "Pre-pulling base image opensearchproject/opensearch:$(OPENSEARCH_VERSION)..."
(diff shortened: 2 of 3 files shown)


.github/workflows/test-compatibility.yml
```diff
@@ -55,6 +55,15 @@ jobs:
...
 
+      - name: Pre-pull OpenSearch image
+        run: |
+          echo "Pre-pulling opensearchproject/opensearch:${{ matrix.entry.opensearch_version }}..."
+          for attempt in $(seq 30); do
+            docker pull opensearchproject/opensearch:${{ matrix.entry.opensearch_version }} && break
+            echo "Pull attempt ${attempt}/30 failed, retrying in 10s..."
+            sleep 10
+          done
+
```
Jakob3xD: This task is not needed if it is already handled in the Makefile. It would be redundant.
```suggestion
```

Makefile
```diff
@@ -336,6 +336,12 @@ cluster.docker-build:
 		fi \
 	))
 	@echo "Building OpenSearch $(OPENSEARCH_VERSION) with role: $(manager_role), secure: $(SECURE_INTEGRATION)"
+	@echo "Pre-pulling base image opensearchproject/opensearch:$(OPENSEARCH_VERSION)..."
+	@for attempt in $$(seq 30); do \
+		docker pull opensearchproject/opensearch:$(OPENSEARCH_VERSION) && break; \
+		echo "Pull attempt $$attempt/30 failed, retrying in 10s..."; \
+		sleep 10; \
+	done
```
Jakob3xD, line 345: 1. Add the variable near the existing tool-version block at the top of the Makefile:

Tool versions:
```
GOLANGCI_LINT_VERSION := v2.11.4
DOCKER_PULL_RETRIES   := 30
```
2. Fix the loop
The for loop exits with the status of its last command. When all 30 pulls fail, the last command executed is sleep 10 (exit 0).
The exit 1 check fires on the last attempt before sleep 10, so there is no wasted 10-second wait on the final failure.
```suggestion
	@echo "Pre-pulling base image opensearchproject/opensearch:$(OPENSEARCH_VERSION)..."
	@for attempt in $$(seq $(DOCKER_PULL_RETRIES)); do \
		docker pull opensearchproject/opensearch:$(OPENSEARCH_VERSION) && break; \
		echo "Pull attempt $$attempt/$(DOCKER_PULL_RETRIES) failed, retrying in 10s..."; \
		[ "$$attempt" -eq $(DOCKER_PULL_RETRIES) ] && { echo "All $(DOCKER_PULL_RETRIES) pull attempts failed, aborting."; exit 1; }; \
		sleep 10; \
	done
```

Jakob3xD (changes requested):

Makefile
```diff
@@ -336,6 +336,13 @@ cluster.docker-build:
 		fi \
 	))
 	@echo "Building OpenSearch $(OPENSEARCH_VERSION) with role: $(manager_role), secure: $(SECURE_INTEGRATION)"
+	@echo "Pre-pulling base image opensearchproject/opensearch:$(OPENSEARCH_VERSION)..."
+	@for attempt in $$(seq $(DOCKER_PULL_RETRIES)); do \
```
Jakob3xD, line 340: `DOCKER_PULL_RETRIES` is not defined anywhere in the Makefile. When Make expands `$(DOCKER_PULL_RETRIES)` and the variable is unset, `seq` receives no argument → `seq: missing operand` → CI breaks immediately. This is confirmed by the current CI output.

The Makefile already uses `$(eval VAR ?= value)` inside targets (e.g. `$(eval OPENSEARCH_VERSION ?= latest)` a few lines above). Following that pattern:

```makefile
$(eval DOCKER_PULL_RETRIES ?= 30)
```

Add this before the `@echo "Pre-pulling..."` line so it follows the same convention as the other configurable defaults in this target.

Jakob3xD (commented):
### Code review

Found 1 issue — posted as inline comment.

Makefile
sean-: @ssunno @Jakob3xD, FWIW, the review feedback from this PR has been folded back into #804:

- The silent-failure retry loop is fixed in `cluster.docker-build` (guard before the final `sleep 10`).
- Redundant workflow-level pre-pull step dropped (the workflow's `Launch OpenSearch cluster` step already invokes `make cluster.build`).

Reference: opensearch-project/opensearch-go@259f7262 (`Fixes: #833`).

PR #825, opened 2026-04-24
chore: Updated Go version to 1.25.9

ramiyengar:
### Description
Updates Go to a known new version. 

### Issues Resolved
Known CVEs have been resolved by updating the version of Go.  

I ran a tool named `osv-scanner` which allows a repo to be scanned for known CVEs. When I ran this on the current repo, it identified 33 known vulnerabilities which could be fixed by updating the version of Go. Details:
```
Total 1 package affected by 33 known vulnerabilities (0 Critical, 0 High, 0 Medium, 0 Low, 33 Unknown) from 1 ecosystem.
33 vulnerabilities can be fixed.

╭──────────────────────────────┬──────┬───────────┬─────────┬─────────┬───────────────┬──────────────────────╮
│ OSV URL                      │ CVSS │ ECOSYSTEM │ PACKAGE │ VERSION │ FIXED VERSION │ SOURCE               │
├──────────────────────────────┼──────┼───────────┼─────────┼─────────┼───────────────┼──────────────────────┤
│ https://osv.dev/GO-2025-3503 │      │ Go        │ stdlib  │ 1.24.0  │ 1.24.1        │ opensearch-go/go.mod │
│ https://osv.dev/GO-2025-3563 │      │ Go        │ stdlib  │ 1.24.0  │ 1.24.2        │ opensearch-go/go.mod │
│ https://osv.dev/GO-2025-3749 │      │ Go        │ stdlib  │ 1.24.0  │ 1.24.4        │ opensearch-go/go.mod │
│ https://osv.dev/GO-2025-3750 │      │ Go        │ stdlib  │ 1.24.0  │ 1.24.4        │ opensearch-go/go.mod │
│ https://osv.dev/GO-2025-3751 │      │ Go        │ stdlib  │ 1.24.0  │ 1.24.4        │ opensearch-go/go.mod │
│ https://osv.dev/GO-2025-3849 │      │ Go        │ stdlib  │ 1.24.0  │ 1.24.6        │ opensearch-go/go.mod │
│ https://osv.dev/GO-2025-3956 │      │ Go        │ stdlib  │ 1.24.0  │ 1.24.6        │ opensearch-go/go.mod │
│ https://osv.dev/GO-2025-4006 │      │ Go        │ stdlib  │ 1.24.0  │ 1.24.8        │ opensearch-go/go.mod │
│ https://osv.dev/GO-2025-4007 │      │ Go        │ stdlib  │ 1.24.0  │ 1.24.9        │ opensearch-go/go.mod │
[...]

PR #834, opened 2026-05-16
chore: modernize internal tests to use WaitGroup.Go

ssunno:
`golangci-lint`'s `modernize` linter flags the classic `go func() { defer wg.Done(); ... }()` pattern in internal tests — a pattern Go 1.25 collapsed into a single `WaitGroup.Go` call. CI runs the linter with `--fix`, so the suggestions silently auto-apply in CI and never surface as a review item, but contributors running the linter locally without `--fix` see the same findings on every run. Landing the change once clears the noise.

The Go bump that made `WaitGroup.Go` available landed in #825.

## Verification

`golangci-lint run --build-tags "integration core plugins plugin_security plugin_index_management multinode" --timeout=5m ./...` → 0 issues after, 2 `modernize` findings before.

`make test-integ race=true` passes locally.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -151,7 +151,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - **BREAKING**: Replace `[]json.RawMessage` with typed `[]BulkByScrollFailure` for `Failures` field in `DocumentDeleteByQueryResp`, `UpdateByQueryResp`, and `ReindexResp` ([#797](https://github.com/opensearch-project/opensearch-go/issues/797)). This is a compile-time change only — callers that were not accessing `.Failures` are unaffected, and callers that were manually unmarshaling `json.RawMessage` can now access typed fields directly.
 - Replace inline `_shards` struct with `ResponseShards` in `IndexResp`, `DocumentCreateResp`, `DocumentDeleteResp`, `UpdateResp`, `IndicesRefreshResp`, and `IndicesCountResp` to expose shard `Failures` and `Skipped` fields ([#797](https://github.com/opensearch-project/opensearch-go/issues/797)). Code accessing `resp.Shards.Total`, `resp.Shards.Successful`, or `resp.Shards.Failed` compiles unchanged.
 - Add `omitempty` to all deprecated `_type` JSON tags so empty values are omitted during marshaling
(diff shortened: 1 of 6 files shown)


Jakob3xD:
Great cleanup! While reviewing this PR, I noticed there are a few more places in the codebase with the same old `wg.Add(1)` / `defer wg.Done()` pattern that could be modernized to `wg.Go()`. They were all missed by the linter because their build tags exclude them from the `--build-tags integration` run:

| File | Lines | Build tag |
|------|-------|-----------|
| [`opensearchtransport/logger_internal_test.go`](https://github.com/opensearch-project/opensearch-go/blob/225cedad286c06515e45fd4b70ab52711c625676/opensearchtransport/logger_internal_test.go#L75-L89) | 76–78 | `!integration` |
| [`opensearchutil/bulk_indexer_internal_test.go`](https://github.com/opensearch-project/opensearch-go/blob/225cedad286c06515e45fd4b70ab52711c625676/opensearchutil/bulk_indexer_internal_test.go#L304-L320) | 306–308 | `!integration` |
| [`opensearchutil/bulk_indexer_internal_test.go`](https://github.com/opensearch-project/opensearch-go/blob/225cedad286c06515e45fd4b70ab52711c625676/opensearchutil/bulk_indexer_internal_test.go#L438-L453) | 440–442 | `!integration` |

Note: `bulk_indexer_internal_test.go` line 776 uses `wg.Add(int(numItems))` before a loop with separate `wg.Done()` calls — that one cannot be replaced with `wg.Go()`.

Would you like to include these in this PR, or should I open a separate follow-up issue?

sean-:
The same `WaitGroup.Go()` rewrite for `opensearchtransport/connection_warmup_internal_test.go` and `opensearchtransport/selector_round_robin_internal_test.go` already lives on two other in-flight branches, so this would land as a no-op once either rebases past #834:

- [`e51f9d3`](https://github.com/sean-/opensearch-go/commit/e51f9d363d52d8fc19c4032e4f4939bc81fb0547) — *Modernize opensearchtransport with Go 1.23 idioms* (on `gh-650`)
- [`a8d6568`](https://github.com/sean-/opensearch-go/commit/a8d6568175aff7b719632483d4646ec8676c9282) — *style(transport): fix goconst and modernize lint issues* (on `gh-814`)

No objection to merging #834 as the focused, standalone change, but FYI it's effectively a noop once other PRs land.

Jakob3xD (approved):

PR #841, opened 2026-05-25
Add WaitForCompletion to IndicesForcemergeParams

### LINKED ISSUE #840: Indices.Forcemerge: missing wait_for_completion parameter
## Summary

`Indices.Forcemerge` is the only long-running OpenSearch API in this client that doesn't expose `wait_for_completion` as a typed option, forcing callers to drop down to `client.Transport.Perform()` with a manually-constructed `*http.Request` to invoke async force-merge.

## Detail

The OpenSearch HTTP API supports `POST /<index>/_forcemerge?wait_for_completion=false` (server-side support since OpenSearch 2.7.0). The server returns a `{ "task": "<node>:<id>" }` payload that callers can then poll via `_tasks/<id>`. This is the recommended pattern for force-merging large indices because synchronous force-merge calls routinely exceed reasonable HTTP read timeouts and break across intermediate proxies/load balancers.

Every other long-running API in `opensearch-go` exposes this:

| API | `WithWaitForCompletion` typed option |
|---|---|
| `Reindex` | ✅ |
| `UpdateByQuery` | ✅ |
| `DeleteByQuery` | ✅ |
| `Indices.Upgrade` | ✅ |
| `Snapshot.Create` | ✅ |
| `Snapshot.Restore` | ✅ |
| `Tasks.Get` / `Tasks.Cancel` / `Tasks.List` | ✅ |
| **`Indices.Forcemerge`** | ❌ |

Verified missing in every released version from v1.1.0 through v4.6.0.

## Repro

```go
package main

import (
	"context"
	"github.com/opensearch-project/opensearch-go"
)

func main() {
	c, _ := opensearch.NewDefaultClient()
	ctx := context.Background()

	// Compiles fine — Reindex has it:
	_, _ = c.Reindex(nil, c.Reindex.WithContext(ctx), c.Reindex.WithWaitForCompletion(false))

	// Compile error — Forcemerge does not:
	_, _ = c.Indices.Forcemerge(
		c.Indices.Forcemerge.WithContext(ctx),
		c.Indices.Forcemerge.WithIndex("my-index"),
		c.Indices.Forcemerge.WithMaxNumSegments(1),
		c.Indices.Forcemerge.WithWaitForCompletion(false), // ← undefined
	)
}
```

```
./main.go:16:24: c.Indices.Forcemerge.WithWaitForCompletion undefined
  (type opensearchapi.IndicesForcemerge has no field or method WithWaitForCompletion)
```

## Motivation

Without a typed option, callers who need async force-merge have to bypass the typed wrapper:

```go
req, _ := http.NewRequestWithContext(ctx, http.MethodPost,
    fmt.Sprintf("/%s/_forcemerge?max_num_segments=1&wait_for_completion=false", index), nil)
res, err := client.Transport.Perform(req)
```

This works (and we use it in production today) — the `Transport` field is exported and goes through the same middleware chain — but it's a smell that the typed API forces this for what is a first-class server-side feature on the most common heavy-write API. Sync force-merge is impractical for indices large enough to actually need force-merging.

## Proposed change

`opensearchapi/api_indices-forcemerge-params.go` (v3/v4) or `api.indices.forcemerge.go` (v1/v2):

```go
type IndicesForcemergeParams struct {
    AllowNoIndices       *bool
    ExpandWildcards      string
    Flush                *bool
    IgnoreUnavailable    *bool
    MaxNumSegments       *int
    OnlyExpungeDeletes   *bool
    WaitForCompletion    *bool   // ← new
    // ...
}
```

…with the matching `params["wait_for_completion"] = strconv.FormatBool(*r.WaitForCompletion)` in the `get()` / `Do()` body. For v1/v2's functional-options style, add a `WithWaitForCompletion(bool) func(*IndicesForcemergeRequest)`.

Happy to send a PR if it would be useful — wanted to check there isn't a deliberate reason for the omission first.

codersofthedark:
## Description

The OpenSearch `_forcemerge` HTTP API supports `wait_for_completion` (server-side since OpenSearch 2.7.0), and every other long-running API in this library exposes it on the typed wrapper. `Indices.Forcemerge` is the only async-capable API where the typed wrapper omits it, forcing callers who need non-blocking force-merge to drop down to `client.Transport.Perform` with a manually-built `*http.Request`.

This PR adds the field to `IndicesForcemergeParams` and the corresponding URL parameter encoding, matching the existing pattern used by sibling params (`OnlyExpungeDeletes`, `Flush`, etc).

## Issues Resolved

Closes #840

## Testing

- Added a unit test (`opensearchapi/api_indices-forcemerge-params_test.go`) verifying `WaitForCompletion = &true`/`&false` is encoded into the request URL as `wait_for_completion=true`/`false`, plus an isolated case for the async (`wait_for_completion=false`) call shape. Test mirrors the pattern used by `api_indices-validate-params_test.go`.
- `go build ./...` passes.
- `go test ./opensearchapi/... -count=1` passes (full opensearchapi suite, non-integration).

## Check List

- [x] New functionality includes testing.
- [x] All tests pass.
- [x] New functionality has been documented in code comments where appropriate.
- [x] Commits are signed per the DCO using `--signoff`.
- [x] CHANGELOG updated under `[Unreleased] / Added`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ### Added
 
+- Add `WaitForCompletion` field to `IndicesForcemergeParams` to enable non-blocking force-merge via `wait_for_completion=false`, matching the existing pattern on sibling async-capable APIs (Reindex, DeleteByQuery, UpdateByQuery, Snapshot.Create, Indices.Upgrade, Tasks.*) ([#840](https://github.com/opensearch-project/opensearch-go/issues/840))
 - Add `primary_terms_map` and `split_shards_metadata` fields to ClusterState index metadata for OpenSearch >=3.6.0 compatibility
 - Add generic `opensearch.Do[T]()` function for compile-time pointer enforcement on response types, preventing a class of bugs where non-pointer values are silently passed to `Client.Do()` and fail at runtime during JSON unmarshaling. Includes `opensearch.NoBody` marker type for calls that expect no response body, unifying all internal dispatch through a single generic path ([#809](https://github.com/opensearch-project/opensearch-go/pull/809))
 - Add `InsecureSkipVerify` config option to disable TLS certificate verification without constructing a custom `http.Transport`, preserving `DefaultTransport` connection pooling, HTTP/2, and timeout defaults ([#786](https://github.com/opensearch-project/opensearch-go/issues/786))
--- a/opensearchapi/api_indices-forcemerge-params.go
+++ b/opensearchapi/api_indices-forcemerge-params.go
@@ -39,6 +39,7 @@ type IndicesForcemergeParams struct {
 	IgnoreUnavailable  *bool
 	MaxNumSegments     *int
 	OnlyExpungeDeletes *bool
+	WaitForCompletion  *bool
 
 	Pretty     bool
 	Human      bool
@@ -71,6 +72,10 @@ func (r IndicesForcemergeParams) get() map[string]string {
 
 	if r.OnlyExpungeDeletes != nil {
 		params["only_expunge_deletes"] = strconv.FormatBool(*r.OnlyExpungeDeletes)
+	}
+
+	if r.WaitForCompletion != nil {
+		params["wait_for_completion"] = strconv.FormatBool(*r.WaitForCompletion)
 	}
 
 	if r.Pretty {
--- a/opensearchapi/api_indices-forcemerge-params_test.go
+++ b/opensearchapi/api_indices-forcemerge-params_test.go
@@ -1,0 +1,117 @@
+// SPDX-License-Identifier: Apache-2.0
+//
+// The OpenSearch Contributors require contributions made to
+// this file be licensed under the Apache-2.0 license or a
+// compatible open source license.
+//
+//go:build !integration
+
+//nolint:testpackage // to test unexported get() method
+package opensearchapi
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestIndicesForcemergeParams_get(t *testing.T) {
+	type fields struct {
+		AllowNoIndices     *bool
+		ExpandWildcards    string
+		Flush              *bool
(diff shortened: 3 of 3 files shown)


sean-:
This manual addition is now handled automatically by PR #804, which improves the code generator to convert all 352 boolean query parameters across the client to use `*bool` pointers.

The broader changes in #804 regenerate all operations with corrected boolean semantics. Since the spec already lists `wait_for_completion` as a parameter, the generator now emits it correctly—eliminating the need for one-off manual field additions like this PR.

**Generated output (from PR #804):**

```go
// IndicesForcemergeParams represents query parameters for the IndicesForcemergeReq.
type IndicesForcemergeParams struct {
      // ... other params ...

      // Should the request wait until the force merge is completed.
      //
      // Available: >= 2.7.0.
      //
      // Default: true.
      WaitForCompletion *bool
}

func (r IndicesForcemergeParams) get() map[string]string {
      var params map[string]string
      set := func(k, v string) {
              if params == nil {
                      params = make(map[string]string)
              }
              params[k] = v
      }
      // ... other params ...

      if r.WaitForCompletion != nil {
              set("wait_for_completion", strconv.FormatBool(*r.WaitForCompletion))
      }

      return params
}
```

This output matches your test pattern: `WaitForCompletion: &true // &false` except it is now generated automatically for all 352 boolean parameters across the client.

sean-:
Quick correction on my previous comment: PR #804 fixed double-slash URL paths across `GetRequest` methods but did not regenerate `IndicesForcemergeParams`, so #841 is still needed for the v4 surface.

That said, the `v5preview` generator already picks this up correctly from the spec. `v5preview/opensearchapi/indices-forcemerge_gen.go` and has has both `WaitForCompletion *bool` and `PrimaryOnly *bool` (the v4 hand-written file is also missing `PrimaryOnly`). Once the `v5` branch is cut and `v5preview` moves to `opensearchapi/`, this functionality will be present without any manual additions.

Given that, I'm going to close this out.  If you want to rebase something for v4, I can review and merge it, but with v5 coming imminently, this PR will be obsolete.  Thank you for your contribution, however!, this was part of the push to automate the creation of the v5 API.

PR #844, opened 2026-05-29
feat: per-wrapper partial-failure error mask + v5preview parity

sean-:
## Summary

Replace `Config.ReturnQueryErrors` (a single boolean) with a 15-bit `errmask.ErrorMask` bitfield, where each bit corresponds to one wrapper schema in the proposed `x-error-responses` OpenAPI spec extension. Generators read those annotations, so `v5preview/opensearchapi` gets the same partial-failure handling that v4 has had as hand-written code. The same branch also reworks discriminated-union decoding (single-pass merge for success|error unions, lazy `As<T>()` for discriminator-less aggregation/suggest unions) and makes node discovery blocking.

Closes #816 (final piece of the gh-816 burn-down stack: #804 / #815 / #801 already merged).

## Why

OpenSearch returns HTTP 200 for partial successes -- bulk item failures, search-shard failures, single-doc replica failures, MSearch sub-response errors -- so callers must remember a second check after `err == nil`. v4's `ReturnQueryErrors` bool flipped _all_ partial-failure conversions on or off as one switch; users wanting shard-level errors but tolerating bulk item failures (or vice versa) had no way to express it.

The bitfield gives one bit per wire shape. Setting a bit _masks_ (suppresses) that category; the unset state reports it. `errmask.Empty` reports everything; `errmask.All` masks everything.

## Lifecycle

| Version      | `Config.Errors == nil` default | Behavior                                                                                                                 |
| ------------ | ------------------------------ | ------------------------------------------------------------------------------------------------------------------------ |
| v4 (this PR) | `errmask.All`                  | No behavior change for existing callers. Opt in to typed errors with `Config.Errors: errmask.New()`.                     |
| v5           | `errmask.Empty`                | Every category reported by default. Callers opt out individual bits via `Config.Errors` or `OPENSEARCH_GO_ERROR_MASK`.   |
| v6           | n/a                            | `Config.Errors` and `OPENSEARCH_GO_ERROR_MASK` removed. Behavior is unconditionally `errmask.Empty` (report everything). |

## Usability

The caller surface is a typed error return + a typed response. Both are non-nil on partial failure: the response is fully populated even when the error fires. `Config.Errors` is a `*errmask.ErrorMask`; because the named values are constants (not addressable), build the pointer with `errmask.New(...)`.

```go
client, _ := opensearchapi.NewClient(opensearchapi.Config{
    Client: opensearch.Config{Addresses: addrs},
    Errors: errmask.New(), // report every partial-failure category
})

[...]

PR #857, opened 2026-06-04
test(osgen): pin perOpErrorTypeName <-> errwrap.OperationWrappers coupling

sean-:
`perOpErrorTypeName`'s hardcoded switch and `errwrap.OperationWrappers`' wrapper-count map are coupled by an unstated invariant: a group has a per-op aggregator type iff its catalog entry declares 2+ wrappers. Today both sides match, but nothing checks them, so a future catalog edit can desync the two without any signal -- the dispatch keeps referencing a per-op type that's no longer reachable, or worse, emits an empty type name when a 2+-wrapper group lacks a switch arm.

Add a coupling test that asserts both directions:
  - every group naming a per-op aggregator type has 2+ wrappers in `OperationWrappers`
  - every catalog entry with 2+ wrappers has a non-empty per-op aggregator type

Iterates the catalog directly rather than a duplicate list of switch arms, so a new switch arm or catalog entry is exercised automatically.

Failure messages are actionable: they name the offending group, the current state, and the remediation (add wrappers, remove switch arm, or add a hand-written aggregator).

Ref: opensearch-project/opensearch-go#844 (review round 3, F7)


cmd/osgen/emit/frag_dispatch_test.go
```diff
@@ -306,6 +306,52 @@ func TestDispatchFragment_Imports(t *testing.T) {
...
+// Without this test, a future catalog edit could desync the two
+// without any signal until generated code stops compiling.
+func TestPerOpErrorTypeName_CatalogConsistency(t *testing.T) {
+	t.Parallel()
+
+	// Forward direction: every group that names a per-op aggregator type
+	// must have 2+ wrappers in the catalog. Iterating the catalog itself
+	// (rather than a duplicate list of switch arms) means a new switch
+	// arm is exercised the moment any group ever appears in either side.
+	for group := range errwrap.OperationWrappers {
```
Jakob3xD, line 346: Non-blocking / forward-looking note — the test is correct and addresses the #844 F7 coupling concern. Just flagging a fidelity gap between what it pins and what its doc comment claims.

This pins against `errwrap.OperationWrappers` (the declared catalog), but whether the dispatch actually emits a per-op aggregator type is decided by `len(emittable) >= 2` from `emittableWrappers()` — a different set. `emittable` is filtered through the emission `wrappers` map + `Applies`, and `Op.ErrorWrappers` is resolved by `resolveErrorWrappers` from the spec extension `x-error-responses` first, with the catalog only as a fallback. Both loops here range over the catalog, never the switch arms.

Consequences (none live today — the catalog and emittable sets coincide for `msearch`/`msearch_template`):
- A switch arm added for a group absent from `OperationWrappers` is visited by neither loop — exactly the "dispatch references an unreachable per-op type" case the doc comment says it prevents.
- A future catalog entry with 2 wrappers where only one has an emission path would pass this test, yet `emittable` stays at 1, so no per-op type is emitted.

So the docstring (L313-318) and the CHANGELOG entry slightly over-claim: they promise coverage of the `emittable`-driven "unreachable per-op type / empty type name" desync, which this test doesn't measure. Two options: soften the wording to "pins the catalog ↔ switch coupling", or additionally assert against `emittableWrappers` / enumerate the switch arms. (F7 named the coupling to `emittableWrappers` specifically.)

Jakob3xD (commented):
### Code review

The test is solid and addresses the #844 F7 coupling concern — it compiles, isn't flaky, and catches the most likely desync (a catalog group dropping below 2 wrappers while the dispatch still names a per-op type). One optional, forward-looking note posted inline; nothing blocking.

cmd/osgen/emit/frag_dispatch_test.go
sean-: > Non-blocking / forward-looking note — the test is correct and addresses the #844 F7 coupling concern. Just flagging a fidelity gap between what it pins and what its doc comment claims.

Good catch, you're right that the doc comment and CHANGELOG were over-claiming. The test pins the *catalog <-> switch* coupling, not the *runtime emission* coupling, and those are different sets once `emittableWrappers` filters through the `wrappers` map + `Applies`, and once `resolveErrorWrappers` prefers `x-error-responses` over the catalog.

Fixed in the latest push by taking both options you offered:

**1. Softened the doc comment + CHANGELOG.** The docstring now says "catalog <-> switch coupling" and has an explicit "What this does NOT pin" paragraph that names `emittableWrappers` / `resolveErrorWrappers` / `x-error-responses` so the next reader sees the gap.

**2. Added a third assertion direction.** The original test ranged over `OperationWrappers` in both loops, so a switch arm whose group is *absent* from the catalog was invisible to it, which is exactly the "dispatch references an unreachable per-op type" case the docstring used to promise. Added an explicit `switchGroups` slice (currently `GroupMSearch`, `GroupMSearchTemplate`) and a third subtest direction:

```go
// (3) Switch-arm catalog presence: every group named by the
// per-op switch must appear in OperationWrappers. A switch arm
// for a group missing from the catalog is dead code: neither
// loop above iterates it, so without this check it could
// outlive the catalog entry that justified it.
for _, group := range switchGroups {
    t.Run("switch_arm_in_catalog_"+group, func(t *testing.T) {
        t.Parallel()
        _, ok := errwrap.OperationWrappers[group]
        require.True(t, ok,
            "perOpErrorTypeName has a switch arm for group %q but the group is absent from errwrap.OperationWrappers; remove the arm or restore the catalog entry",
            group)
    })
}
```

The remaining emittable-driven case (a catalog entry with 2 wrappers where only one survives Applies, so `len(emittable) == 1` and no per-op type is emitted) still isn't pinned here — that's a richer fixture to build and felt out of scope for this test, so I called it out in the docstring instead. Happy to follow up with a separate PR if you'd rather pin it now.

Jakob3xD (approved):
Re-reviewed on 63d73ca7. My earlier non-blocking note is fully addressed: the hand-maintained `switchGroups` list plus assertion (3) now close the "switch arm for a catalog-absent group is visited by neither loop" gap, and the doc comment's new "What this does NOT pin" paragraph correctly scopes the test to the catalog <-> switch coupling (documenting the `emittable` / `resolveErrorWrappers` / `x-error-responses` boundary rather than over-claiming it).

Verified: `errwrap.GroupMSearch` / `GroupMSearchTemplate` exist and carry 2 wrappers each, all three assertions are sound, and the parallel subtests are safe under go 1.25 loop-var semantics. LGTM.

PR #859, opened 2026-06-05
fix(transport): header override, gzip pool, body-read error, and body-close correctness

sean-:
Small correctness/robustness fixes found during a pre-v5 audit of the transport and error-handling paths. None are security-sensitive; all are ordinary bugs present in v4.6.0.

1. *`opensearchtransport`: honor per-request header override over global headers*: `setReqGlobalHeader` compared the per-request header **value** against the global header **name**, so a request-level header never suppressed the global default — both were sent. Now checks for key presence via `req.Header[http.CanonicalHeaderKey(k)]`.

2. *`opensearchtransport`: prevent gzip buffer-pool nil poisoning on compress error*: On `io.Copy` or `writer.Close` failure, `compress()` returned `(nil, err)` while the caller's `defer collectBuffer(buf)` still ran, putting a typed-nil `*bytes.Buffer` into the `sync.Pool`. The next `Get()` would hand that nil out and panic on `Reset()`. Now returns the buffer on error so it's recycled, and `collectBuffer` nil-guards defensively.

3. *`opensearchtransport`: surface body-read errors during response buffering*: The post-`Perform` buffering block used `:=` and shadowed the outer `err`, silently dropping `io.ReadAll` failures. Now propagates the read error and adjusts `Client.Do` to handle the new `(resp != nil, err != nil)` case while preserving the existing `ErrReadBody` contract.

4. *Close error-response body in `ParseError` and v5preview `do()`*: `ParseError` read the body but never closed the original `io.ReadCloser` before replacing it with a `NopCloser`, and the v5preview `do()` error path (when `dataPointer == nil`) returned without draining/closing the body. Both prevented `http.Transport` from reusing the connection on error responses.


opensearch.go
```diff
@@ -443,6 +443,10 @@ func (c *Client) Do(ctx context.Context, method string, req Request, dataPointer
 		Body:       resp.Body,
 	}
 
+	if err != nil {
```
Jakob3xD, line 516: `ErrReadBody` is used here as a catch-all label for any `(resp != nil, err != nil)` returned by `Perform`. But that combination also arises on a pre-existing path unrelated to body reads: if the context is cancelled during the retry backoff (`opensearchtransport.go:1436-1449`) after a retryable 429/5xx response was received, `Perform` returns `(res != nil, context.Canceled)`. This block then wraps it as `ErrReadBody`, so `errors.Is(err, ErrReadBody)` becomes a false positive (no body read occurred) and the message reads `failed to read body, status: 503, err: context canceled`.

(`errors.Is(err, context.Canceled)` still holds thanks to the double `%w`, so cancellation detection itself is intact — the problem is the misclassification and misleading message.)

Suggested fix: have `Perform` wrap the read error in a detectable sentinel (e.g. `ErrReadBody`) and classify here via `errors.Is`, rather than assuming every `(resp != nil, err != nil)` is a read failure.

v5preview/opensearchapi/api.go
```diff
@@ -243,9 +244,13 @@ func do[T any](ctx context.Context, c *Client, method string, req opensearch.Req
 	if resp.IsError() {
 		if dataPointer != nil {
 			return resp, opensearch.ParseError(resp)
-		} else {
-			return resp, fmt.Errorf("status: %s", resp.Status())
 		}
+		if resp.Body != nil {
+			//nolint:errcheck // best-effort drain so the connection can be reused
+			io.Copy(io.Discard, resp.Body)
```
Jakob3xD: In the default buffered mode (`DisableResponseBuffering=false`), `Perform` has already read the body into an in-memory `io.NopCloser` and closed the connection — so this drain is unnecessary and leaves the returned `resp` with an emptied body. That is a regression versus the previous behavior, and inconsistent with the `dataPointer != nil` branch above, which routes through `ParseError` and re-wraps the bytes (`io.NopCloser(bytes.NewReader(...))`) so the error body stays readable. The connection-reuse benefit only applies under `DisableResponseBuffering=true`.

Suggested fix: re-wrap the body like `ParseError` does, or only drain when `c.disableResponseBuffering` is set.

Jakob3xD (commented):
### Code review

Two issues are posted as inline comments (the `ErrReadBody` labeling in `Client.Do` and the v5preview error-body drain). Below are additional non-blocking observations, plus pre-existing instances of the same body-lifecycle bug class surfaced during a deeper audit.

#### Review notes (no specific added line)

1. After the body-read fix, `Perform` can return `(resp != nil, err != nil)`, but its godoc still reads "returns a response or error". Callers must now check `resp == nil` (not `err != nil`) to detect a hard transport failure — `Client.Do` was updated, but the `Interface` is implemented by custom types too. Worth documenting the new invariant on both.

https://github.com/opensearch-project/opensearch-go/blob/5bbbd75ed24205ee9fcd3fc5d9d00d6d46c6efbb/opensearchtransport/opensearchtransport.go#L1122-L1124

2. The CHANGELOG entry for the body-close fix states it prevents `http.Transport` from reusing the connection on error responses. That holds only under `DisableResponseBuffering`; in the default config `Perform` already buffers and closes the body, so the connection is already reusable. Recommend scoping the claim to the non-buffered mode.

3. No tests accompany the five fixes. The gzip nil-poisoning panic, the header-override suppression, and the `:=` shadowing are all trivially unit-testable. Note the existing override test passes even against the buggy code, because it only asserts `req.Header.Get(...)` and never `len(req.Header[...]) == 1` — so it can neither detect the duplicate-header regression nor confirm the fix.

#### Related, pre-existing instances of the same bug class (not introduced by this PR)

The PR fixes "close/drain the body on error" where the author found it, but the same class persists on direct `RoundTrip` paths that lack `Perform`'s buffering safety net — there keep-alive is genuinely defeated:

- Stats poller closes the body without draining on the non-200 path:

https://github.com/opensearch-project/opensearch-go/blob/5bbbd75ed24205ee9fcd3fc5d9d00d6d46c6efbb/opensearchtransport/cluster_health.go#L227-L233

- Discovery `/_cat/shards` and `/_cluster/state/metadata` non-200 paths close without draining:

https://github.com/opensearch-project/opensearch-go/blob/5bbbd75ed24205ee9fcd3fc5d9d00d6d46c6efbb/opensearchtransport/discovery.go#L1557-L1561

https://github.com/opensearch-project/opensearch-go/blob/5bbbd75ed24205ee9fcd3fc5d9d00d6d46c6efbb/opensearchtransport/discovery.go#L1804-L1808

- The discovery `json.Decode` success paths (`discovery.go:1132/1572/1835`) don't read to EOF, so the deferred `Close` often runs before EOF — same keep-alive defeat on healthy responses.

- `Response.String()` consumes the body via `io.ReadAll` without restoring it (state footgun; no leak since the body is already buffered):

https://github.com/opensearch-project/opensearch-go/blob/5bbbd75ed24205ee9fcd3fc5d9d00d6d46c6efbb/response.go#L82-L88

- The signer read-error paths (`signer/aws/aws.go:159`, `signer/awsv2/sdkv2signer.go:103`) return without closing the request body.

These are out of scope for this PR, but flagging them since they are the same class. The cleanest long-term fix is to own body lifecycle in one layer — e.g. fix the codegen template `cmd/osgen/emit/frag_plugin.go:159` rather than hand-patching a single `do()`.

opensearch.go
sean-: Good catch, and you're right about the root cause. `Perform` returns `(res != nil, context.Canceled)` from the retry-backoff path, and `Client.Do` was labeling any `(resp != nil, err != nil)` as `ErrReadBody`, so cancellation during backoff got the misleading "failed to read body" prefix even though no read was attempted.

Fixed as you suggested: `Perform` now wraps the buffering read error in a new detectable sentinel, `opensearchtransport.ErrResponseBodyRead`, and `Client.Do` classifies via `errors.Is` -- only a genuine body-read failure is wrapped as `ErrReadBody`; any other error returned alongside a response is surfaced with its identity intact and no read-failure prefix. `errors.Is(err, context.Canceled)` still holds in both cases.

Added a table-driven test covering both arms (the body-read sentinel maps to `ErrReadBody`; `context.Canceled` does not and keeps its message).

v5preview/opensearchapi/api.go
sean-: Agreed, the unconditional drain was a regression in buffered mode. The no-decode error path now reads the body to EOF and re-wraps it in an `io.NopCloser`, mirroring the `ParseError` branch, so the returned `resp.Body` stays readable for the caller. Reading to EOF still lets `http.Transport` reuse the connection under `DisableResponseBuffering`; in the default buffered mode the body is already an in-memory `NopCloser`, so the read is cheap and non-destructive.

Added a unit test asserting the error-path body is still readable rather than drained to empty.

sean-:
Thanks for the review. The two inline items are addressed in the threads above; here's everything else.

Review notes
- Documented the (`resp != nil`, `err != nil`) invariant on both `Perform` and the `Interface` contract: callers must treat `resp == nil`, not `err != nil`, as the signal for a hard transport failure.
- Scoped the `CHANGELOG` body-close entry to `DisableResponseBuffering` -- in the default buffered mode `Perform` already drains and closes the body, so the connection is already reusable; the fix only changes reuse behavior in the non-buffered mode.
- Added unit tests for all five original fixes (`gzip` `nil`-poisoning, header override asserting `len == 1`, body-read error surfacing, the `Do` error classification, and `v5preview` error-body readability). Each was confirmed to fail against the pre-fix code and pass after.

Same-class instances from your audit (fixed here, not deferred)
- Drain-before-close on the raw `RoundTrip` paths that lack `Perform`'s buffering: the stats poller (`cluster_health.go`) and `discovery`'s `/_cat/shards`, `/_cluster/state/metadata`, and `/_nodes`. This covers both the non-`200` returns and the `json.Decode` success paths, which stop at the end of the JSON value without consuming trailing bytes. The drain/close is inlined at each deferred cleanup so `bodyclose` can still verify the close statically.
- `Response.String()` is now non-consuming: it restores `Body` after rendering, so logging a response no longer empties a body other code expects to read.
- The AWS v1/v2 signers close the request body on the read-error path in `hexEncodedSha256OfRequest`.

On the codegen-template suggestion (own body lifecycle in one layer): I agree that's the right long-term direction. I'd like to keep this PR focused on the transport fixes and the same-class cleanups above, and open a follow-up to track centralizing body lifecycle in the `do()` template.

Tooling note: while validating, I found `make lint.local` *and* the CI lint job passed a single build-tag set including `integration` and `multinode`, which silently skipped every `!integration` unit-test file and every `!multinode` single-node integration file -- so a class of files was never linted. I added a build-tag matrix (one `golangci-lint run` per `integration`/`multinode` combination, feature tags unioned in) and fixed the pre-existing findings that surfaced. Those are in their own commits, separate from the transport fixes.

sean-:
I picking up the codegen-template thread from [#859 (comment)](https://github.com/opensearch-project/opensearch-go/pull/859#issuecomment-4650691087): the new commits stacked onto this PR split `Perform` and unbuffered streaming into two distinct entry points. The result is one rule per surface, no hidden flag.

- **`(*opensearchtransport.Client).Perform`** -- always buffers. Reads the response body to EOF, closes it, and replaces `res.Body` with an `io.NopCloser` over a `bytes.Reader`. Read failures still surface via `ErrResponseBodyRead`. Marked `// Deprecated:` for removal in v5; current v4 callers see no behavior change because `Do[T]` keeps routing through it.
- **`(*opensearchtransport.Client).Stream(req)`** -- never buffers. Returns the raw `RoundTrip` body; the caller owns reading and closing `res.Body`. Same retry/seed-fallback/signing path as `Perform`, just without the trailing buffer step.
- **`(*opensearch.Client).Stream`** passthrough so callers don't have to type-assert `c.Transport`. No typed `DoStream`/`DoReader` -- "stream and decode into `T`" is a contradiction (want `T` -> `Do[T]`; want bytes -> `Stream`), and adding one would invite the same lifecycle drift back.
- **`DisableResponseBuffering` is gone.** Untagged, never released, no migration path. `Perform` always buffers; `Stream` never does. The behavior used to depend on a construction-time flag; it now depends on which method you call.
- `opensearchtransport.Interface` stays unchanged in v4 (a `TODO(v5)` notes that v5 should add `Stream` and remove the deprecated `Perform`).

The no-decode error-path drain that #859 added in the plugin `do[T]` helpers was only meaningful when `DisableResponseBuffering=true`. With `Perform` now always buffering and the flag gone, the drain was dead -- `opensearch.Do` routes through buffered `Perform`, so `resp.Body` is already a `NopCloser` over an in-memory `bytes.Reader` by the time the helper sees it. Removed this dead code from the `cmd/osgen` plugin template and the four hand-written copies (`opensearchapi`, `v5preview/opensearchapi`, `plugins/security`, `plugins/ism`); `make regen` regenerated `v5preview/opensearchapi/plugins/**/client_gen.go` accordingly.

The same-class fixes you flagged (stats poller and discovery's `/_cat/shards`, `/_cluster/state/metadata`, `/_nodes` drain-before-close; `Response.String()` non-consuming; AWS v1/v2 signer body close) stay in -- those are on direct `RoundTrip` paths that bypass `Perform`, so they still need explicit drain/close regardless of how `Perform` is shaped.

Jakob3xD (commented):
### Code review

Re-review on the updated head (`49cc871`). The prior round is correctly addressed — the `ErrReadBody` vs `context.Canceled` classification (now via the `ErrResponseBodyRead` sentinel + `errors.Is`), the v5preview error-body readability, the raw-`RoundTrip` drain cleanups in `cluster_health.go`/`discovery.go`, the signer body-close, the documented `Perform` invariant, the scoped CHANGELOG, and the added regression tests all check out. Not re-raising those.

One issue is posted as an inline comment (`Do` can surface a phantom 503 alongside `ErrNoConnections`). Additional notes below.

#### Same-class drain still missing in three sibling pollers

The drain-before-close cleanup covered `cluster_health.go` and `discovery.go`, but three health-check functions in `opensearchtransport.go` use the same raw `RoundTrip` pattern and still close the body without draining on the non-200 path — defeating keep-alive on the exact poller hot path the cleanup targeted:

`fetchClusterHealth`:

https://github.com/opensearch-project/opensearch-go/blob/49cc871fbba46aef81dbdbb2709b7da3e0f17926/opensearchtransport/opensearchtransport.go#L2112-L2120

`baselineHealthCheck`:

https://github.com/opensearch-project/opensearch-go/blob/49cc871fbba46aef81dbdbb2709b7da3e0f17926/opensearchtransport/opensearchtransport.go#L1912-L1916

`hardwareInfoHealthCheck`:

https://github.com/opensearch-project/opensearch-go/blob/49cc871fbba46aef81dbdbb2709b7da3e0f17926/opensearchtransport/opensearchtransport.go#L1985-L1989

Each does a bare `res.Body.Close()` on `res.StatusCode != http.StatusOK` without a preceding `io.Copy(io.Discard, res.Body)`. Same fix as the discovery/cluster_health paths.

#### Minor

1. `error.go` `ParseError`: on the `io.ReadAll` error path the body is `Close()`d but not replaced with a `NopCloser`, so partial bytes are lost and `resp.Body` is left closed. Harmless in the default buffered path (already a `NopCloser`), but inconsistent with the success branch right below it.

https://github.com/opensearch-project/opensearch-go/blob/49cc871fbba46aef81dbdbb2709b7da3e0f17926/error.go#L146-L153

2. `response.go` `String()` receiver changed value→pointer. Correct — the body-restore needs a pointer receiver — but it's a silent API-surface change on an exported type (a `Response` value no longer satisfies `fmt.Stringer`). Worth a CHANGELOG line.

https://github.com/opensearch-project/opensearch-go/blob/49cc871fbba46aef81dbdbb2709b7da3e0f17926/response.go#L89-L96

3. `opensearchapi/testutil/helpers.go`: switching `assert.Empty` → `require.Empty` makes the following per-operation `fmt.Printf` diagnostic loop unreachable (require halts the test first), so a failing schema comparison now prints nothing. Move the assertion after the loop, or drop the now-dead loop. (test-only)

https://github.com/opensearch-project/opensearch-go/blob/49cc871fbba46aef81dbdbb2709b7da3e0f17926/opensearchapi/testutil/helpers.go#L433-L441

opensearch.go
```diff
@@ -433,7 +466,7 @@ func (c *Client) Do(ctx context.Context, method string, req Request, dataPointer
 
 	//nolint:bodyclose // body got already closed by Perform, this is a nopcloser
 	resp, err := c.Perform(httpReq)
-	if err != nil {
+	if resp == nil {
```
Jakob3xD, line 506: This is the very invariant this PR documents — `resp == nil` (not `err != nil`) signals a hard transport failure. But `Stream` can return a **non-nil stale response** on a hard *routing* failure, breaking that invariant: after a retryable 429/5xx is received and drained, if the next retry iteration gets `ErrNoConnections` from the pool/router, the retry loop does `if errors.Is(err, ErrNoConnections) { break }` **without resetting `res`** — so `res` still holds the earlier 503. When seed-fallback is off (`OPENSEARCH_GO_FALLBACK=false`) or no seed pool is configured (so the post-loop `performSeedFallback` is skipped), that stale `res` survives. `Perform` then buffers it and returns `(staleRes{503}, …ErrNoConnections)`, and this block falls through to `return response, fmt.Errorf("status: %d, …", resp.StatusCode, err)` — surfacing a phantom 503 from an abandoned connection on what is actually a connection-exhaustion failure.

(`errors.Is(err, ErrNoConnections)` still holds, so a caller checking the error is fine; the problem is the non-nil `response` contradicting the documented `resp == nil` contract.)

Suggested fix: reset `res = nil` before the `ErrNoConnections` break in `Stream`, since no round-trip occurred on that iteration.

sean-: Good catch, and you're right about the root cause. The `ErrNoConnections` break in `Stream` skipped past the documented `resp == nil` invariant: the retry loop received a retryable 5xx, drained it, then on the next iteration the pool/router returned `ErrNoConnections` and broke out *without resetting `res`*. With seed fallback off (or no seed pool configured), `Perform` then buffered the stale 503 and returned `(staleRes{503}, ErrNoConnections)`, contradicting the "`resp == nil` signals a hard transport failure" contract that this PR's `Client.Do` relies on.

Fixed as you suggested -- `res = nil` before the break. No round-trip occurred on that iteration, so a non-nil response there was always a stale carry-over.

```go
if errors.Is(err, ErrNoConnections) {
    // No round-trip occurred this iteration -- clear any stale
    // response from a prior retryable status (e.g. 503 from the
    // previous attempt) so the documented (resp == nil signals
    // hard transport failure) invariant holds when seed fallback
    // is unavailable.
    res = nil
    break
}
```

`errors.Is(err, ErrNoConnections)` continues to hold in both seed-fallback-on and seed-fallback-off paths. Squashed into the commit that introduced `Stream`.

sean-:
Thanks for the second pass. All five items are addressed; details inline below. The same-class drain audit you ran on `opensearchtransport.go` was a clean catch -- the three health-check functions were exactly the same class as the
discovery/cluster_health paths from the previous round, just easy to miss because they sit in a different file.

Quick summary of the changes (each squashed into the commit it amends):

- **Phantom 503 alongside `ErrNoConnections`** in `Stream` (opensearch.go:469): the retry loop now resets `res = nil` before the `ErrNoConnections` break, since no round-trip occurred on that iteration. The documented "`resp == nil` signals a hard transport failure" invariant is now actually upheld even when `OPENSEARCH_GO_FALLBACK=false` or no seed pool is configured. Squashed into the commit that introduced `Stream`.
- **Same-class drain** on the three sibling pollers in `opensearchtransport.go` (`fetchClusterHealth`, `baselineHealthCheck`, `hardwareInfoHealthCheck`): all three now drain to EOF before close on the non-200 and `io.ReadAll` error paths, mirroring the discovery/cluster_health pattern. Squashed into the original drain commit.
- **`ParseError` read-error path** (error.go:146-153): on `io.ReadAll` failure we now restore `resp.Body` to an `io.NopCloser` over whatever bytes ReadAll captured, so callers see a usable `Body` on either arm. Added a table-driven
`TestParseError_BodyState` covering both arms (the read-error case `iotest.ErrReader` -> `require.NotNil(resp.Body)` + readable). Squashed into the original ParseError fix.
- **`Response.String()` value -> pointer receiver** (response.go:89-96): added a CHANGELOG entry under `### Changed` documenting that a `Response` *value* (non-pointer) no longer satisfies `fmt.Stringer`, with the migration note that pointer values continue to. Squashed into the relevant fix commit.
- **`opensearchapi/testutil/helpers.go` unreachable diagnostic** (lines 433-441): moved the per-operation diagnostic loop *before* the `require.Empty` assertion so a failing schema comparison actually surfaces the diff. Also switched the loop from `fmt.Printf` to `t.Logf` while I was there -- gates output on `-v`, suppresses for passing tests, and attributes lines to the failing subtest. Squashed into the test-lint commit.

Per-fix unit tests from the previous round (gzip nil-poisoning, header override `len == 1`, body-read error surfacing, `Client.Do` error classification, v5preview error-body readability) are in the corresponding fix commits. Each was confirmed to fail against the pre-fix parent and pass against the fix; the new `TestParseError_BodyState` from this round follows the same pattern.

response.go
```diff
@@ -78,12 +79,20 @@ func NewResponse(statusCode int, body io.ReadCloser, header http.Header) *Respon
...
+// (or a later String call) still see the full payload. This matters because
+// Response.Body is typically the buffered NopCloser produced by Client.Do;
+// without the restore, calling String for logging would silently empty the
+// body other code expects to read.
+func (r *Response) String() string {
 	if r.Body != nil {
 		body, err := io.ReadAll(r.Body)
 		if err != nil {
-			body = fmt.Appendf(nil, "<error reading response body: %v>", err)
+			return fmt.Sprintf("%s <error reading response body: %v>", r.Status(), err)
```
Jakob3xD, line 97: The new `String()` doc comment says it is non-consuming (restores `Body` after rendering), but this `io.ReadAll` error path returns without restoring `r.Body` — it is left as the consumed/failed reader. Harmless in the common case (after `Do`/`Perform`, `Body` is an in-memory `NopCloser` over a `bytes.Reader` that won't fail `ReadAll`), but the stated invariant is broken on this branch and inconsistent with the success path right below.

Suggested fix: restore `r.Body = io.NopCloser(bytes.NewReader(body))` on this path too (the partial `body` is fine), or narrow the doc to "non-consuming on the success path".

Jakob3xD (commented):
### Code review

Full re-review on the final head `d1e36676`. The code is clean: all findings from the two prior rounds (header override, gzip pool, `Perform` body-read shadowing, `ParseError`/v5preview body close, the `Stream` phantom-503, the three sibling poller drains, the `ParseError` read-error restore, and the `String()`/testutil items) are verified fixed in the code, and no new functional bugs surfaced. The remaining notes are documentation/consistency only — non-blocking.

One issue is posted as an inline comment (`Response.String()` error path). Additional notes below.

#### CHANGELOG describes operations the code doesn't perform

Two `[Unreleased]` Fixed entries describe behavior that isn't in the shipped code:

1. The entry says the raw-`RoundTrip` paths "drain to EOF before close via a shared `drainAndClose` helper" — there is no `drainAndClose` helper in the tree (`git grep` finds none); the drain is inlined as `io.Copy(io.Discard, …)` + `Close()` at each site (~9 of them). Either extract the helper the note promises, or reword to an inlined drain.

https://github.com/opensearch-project/opensearch-go/blob/d1e366768b9540c0c5ebb53f56d3827a4a16c1ca/CHANGELOG.md#L218-L219

2. The same area says the v5preview no-decode path "now reads the body to EOF and re-wraps it in a `NopCloser` rather than discarding it" — but the code (`v5preview/opensearchapi/api.go`) just does `return resp, fmt.Errorf("status: %s", …)`; that drain/re-wrap was removed because `Perform` already buffered the body into a `NopCloser` (the code comment there states exactly this). The CHANGELOG line describes the removed behavior, not the shipped one.

https://github.com/opensearch-project/opensearch-go/blob/d1e366768b9540c0c5ebb53f56d3827a4a16c1ca/v5preview/opensearchapi/api.go#L238-L252

#### Minor

3. `opensearch.Client.Do` now returns `(response, err)` where it previously returned `(nil, err)` when a response was received but a later transport error occurred (e.g. context cancellation during retry backoff). That is an observable contract change — callers can now inspect the response in that error case. Worth a line in the CHANGELOG Fixed entry or the `Do` doc comment.

https://github.com/opensearch-project/opensearch-go/blob/d1e366768b9540c0c5ebb53f56d3827a4a16c1ca/opensearch.go#L487-L492

sean-:
Thanks for the thorough re-review, and for confirming the functional findings from the prior rounds are settled. All three documentation/consistency notes are now addressed:

1. `drainAndClose` *helper that doesn't exist anymore* -- reworded the CHANGELOG to describe what the code actually does: an inlined `io.Copy(io.Discard, ...)` + `Close()` at each raw-`RoundTrip` site (we deliberately keep it inlined rather than extracting a helper, so `bodyclose` can verify the close statically; a shared helper had been tried before and removed for exactly that reason). The entry now also names the three sibling pollers (`fetchClusterHealth` / `baselineHealthCheck` / `hardwareInfoHealthCheck`) that got the same treatment.
2. **v5preview` no-decode CHANGELOG describing removed behavior** -- corrected. The line no longer claims a drain/re-wrap on that path. It now states the shipped reality: `opensearch.Do` routes through the buffered `Perform`, so `resp.Body` is already an in-memory `NopCloser` over the full payload and stays readable for the caller, which is why the no-decode path needs no drain of its own.
3. `Client.Do (response, err)` contract change -- documented in both places. Added a paragraph to the `Client.Do` godoc spelling out that it may now return a non-nil `*Response` alongside a non-nil error (body-read failure during buffering, or an unrelated transport error such as context cancellation during retry backoff), and that callers detecting a hard transport failure should check `resp == nil` rather than `err != nil`. Also noted the change in the corresponding CHANGELOG Fixed entry.

response.go
sean-: Good catch -- you're right that the doc comment promised a non-consuming invariant the error path didn't honor. Fixed as you suggested: `String` now restores `r.Body` with `io.NopCloser(bytes.NewReader(body))` *before* the error check, so the partial bytes (empty on a hard `ErrReader`) remain readable and both arms behave consistently.

```go
func (r *Response) String() string {
      if r.Body != nil {
              body, err := io.ReadAll(r.Body)
              // Restore Body even on a partial read, so the non-consuming
              // invariant holds on the error path too: whatever bytes were
              // read remain available to subsequent reads.
              r.Body = io.NopCloser(bytes.NewReader(body))
              if err != nil {
                      return fmt.Sprintf("%s <error reading response body: %v>", r.Status(), err)
              }
              return fmt.Sprintf("%s %s", r.Status(), body)
      }
      return r.Status()
}
```

Added a test on the read-error path asserting Body is left non-nil and readable (drained to empty) rather than a consumed/failed reader, alongside the existing non-consuming success-path test.

error.go
```diff
@@ -146,7 +146,14 @@ func ParseError(resp *Response) error {
...
 
 	body, err := io.ReadAll(resp.Body)
+	resp.Body.Close()
 	if err != nil {
+		// Restore Body so callers don't see a closed reader. The partial
+		// bytes are intentionally dropped because the underlying read
+		// already failed; what matters is keeping the Body field
+		// consistent with the success branch below (an in-memory
+		// NopCloser the caller can safely Read or Close).
+		resp.Body = io.NopCloser(bytes.NewReader(body))
```
Jakob3xD, line 156: The comment above says "The partial bytes are intentionally dropped", but this line restores exactly those read `body` bytes as an in-memory `NopCloser`, making them available to the caller (the same intended pattern as `response.go` `String()`). So the code is correct — only the comment is misleading.

Suggest rewording, e.g.:

```go
// Restore Body so callers don't see a closed reader. Whatever bytes were
// read before the failure remain available — an in-memory NopCloser the
// caller can safely Read or Close, consistent with the success branch below.
resp.Body = io.NopCloser(bytes.NewReader(body))
```

Jakob3xD (approved):
### Code review

Re-review (round 4) of head `b0ab2265`. The functional core is clean and every finding from the three prior rounds is verified fixed at head (header override, gzip pool nil-poisoning, body-read sentinel, ParseError close, stale-503, `String()` restore). The `Perform` → `Stream` refactor preserves the historical invariants (#543 `GetBody`, retry drain, `cancelOnCloseBody`). The remaining items are documentation/comment consistency — none blocking, hence approving.

Found 2 additional non-blocking item(s) (1 more posted inline):

1. `guides/json.md` still steers callers to `client.Perform` for raw/streaming use — "Raw `*http.Response`; caller reads and closes body", "Proxying, streaming, full control needed" (L11, L94–98). After this PR `Perform` is deprecated **and** buffers + self-closes the body, so that description now matches `Stream`, not `Perform`. This contradicts the new `guides/response_buffering.md` ("New code should call `Do[T]` … or `Stream` for raw byte forwarding"). The file isn't touched by this PR, so flagging here rather than inline — consider updating the L94–98 table and the L11 intro to point at `Stream`.

https://github.com/opensearch-project/opensearch-go/blob/b0ab2265b84e6fb399af4757a02b0e4a8bed028b/guides/json.md#L94-L98

2. (low / optional) The v5preview `do()` no-decode error path returns `resp` without draining/closing, and `Client.Do` assumes `Perform` always buffers + closes (the `//nolint:bodyclose` note). That holds for the built-in transport — this is **not** a reintroduction of the #300 FD leak — but a custom `opensearchtransport.Interface` returning a live body would leak a connection on this path. The buffering-transport assumption isn't documented as a precondition. Optional: a one-line note on `Do` that a buffering transport is assumed, or keep a defensive `Close()` in the no-decode error path.

https://github.com/opensearch-project/opensearch-go/blob/b0ab2265b84e6fb399af4757a02b0e4a8bed028b/opensearch.go#L500-L508

Issue #336, opened 2023-06-20
[FEATURE] Add an exported flush function to the BulkIndexer


ae-ou:
### Is your feature request related to a problem?
I have a Lambda that consumes events from a Kinesis stream and posts them to an OpenSearch Cluster - I use `BulkIndexer` to send in a large number of documents more efficiently (i.e. via the `/bulk` API).

Currently, I build up most of my dependencies (e.g. DB connection pools, AWS Signer, etc) in the `main()` function of my Lambda, and then I pass these dependencies to a struct. I create a handler function which receives the struct containing all of these dependencies - this way, as long as my Lambda stays warm, the dependencies are readily available for the handler to use on a subsequent invocation. e.g.:

```go
package main

import (
	"context"
	"database/sql"
	"github.com/aws/aws-lambda-go/lambda"
)

//Dependencies can be reused across subsequent invocations of the Lambda - which speeds up non-cold starts (as we don't have to create the dependencies for injection)
type Dependencies struct {
	DbCon *sql.DB
	//... more dependencies
}

func main() {
	myDb := sql.OpenDB() //an imaginary DB that you still shouldn't call DROP on.

	d := Dependencies{
		DbCon: myDb,
	}

	lambda.Start(d.MyHandler)
}

func (d Dependencies) MyHandler(c context.Context) error {
	tx, err := d.DbCon.Begin()
	//Business logic here
}
```

I want to be able to pass the `BulkIndexer` on my dependency struct, but there's an issue. [Your documentation for the `BulkIndexer` says](https://github.com/opensearch-project/opensearch-go/blob/main/opensearchutil/bulk_indexer.go#L51):
> // You must call the Close() method after you're done adding items.

The problem is that [when you call `Close()` against `BulkIndexer`, you close `queue`](https://github.com/opensearch-project/opensearch-go/blob/main/opensearchutil/bulk_indexer.go#L283) (which is the channel that `Add()` appends requests to).
This creates a problem on subsequent invocations of the Lambda because if the channel is closed (by the call to `Close()` on a previous invocation), then calling `Add()` will result in trying to pass data to a nonexistent channel - which results in a panic.

This means that I can't store the `BulkIndexer` in my dependency struct because the channel underpinning the whole thing may be nonexistent on subsequent invocations. This means that I have to instantiate the `BulkIndexer` for each invocation of the Lambda (directly on the handler - even if the function is still warm).

### What solution would you like?
[...]

PR #865, opened 2026-06-09
fix(opensearchutil): avoid nil Error in BulkIndexer OnFailure callbacks

### LINKED ISSUE #679: [BUG] BulkIndexer Onfailure can give segmentation violation
### What is the bug?
Reading the bulkIndexer response BulkRespItem in OnFailure method can give a segmentation fault since the Error struct in the BulkRespItem is a pointer. 

The Error struct should be empty rather than a nil pointer.
_Operating system, version._

_If applicable, add screenshots to help explain your problem._

`// BulkRespItem represents an item of the BulkResp
type BulkRespItem struct {
	Index   string `json:"_index"`
	ID      string `json:"_id"`
	Version int    `json:"_version"`
	Type    string `json:"_type"` // Deprecated field
	Result  string `json:"result"`
	Shards  struct {
		Total      int `json:"total"`
		Successful int `json:"successful"`
		Failed     int `json:"failed"`
	} `json:"_shards"`
	SeqNo       int `json:"_seq_no"`
	PrimaryTerm int `json:"_primary_term"`
	Status      int `json:"status"`
	Error       *struct {
		Type   string `json:"type"`
		Reason string `json:"reason"`
		Cause  struct {
			Type        string    `json:"type"`
			Reason      string    `json:"reason"`
			ScriptStack *[]string `json:"script_stack,omitempty"`
			Script      *string   `json:"script,omitempty"`
			Lang        *string   `json:"lang,omitempty"`
			Position    *struct {
				Offset int `json:"offset"`
				Start  int `json:"start"`
				End    int `json:"end"`
			} `json:"position,omitempty"`
			Cause *struct {
				Type   string  `json:"type"`
				Reason *string `json:"reason"`
			} `json:"caused_by"`
		} `json:"caused_by,omitempty"`
	} `json:"error,omitempty"`
}`
Would it be better if the Error struct here were a struct instead of a pointer?

venkateshwaracholan:
## Summary

Fixes #679 by ensuring `BulkIndexer` `OnFailure` callbacks always receive a `BulkRespItem` with a non-nil `Error` field.

## Problem

In some failure scenarios (for example, status-only failures like HTTP 404 responses without an `error` object, or transport-level bulk failures), `OnFailure` can receive a `BulkRespItem` where `Error == nil`.

This can lead to nil pointer dereferences when callbacks access `Error.Type` or `Error.Reason`.

## Solution

Normalize the response item before invoking `OnFailure`. If `Error` is nil, attach an empty `BulkRespItemError` so callbacks can safely read error fields without additional nil checks.

This fixes the issue while preserving the existing `BulkRespItem.Error` API.

## Testing

Added tests covering:

* Failures with an `error` object
* Status-only failures without an `error` object
* Transport-level bulk failures
* Safe access to `Error.Type` and `Error.Reason` in `OnFailure`

Fixes #679.


sean-:
See also #336

venkateshwaracholan:
Thanks @sean- , I checked #336; makes sense as related follow-up work on BulkIndexer.

venkateshwaracholan:
The only remaining CI failures are integ-test-compat (true/false, latest, 3) in TestNodes/Stats looks like schema drift from new native_memory-related fields in OpenSearch latest, not related to this PR. All other checks pass. Happy to investigate further if useful.

sean-:
@venkateshwaracholan, thanks for looking at #336, too!

I went ahead and rebased the branch onto current `main` and squashed the merge-commit history into a single clean commit.  The change itself is unchanged: `bulkRespItemForOnFailure` normalizes `BulkRespItem.Error` to non-nil at all four `OnFailure` call sites.

Just waiting on CI to come back green (modulo the unrelated `latest` drift above), but this looks good to go from my side. Appreciate the fix and tests!

sean- (approved):

PR #867, opened 2026-06-09
chore: v4 to v5 migration

sean-:
Update `opensearch-go` to use `v5`-native APIs and deprecate v4 primitives.


Jakob3xD (commented):
### Code review

Reviewed on head `26e347c1`. **Scope:** this is a 2,644-file v4→v5 migration. The hand-written core — codegen engine (`cmd/osgen/*`), the partial-failure error model (`error.go`, `errmask/`), path encoding, and client/router wiring — was reviewed in depth and is clean. The ~518 generated files were reviewed via generator-correctness plus representative spot-checks (not line-by-line). Transport changes that overlap #859/#857/#822 were treated as already-verified. The module-path bump to `/v5` is consistent across live Go files and both `go.mod`s.

The findings below are all **non-blocking** and cluster on one theme: the import paths were bumped to v5, but the migration is not finished — the PR's own samples/guides and parts of the migration docs do not match the v5 API this PR ships.

#### 1. The migration's own samples and guides do not compile against the shipped v5 API

`_samples/*.go` (and the matching `guides/*.md`, `USER_GUIDE.md`) had their imports bumped to `/v5` but kept v4-era type/field/method names. Against the v5 `opensearchapi` this PR generates, those names no longer exist. `MIGRATING.md` prescribes exactly these renames — its own samples don't apply them. This does not fail CI because Go ignores directories with a leading underscore (`_samples/` is never compiled by `go build ./...`).

Representative (`_samples/search.go`):

https://github.com/opensearch-project/opensearch-go/blob/26e347c1d10892218813a15c152facee82e67d6a/_samples/search.go#L93-L131

- `SearchReq{Indices: …}` → v5 field is `Index []string` (`search_gen.go:42`)
- `Params: opensearchapi.SearchParams{…}` (value) → `SearchReq.Params` is `*SearchParams` (`search_gen.go:56`)
- `opensearchapi.SearchParams{Query: …}` → field is `Q string` (`search_gen.go:265`); no `Query`
- `opensearchapi.ToPointer(…)` → `ToPointer` is in package `opensearch` (`opensearch.go:645`), not `opensearchapi`
- `IndexReq{DocumentID: …}` → field is `ID`
- also across the guides: `Document*Req` → `*Req` (+ `client.Document.*` → `client.*`), `PointInTimeCreateReq` → `CreatePITReq`, `IndicesForcemergeReq` → `IndicesForceMergeReq`, `IndicesBlockReq` → `IndicesAddBlockReq`

#### 2. Migration docs are inaccurate / incomplete

- `UPGRADING.md` states `Response.Body` "changes from a public `io.ReadCloser` field to a `Body()` method" with `io.ReadAll(resp.Body())` — but `Response.Body` is still a field at HEAD; no `Body()` method exists. The documented pattern won't compile.

  https://github.com/opensearch-project/opensearch-go/blob/26e347c1d10892218813a15c152facee82e67d6a/UPGRADING.md#L140-L154

- `MIGRATING.md` omits the largest break: the `Document*Req` → `*Req` prefix drop and the `client.Document.*` → `client.*` method routing (also `PointInTimeCreateReq` → `CreatePITReq`, `IndicesForcemerge`/`IndicesBlock` renames).
- Several historical `UPGRADING.md` sections (the "v4" upgrade example, the AWS-SDK-v2 example) now show `/v5` import paths — a bulk find-replace that hit sections describing older releases.

#### 3. Default-router unset behavior: code and docs disagree

The doc comment says discovery is inherited "when `OPENSEARCH_GO_ROUTER` is truthy" (opt-in), but the code uses `!envvars.Falsy(envRouter)` (opt-out — on unless explicitly false), so an **unset** env var now enables on-start discovery (`/_nodes/http` calls to seed nodes). The `UPGRADING.md` default-router table row for "unset" reads "no auto-discovery", which contradicts the code. Worth confirming which is intended and aligning comment + table.

https://github.com/opensearch-project/opensearch-go/blob/26e347c1d10892218813a15c152facee82e67d6a/opensearch.go#L125-L133

https://github.com/opensearch-project/opensearch-go/blob/26e347c1d10892218813a15c152facee82e67d6a/opensearch.go#L366-L373

#### 4. Codegen de-stutter emits double-prefixed exported type names

The new acronym table (ISM/KNN/…) makes `schemaTypeName` compute prefix `ISM`/`KNN`, but `deStutterPrefix` compares against the still-old-cased local part (`strings.Index("IsmTemplate", "ISM") == -1`), so the stutter isn't removed — producing `ISMIsmTemplate` and `KNNKnnMethod`:

https://github.com/opensearch-project/opensearch-go/blob/26e347c1d10892218813a15c152facee82e67d6a/opensearchapi/types_gen.go#L8665

These compile, but they are unintended public names and a breaking rename from the v5preview surface (`IsmTemplate` / `KnnMethod`). Root cause is the case-sensitive match in `deStutterPrefix`:

https://github.com/opensearch-project/opensearch-go/blob/26e347c1d10892218813a15c152facee82e67d6a/cmd/osgen/naming.go#L289-L298

#### 5. `_samples/discovery_demo.go` imports `/v4`

The new sample imports `github.com/opensearch-project/opensearch-go/v4/opensearchtransport`. It carries `//go:build ignore` so it's outside the module graph, but it's a published example with the wrong major version.

https://github.com/opensearch-project/opensearch-go/blob/26e347c1d10892218813a15c152facee82e67d6a/_samples/discovery_demo.go#L22

#### 6. `RELEASING.md` tag examples use v4 versions against the /v5 path

Step 9 shows `go list -m github.com/opensearch-project/opensearch-go/v5@v4.3.0` — the path is `/v5` but the tag example is `v4.3.0`. Following it literally on the first v5 release would tag a v4 semver, invisible to `/v5` consumers.

https://github.com/opensearch-project/opensearch-go/blob/26e347c1d10892218813a15c152facee82e67d6a/RELEASING.md#L41-L63

opensearchapi/MIGRATING.md
```diff
@@ -0,0 +1,155 @@
...
+
+Concretely: a v4 caller who never set `Config.Errors` does not see partial failures as `error`. The same code on v5 does.
+
+If you need v4-shaped silence on v5, set `Errors: errmask.New(errmask.All)` explicitly. If you want to opt v4 in to v5-style surfacing, set `Errors: errmask.New()`.
+
+The `OPENSEARCH_GO_ERROR_MASK` env var overrides whatever `Config.Errors` resolves to. See [`README.md` Partial Failure Errors](README.md#partial-failure-errors) for the full guide.
+
+## Plugins
+
+Plugin APIs (k-NN, ML, Security, ISM, ...) live under [`plugins/`](plugins/README.md) as independent packages that share the same `opensearch.Client` transport.
```
Jakob3xD: Broken relative link: `[`plugins/`](plugins/README.md)` here is relative to `opensearchapi/`, so it resolves to `opensearchapi/plugins/README.md`, which does not exist — the file is at the repo top-level (`plugins/README.md`). Line 155 (`[`plugins/README.md`](plugins/README.md)`) has the same problem. Fix both to `../plugins/README.md`.

This is the actionable cause of the currently-red linkchecker check; the remaining 404s are `pkg.go.dev/.../v5`, which are expected pre-release.

opensearchapi/README.md
```diff
@@ -0,0 +1,360 @@
...
+
+A caller-supplied `DiscoverNodesOnStart` value always wins over the env-var-driven side-effect: setting `DiscoverNodesOnStart: &false` keeps auto-discovery off even when `OPENSEARCH_GO_ROUTER=true`.
+
+For routing semantics (role awareness, AIMD, shard-cost weighting) see [`../guides/routing.md`](../guides/routing.md). For node discovery see [`../guides/node_discovery_and_roles.md`](../guides/node_discovery_and_roles.md).
+
+## Plugins
+
+Plugin APIs (k-NN, ML, Security, ISM, etc.) live in separate top-level packages under [`plugins/`](../plugins/) (`github.com/opensearch-project/opensearch-go/v5/plugins/<name>`). They share the same `opensearch.Client` transport but have independent type hierarchies.
+
+See [plugins/README.md](plugins/README.md) for usage details and available plugins.
```
Jakob3xD: Broken relative link: `[plugins/README.md](plugins/README.md)` is relative to `opensearchapi/`, so it resolves to `opensearchapi/plugins/README.md` (404) — the file lives at the repo top-level. Fix to `../plugins/README.md`.

Jakob3xD (commented):
### Code review

Re-review at head `c994174` (the v4→v5 migration). The transport-correctness code is identical to PR #859 (already reviewed and approved) and was skipped; the generated library code is consistent (`check-gen` green) and the prior-round `ISMIsmTemplate` de-stutter is fixed by the generator (regeneration-stable). The shipped **library** is clean — the remaining items live in the **example/doc layer**, but one of them currently turns the linkchecker CI red (posted inline).

Found 3 issue(s) here (plus the inline linkchecker finding):

1. Migration incomplete — bundled example/doc code does not compile against the shipped v5 API. The v5 regeneration renamed/restructured the surface (`Indices` → `Index`, value → pointer `*Req`s, `ToPointer` / `NewFromClient` / `ParseTaskStatus` removed, sub-clients flattened onto `*Client`), but the non-compiled artifacts were not updated. CI does not catch this: Go ignores `_`-prefixed dirs, and doc snippets are never built. Empirically 9 of 11 `_samples/*.go` fail to build (`search.go`, `tasks.go`, `bulk.go`, `data_stream.go`, `index_lifecycle.go`, `advanced_index_actions.go`, `client_from_existing.go`, `document_lifecycle.go`), plus `USER_GUIDE.md` (~L199/L400 use `IndicesDeleteReq{Indices: …}` — now `Index`, and a pointer) and `guides/search.md`. Representative breakage (`SearchReq{Indices: …}` + value `Params`, both invalid in v5):

https://github.com/opensearch-project/opensearch-go/blob/c99417476e5e36cd5cd4a13e0e0cc9499285537a/_samples/search.go#L93-L111

2. "will be removed in v5" anachronism — this PR ships v5 with these symbols still present, so the deprecation horizon is self-contradictory (should read "a future major" / v6). CHANGELOG.md L16 ("v5 will … remove the deprecated `Perform`"), L206 and L208 ("will be removed in v5"):

https://github.com/opensearch-project/opensearch-go/blob/c99417476e5e36cd5cd4a13e0e0cc9499285537a/CHANGELOG.md#L205-L208

The same horizon appears in godoc — `Perform` and `ToPointer`:

https://github.com/opensearch-project/opensearch-go/blob/c99417476e5e36cd5cd4a13e0e0cc9499285537a/opensearch.go#L430-L433

https://github.com/opensearch-project/opensearch-go/blob/c99417476e5e36cd5cd4a13e0e0cc9499285537a/opensearch.go#L639-L641

3. `UPGRADING.md` — in the historical "Upgrading to >= 4.0.0 → Import Path" section, the heading says "Update all import paths from `v3` to `v4`" and the `// Before (v3)` block is correct, but the `// After (v4)` block shows `/v5` imports (a global `s/v4/v5` artifact):

https://github.com/opensearch-project/opensearch-go/blob/c99417476e5e36cd5cd4a13e0e0cc9499285537a/UPGRADING.md#L227-L231

Minor doc staleness (optional, non-blocking): `UPGRADING.md` TOC anchor still reads "Response.Body becomes a method" (the body text is already corrected); `RELEASING.md` step 6 still shows the example tag `v4.3.0`; `opensearchtransport/discovery_internal_test.go` L92/L330 comments still say `v4`.

Note on CI: fixing the plugins link addresses the actionable cause of the red linkchecker; the remaining 404s are `pkg.go.dev/.../v5`, expected pre-release.

opensearchapi/MIGRATING.md
sean-: Fixed in f19313c2. Both links in this file (this line and line 155) now point to `../plugins/README.md`, which resolves to the repo-root `plugins/README.md`. linkchecker passes on that change; the remaining `pkg.go.dev/.../v5` 404s are expected pre-release as you noted.

opensearchapi/README.md
sean-: Fixed in f19313c2 — changed to `../plugins/README.md`. Thanks for catching that the relative base is `opensearchapi/`, not the repo root.

sean-:
Thanks again for the re-review, @Jakob3xD. Every finding from both review rounds is addressed on the current head. Summary by finding:

**Inline linkchecker finding (MIGRATING.md / README.md `plugins/` links)**: fixed in f19313c2; both relative links now resolve to `../plugins/README.md`. linkchecker is green (remaining 404s are the expected pre-release `pkg.go.dev/.../v5`).

**#1 : samples/guides/docs didn't compile against the shipped v5 API**: migrated all 11 `_samples/*.go` to the v5 surface (verified each builds with `go build`), plus the `guides/*.md`, `USER_GUIDE.md`, and `opensearchapi/README.md` snippets. `_samples/discovery_demo.go` now imports `/v5` and handles the 2-value `NewDefaultRouter` return (also closes the old review-1 #5).

**#2: "will be removed in v5" anachronism**: replaced with "a future major version" on `Perform` and `ToPointer` (CHANGELOG + godoc), and dropped the stale `opensearchapi.ToPointer` reference (it only ever lived in package `opensearch`).

**#3: UPGRADING.md inaccuracies**: corrected the `Response.RawBody()` TOC anchor (body text was already right), and restored `/v4` import paths in the historical "Upgrading to >= 4.0.0" Import Path and AWS Signer sections (the bulk s/v4/v5 had leaked into them).

**Review-1 #3: default-router unset docs vs. code**: the code, the `opensearchapi.NewClient` godoc, the `opensearch.Config.DiscoverNodesOnStart` comment, the `opensearchtransport` package doc, and the UPGRADING.md truth table now all agree: router + on-start discovery are on unless `OPENSEARCH_GO_ROUTER` is explicitly false (`!Falsy`).

**Review-1 #4: codegen de-stutter (`ISMIsmTemplate` / `KNNKnnMethod`)**: confirmed fixed (0 occurrences in generated code); the central de-stutter pass in `cmd/osgen/naming.go` handles it.

**Review-1 #6: RELEASING.md v4 tag examples**: step 6 now shows `v5.0.0`; also corrected the stale `/v4` import-chain comments in `discovery_internal_test.go`.

**Additional change in this round (client surface):** while addressing the empty-sub-client observation, I found seven sub-client fields (`Document`, `PointInTime`, `Script`, `ComponentTemplate`, `IndexTemplate`, `Template`, `DataStream`) were declared but had zero operations routed to them. The generator now only emits a sub-client when an operation targets it. Document ops are grouped under `client.Doc.*`, point-in-time under `client.PIT.*`, and the indices sub-client's canonical field is `Index` (with `Indices`/`Indexes` aliases). Opt-in v4 compatibility forwarders (`--emit-v4-compat`, default on) keep `client.Bulk`/`MGet`/`Update`, `client.Document.*`, `client.PointInTime.*`, and `client.Indices.*` compiling. This is documented in `opensearchapi/MIGRATING.md` (which also closes review-1 #2's "MIGRATING omits the grouping break").

Jakob3xD (commented):
### Code review

Re-review at head `3af704f01` (force-pushed since my last review). All prior findings — `_samples` compiling against the v5 API, the migration/UPGRADING/RELEASING docs, the `ISMIsmTemplate`/`KNNKnnMethod` de-stutter codegen bug, and the broken `plugins/README.md` links — are resolved.

Found 4 issue(s), all posted as inline comments. The first three are the same underlying problem: the docs and changelog still advertise `client.Index` as a working backward-compatibility forwarder, but the final commit makes `Index` the indices sub-client field, so `client.Index(ctx, req)` no longer compiles.

Non-blocking observation (out of scope for this PR — pre-existing in `v5preview/`, generator unchanged here): optional numeric query params are generated as `int` with a `!= 0` guard (e.g. `IfSeqNo`/`IfPrimaryTerm` in `index_gen.go`/`delete_gen.go`/`update_gen.go`, `Size`/`From` in `search_gen.go`), so the legitimate value `0` can't be sent on the query-param path. The clearest impact is optimistic concurrency with `if_seq_no=0` on the first document in a shard. As this is a pre-existing generator behavior rather than something introduced by this PR, it may warrant a separate issue against `cmd/osgen` instead of blocking the migration.

opensearchapi/MIGRATING.md
```diff
@@ -0,0 +1,201 @@
...
+| `client.GetAllPits(...)`    | `client.PIT.GetAll(...)`    |
+| `client.DeleteAllPits(...)` | `client.PIT.DeleteAll(...)` |
+
+The request and response types are unchanged (`opensearchapi.IndexReq`, `opensearchapi.CreatePITReq`, and so on); only the canonical call path changes. The `Doc` sub-client is also reachable as `client.Document`, and `PIT` as `client.PointInTime`, for convenience.
+
+### Backward-compatibility forwarders
+
+To ease migration, the client ships forwarder methods so existing call sites keep compiling. These forward to the canonical sub-client method:
+
+- Top-level `client.Bulk`, `client.Index`, `client.MGet`, and `client.Update` forward to their `client.Doc.*` equivalents (these four were top-level methods historically).
```
Jakob3xD: `client.Index` is listed as a retained forwarder, but it no longer exists. The final commit makes `Index` a struct field on `Client` (type `indicesClient`, `clients_gen.go:33`), and a field and a method of the same name cannot coexist in Go — so `compatForwarders` in `cmd/osgen/route.go` has no `index` entry. `client.Index(ctx, req)` now fails to compile (`cannot call non-function client.Index`).

Suggest dropping `client.Index` here; only `client.Bulk`, `client.MGet`, and `client.Update` forward to `client.Doc.*`. The migration table at line 142 already correctly maps `client.Index(...)` to `client.Doc.Index(...)`.

opensearchapi/README.md
```diff
@@ -0,0 +1,357 @@
...
+| `client.Ingest`              | `client.Ingest.GetPipeline(ctx, nil)`           |
+| `client.Tasks`               | `client.Tasks.List(ctx, nil)`                   |
+| `client.Scroll`              | `client.Scroll.Get(ctx, req)`                   |
+| `client.SearchPipeline`      | `client.SearchPipeline.Get(ctx, nil)`           |
+| `client.Snapshot`            | `client.Snapshot.Get(ctx, req)`                 |
+| `client.Snapshot.Repository` | `client.Snapshot.Repository.Get(ctx, req)`      |
+
+Component-template, index-template, legacy-template, and data-stream operations live on `client.Cluster` and `client.Indices` (e.g. `client.Cluster.GetComponentTemplate`, `client.Indices.GetIndexTemplate`, `client.Indices.GetTemplate`, `client.Indices.GetDataStream`). Script operations (`client.GetScript`, `client.PutScript`) live directly on `client`.
+
+Top-level operations (Search, Reindex, DeleteByQuery, UpdateByQuery, etc.) live directly on `client`. Document operations are canonical on `client.Doc` (with `client.Bulk`, `client.Index`, `client.MGet`, and `client.Update` retained as backward-compatible forwarders); point-in-time operations are on `client.PIT`.
```
Jakob3xD: `client.Index` is listed as a retained backward-compatible forwarder, but it was dropped: `Index` is now the indices sub-client field on `Client` (`clients_gen.go:33`), so there is no `client.Index(...)` method and `client.Index(ctx, req)` does not compile.

Suggest listing only `client.Bulk`, `client.MGet`, and `client.Update` as the retained forwarders.

CHANGELOG.md
```diff
@@ -6,14 +6,15 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ### Added
 
+- Group document operations under a `client.Doc` sub-client and point-in-time operations under `client.PIT` (`Create`/`Delete`/`GetAll`/`DeleteAll`); `client.Document` and `client.PointInTime` remain as field aliases. `cmd/osgen` gains `--emit-v4-compat` (default true) to emit backward-compatibility forwarders so top-level `client.Bulk`/`Index`/`MGet`/`Update`, `client.Document.Source`, and `client.PointInTime.Get` keep working, and `--emit-v4-deprecation` (default false) to mark those forwarders deprecated
```
Jakob3xD: This states the top-level `client.Bulk`/`Index`/`MGet`/`Update` forwarders "keep working", but the `Index` forwarder was intentionally dropped — a later commit makes `Index` the indices sub-client field, which cannot coexist with a method of the same name. `client.Index(ctx, req)` does not compile.

Suggest removing `Index` from this list so the changelog matches the shipped API.

opensearchtransport/doc.go
```diff
@@ -200,21 +200,22 @@ logging globally without code changes. When enabled, debug output is written to
...
+Router nil; unset or any non-falsy value makes [New] create a [NewDefaultRouter].
+(In the v4 line the router is off unless OPENSEARCH_GO_ROUTER is set to true.)
+Programmatic Config.Router always takes precedence.
 
-The opensearch package observes the same variable independently: when
-OPENSEARCH_GO_ROUTER is truthy and opensearch.Config.DiscoverNodesOnStart has
-not been explicitly set, opensearch.NewClient enables on-start topology
+The opensearch package observes the same variable independently: unless
+OPENSEARCH_GO_ROUTER is explicitly false and opensearch.Config.DiscoverNodesOnStart
+has not been explicitly set, opensearch.NewClient enables on-start topology
```
Jakob3xD: This says on-start discovery is enabled "unless OPENSEARCH_GO_ROUTER is explicitly false and DiscoverNodesOnStart has not been set", but the actual guard in `opensearch.go` also requires `cfg.Router == nil`. When a caller passes a programmatic `Config.Router`, the env-var path is skipped and on-start discovery is not enabled.

Consider noting the `Router == nil` precondition here, as the parallel comment in `opensearch.go` does.

opensearchapi/MIGRATING.md
sean-: Fixed in 333b2b46. The forwarder list now reads `client.Bulk`, `client.MGet`, and `client.Update` only, and notes that `client.Index` is not a top-level forwarder because `Index` is the indices sub-client field -- callers should use `client.Doc.Index`. The migration table above already maps `client.Index(...)` -> `client.Doc.Index(...)`, so the two are now consistent.

opensearchapi/README.md
sean-: Fixed in 333b2b46 -- the retained-forwarders prose now lists only `client.Bulk`, `client.MGet`, and `client.Update`, and explicitly states `client.Index` is the indices sub-client field (use `client.Doc.Index`).

CHANGELOG.md
sean-: Fixed in 333b2b46. Dropped `Index` from the "keep working" forwarder list, and added the missing detail that the indices sub-client's canonical field is `client.Index` (with `client.Indices`/`client.Indexes` aliases) and that `client.Index` is intentionally not forwarded -- use `client.Doc.Index`. The changelog now matches the shipped API.

opensearchtransport/doc.go
sean-: Fixed in 333b2b46. The comment now states on-start discovery is enabled when `Config.Router` is nil, `OPENSEARCH_GO_ROUTER` is not explicitly false, and `DiscoverNodesOnStart` is unset -- and adds that a caller-supplied programmatic `Config.Router` skips the env-var path entirely, matching the guard in `opensearch.go`. (Also refreshed an adjacent paragraph that still described the v4 "default will flip in v5" behavior as a future event.)

sean-:
Thanks for the multi-round review, @Jakob3xD. Every finding from all three review rounds plus the inline comments is addressed on the current head. Verified each against HEAD before posting:

## Review round 1 (head `26e347c1`)
- **1. Samples/guides don't compile against v5**:  all 11 `_samples/*.go` now build and vet clean against the shipped v5 surface (`Indices`->`Index`, value->pointer `*Params`, `Query`->`Q`, `DocumentID`->`ID`, `opensearch.ToPointer`, `Document*Req`->`*Req`, `CreatePITReq`, `IndicesForceMergeReq`, `IndicesAddBlockReq`).  Guides/`USER_GUIDE.md` migrated to match.
- **2. Migration docs inaccurate/incomplete**: `UPGRADING.md` now documents the real `RawBody()` method (no phantom `Body()`); `MIGRATING.md` covers the `Document*Req`->`*Req` / `client.Document.*`->`client.*` routing and the PIT/forcemerge/block renames; the historical v3->v4 and AWS-SDK-v2 sections are restored to their correct `/v3` and `/v4` import paths.
- **3. Default-router unset behavior**: confirmed opt-out is intended. Code uses `!envvars.Falsy(envRouter)`; the `UPGRADING.md` table now reads "default Router injected, auto-discovery on" for the unset row, and the
`opensearchtransport/doc.go` comment matches.
- **4. `ISMIsmTemplate`/`KNNKnnMethod` de-stutter**: fixed in the generator (`deStutterPrefix` is now case-insensitive); regeneration-stable, no double-prefixed names remain.
- **5. `discovery_demo.go` imported `/v4`**: corrected to `/v5`.
- **6. `RELEASING.md` `v4.3.0` tag against `/v5`** — examples now use `v5.0.0`.

## Review round 2 (head `c994174`)
- **"removed in v5" anachronism**: unified to "before the first stable release" across `opensearchtransport.Client.Perform`, the `opensearch.Client.Perform` passthrough, `opensearch.ToPointer`, the `response_buffering.md` guide, and CHANGELOG. No "removed in v5/v6" remains in code or active docs.
- **`UPGRADING.md` historical section**: `// After (v4)` block restored to `/v4`; TOC anchor corrected; stale `/v4` comments in `discovery_internal_test.go` updated to `/v5`.

## Review round 3 (head `3af704f01`)
- **`client.Index` advertised as a working forwarder (3 inline: MIGRATING.md, README.md, CHANGELOG.md)**: all three now state `client.Index` is the indices sub-client field (a field and method of that name cannot coexist) and direct callers to `client.Doc.Index`. Retained forwarders are listed as `client.Bulk`/`client.MGet`/`client.Update` only. Confirmed `clients_gen.go` has `Index indicesClient` as a field and no top-level `Index` method.
- **`opensearchtransport/doc.go` missing `Router == nil` precondition (inline)**: added; the comment now states on-start discovery requires `Config.Router == nil`, `OPENSEARCH_GO_ROUTER` not explicitly false, and `DiscoverNodesOnStart` unset, and that a programmatic `Config.Router` skips the env-var path.
- **Non-blocking: optional numeric params can't send `0`**: addressed rather than deferred. `cmd/osgen` now promotes 0-meaningful integer query params to `*int` with a nil guard (mirroring the existing `*bool` treatment), scoped per operation: `if_seq_no`/`if_primary_term` on `delete`/`index`/`update` plus the plugin policy writes (`ism.put_policy`/`put_policies`, `rollups.put`, `sm.update_policy`, `transforms.put`), and `size` on `search`. `from` stays `int` (`from=0` == omit). `_create` is intentionally excluded (it rejects those params). Regenerated `opensearchapi/` and `plugins/`.

## Inline linkchecker finding
- `MIGRATING.md` / `README.md` `plugins/` links now resolve to `../plugins/README.md`. The remaining red `linkchecker` entries are the `pkg.go.dev/.../v5` 404s, which are expected pre-release as you noted.

## Additional hardening from this round
- Added a `build-samples` Makefile target and a CI job that compiles and vets every `_samples/*.go`, so example breakage (which Go hides because `_`-prefixed dirs are excluded from `go build ./...`) is caught going forward.
- `_samples/discovery_demo.go` now blocks on `transport.DiscoverNodes(ctx)` instead of `time.Sleep`.
- `opensearch.Response.String()` is non-consuming and keeps a value receiver, so both `Response` and `*Response` satisfy `fmt.Stringer`; it renders from the buffered body without draining `Body`.

## End state
All non-linkchecker CI is green: 3 Lint jobs, unit tests (ubuntu/macOS/windows), secure-cluster tests, all 20 `integ-test-compat` version combos (2.19.5 -> 3.7.0 + latest, secure & insecure), `check-gen`, DCO, codecov, and the new Build-samples job. The only red check is `linkchecker` (`pkg.go.dev/.../v5` 404s, expected pre-release).

Jakob3xD (commented):
### Code review

Re-review at head `0a3f83993`. All prior findings are resolved: the `client.Index` forwarder references (MIGRATING.md/README.md/CHANGELOG.md), the `doc.go` router-discovery precondition, and — nicely — the `if_seq_no`/`if_primary_term` `*int` fix (complete across all 7 affected ops, correct `!= nil` guard + deref). `_samples` are now built in CI, closing the long-standing gap where `_`-prefixed dirs were invisible to `go build`.

One new issue, posted as an inline comment: the reworked value-receiver `String()` drains `Body` for error/unbuffered responses, which silently breaks the documented `ParseError` flow.

response.go
```diff
@@ -75,30 +89,60 @@ func NewResponse(statusCode int, body io.ReadCloser, header http.Header) *Respon
...
 		Body:       body,
 		Header:     header,
+		render:     &renderCache{},
+	}
+}
+
+// String returns the response status and body as a string.
+//
+// String uses a value receiver, so both Response and *Response satisfy
+// fmt.Stringer, and rendering a response (e.g. for logging) does not drain its
```
Jakob3xD: This comment claims rendering "does not drain its Body", but that is now false for error (and any unbuffered) responses, and it breaks the documented error-handling flow.

`Client.Do` only buffers `rawBody` on the success path (`opensearch.go:532`, guarded by `!response.IsError()`), so for an error response `rawBody` is nil and `Body` is the single-use stream. `String()` is now a value receiver, so `renderedBody()` reads `Body` via `io.ReadAll` (line 141) and caches into the shared `render` — but, unlike the previous pointer-receiver `String()` which did `r.Body = io.NopCloser(...)`, it cannot restore the caller's `Body`. After a log call like `log.Printf("%s", resp)`, `resp.Body` is exhausted.

Any later `resp.Body` read then returns zero bytes. The canonical pattern breaks:
```go
if resp.IsError() {
    log.Printf("error: %s", resp)      // String() drains Body
    return opensearch.ParseError(resp) // io.ReadAll(resp.Body) -> empty -> ErrJSONUnmarshalBody, masking the real API error
}
```
This worked before this change (the old `String()` restored `Body`), and the test that pinned the Body-readable-after-`String()` invariant was dropped in this same delta.

Fix direction: also buffer error-response bodies into `rawBody` in `Do()` (so `String()` takes the `rawBody` fast-path and never touches `Body`, and `ParseError` reads an intact `Body`), or restore `Body` through the shared `render` pointer.

sean-: Good catch -- you're right, and thanks for tracing the full `log` -> `ParseError` flow. The value-receiver `String()` can't restore the caller's `Body`, so for an unbuffered error response it drained the single-use stream before `ParseError` could read it.

Fixed via your first suggested direction: `Client.Do` now buffers error-response bodies into `rawBody` (just like success responses), so `String()` takes the `rawBody` fast-path and never touches `Body`, and `ParseError` reads an intact payload. In the default buffered mode `Perform` already returned an in-memory `NopCloser`, so this only copies bytes already resident in memory -- no extra read.

Also in this change:
- Corrected the `String()` godoc -- dropped the absolute "does not drain its Body" claim and documented the precise invariant (non-draining for any `Do`-returned response; only an unbuffered/streamed or hand-built `Body` is consumed, since a value receiver can't restore the caller's field).
- Updated `RawBody()` godoc -- now populated for error responses too.
- Re-added the dropped invariant test as `TestResponseString_ErrorBodyNotDrained`, which reproduces the `log.Printf("%s", resp)` -> `ParseError` flow and asserts the real API error surfaces instead of `ErrJSONUnmarshalBody`.

Pushed to `357181c0`.

sean-:
Thanks for the re-review, @Jakob3xD. The one new finding -- the value-receiver `String()` draining error/unbuffered `Body` and breaking the `ParseError` flow -- is fixed (details on the inline thread): `Client.Do` now buffers error-response bodies into `rawBody`, so `String()` renders from `rawBody` without touching `Body`, and `ParseError` reads an intact payload. I also re-added the dropped invariant test (`TestResponseString_ErrorBodyNotDrained`) covering the exact `log.Printf("%s", resp)` -> `ParseError` sequence, and corrected the `String()`/`RawBody()` godoc.

Pushed to `357181c0`. With all prior findings resolved, I believe this should be good to go now -- the only red check is `linkchecker`, which is the expected pre-release `pkg.go.dev/.../v5` 404s you noted earlier.

response.go
Jakob3xD: Thanks — the error-response buffering fix looks right, and `TestResponseString_ErrorBodyNotDrained` pins exactly the `log` → `ParseError` flow.

One small thing remains in the godoc wording (not the error fix itself). The new comment states the invariant as non-draining for **any** `Do`-returned response, but that's still slightly broader than the code: a **success** response decoded with a nil `dataPointer` — e.g. `Do[T](ctx, c, m, req, nil)` or `client.Do(ctx, m, req, nil)` — hits neither buffering block. `opensearch.go:538` requires `dataPointer != nil`, and the new error block at `:559` requires `IsError()`. So `rawBody` stays nil and the value-receiver `String()` drains that response's `Body` — the same class as the error case you just fixed.

Two ways to square it:
- Buffer unconditionally in `Do`: since `Perform` already returns an in-memory `NopCloser`, dropping the `dataPointer != nil` / `IsError()` guards and always doing `rawBody = io.ReadAll(Body); Body = NopCloser(...)` makes the "any `Do` response is non-draining" claim literally true and lets the success / error / no-decode paths share one block.
- Or narrow the godoc to "any `Do` response whose body was buffered (a decoded success or an error)" and note that a nil-`dataPointer` success is not buffered.

Minor / non-blocking.

Separately: both this godoc and `RawBody()` reference a `DisableResponseBuffering` option, but I don't see that symbol anywhere in the tree — the unbuffered path appears to be the transport's `Stream` method (or a hand-built `Response`). Might be worth aligning the name to avoid pointing readers at a knob that doesn't exist.

Jakob3xD (approved):

PR #869, opened 2026-06-10
Backport transport correctness fixes to v4 (#859)

sean-:
## Description

Backports the transport-layer correctness fixes from #859 (targeting `main`) to the `v4` branch. No new behavior beyond what #859 introduced; this is the same 18-commit series replayed onto `v4`, with the `main`-only base commits (e.g. #822
AddressResolver, dependabot bumps, CI/backport workflow removal) deliberately excluded.

### Fixed

- **Per-request header override** no longer suppressed by global defaults: `setReqGlobalHeader` was comparing the per-request header *value* against the global header *name*, so a request-level header never overrode the matching global default and both were sent.
- **gzip buffer-pool nil poisoning**: `gzipCompressor.compress` returned `(nil, err)` while the deferred `collectBuffer` still ran, putting a typed-nil `*bytes.Buffer` into the `sync.Pool` that panicked on the next `Get().Reset()`.
- **Silently dropped body-read errors** in response buffering: `io.ReadAll` errors lost to `:=` shadowing now propagate wrapped in the new `opensearchtransport.ErrResponseBodyRead` sentinel; `Do` classifies the `(resp != nil, err != nil)` case via `errors.Is` so only genuine read failures are labeled `ErrReadBody`.
- **Unclosed error-response body** in `ParseError`: the original body is now closed before re-wrapping the read bytes in a `NopCloser`.
- **Response-body lifecycle on raw `RoundTrip` paths** (stats poller, discovery `/_cat/shards`, `/_cluster/state/metadata`, `/_nodes`, and the health pollers) now drain to EOF before close to preserve HTTP keep-alive. `Response.String()` is now non-consuming. AWS v1/v2 signers close the request body on the read-error path.

### Changed

- Split `Perform` into `Stream` + a buffering wrapper; drop `DisableResponseBuffering` in favor of the explicit `Stream(*http.Request) (*http.Response, error)` API (unbuffered body, caller owns close). `Perform` remains functional and is marked deprecated.
- `Response.String()` receiver changed from value to pointer so the body-restore mutation is observable; a non-pointer `Response` value no longer satisfies `fmt.Stringer` (pass `&resp`).

### Behavioral notes for reviewers

- `opensearch.Client.Do` now returns a non-nil `*Response` alongside a non-nil error on a body-read failure where it previously returned `(nil, err)`. Callers detecting a hard transport failure should check `resp == nil`, not `err != nil`.
- The remaining commits are supporting work: osgen regeneration, lint cleanup across all build-tag combinations, and benchmark fixes (goroutine leak + stateful body).


Jakob3xD (commented):
### Code review

Backport of #859's transport-correctness fixes onto `v4`. The backport itself is faithful and clean: all seven #859 fixes are present and correct, the v4-only `policy_*` router integration was adapted correctly (no dangling main-only `AddressResolver` references), the `/v4` module path is intact (no `/v5` leak across the ~1800 `v5preview/` files), and the CHANGELOG was genuinely v4-adapted rather than copied. The two red CI checks are not backport defects — `integ-test-compat` fails only against the `latest` nightly tag (all pinned versions 1.3.20–3.6.0 are green) and `codecov/project` is a coverage threshold.

Blocking concern (posted inline on `response.go`): a backport to v4 should not introduce breaking changes, and the `Response.String()` value→pointer receiver change is one — a `Response` value no longer satisfies `fmt.Stringer`. It ships onto a line whose version constant (`internal/version/version.go:30`) is `4.6.1`, and the CHANGELOG entry (line 202) documents it under `### Changed` without the `**BREAKING**` label its peers carry. Recommendation in the inline comment: drop the `String()` receiver change from the v4 backport (the non-consuming fix needs a pointer receiver and so can't be made non-breaking) and keep the other transport fixes.

Secondary (non-breaking, for awareness): `Client.Do` now returns a non-nil `*Response` alongside a non-nil error on some paths — callers should check `resp == nil` rather than `err != nil`. This is a documented behavioral refinement, not an API break, but v4 callers relying on the old `(nil, err)` shape should be aware.

Non-blocking (inherited from #859, file not modified by this PR): `guides/json.md` still steers callers to the now-deprecated `client.Perform` for raw/streaming instead of the new `Stream`; it will ship on v4 as-is.

response.go
```diff
@@ -78,11 +79,22 @@ func NewResponse(statusCode int, body io.ReadCloser, header http.Header) *Respon
...
 // String returns the response as a string.
-func (r Response) String() string {
+//
+// String is non-consuming: it reads the body to render it, then restores
+// Body with an in-memory reader over the same bytes so subsequent reads
+// (or a later String call) still see the full payload. This matters because
+// Response.Body is typically the buffered NopCloser produced by Client.Do;
+// without the restore, calling String for logging would silently empty the
+// body other code expects to read.
+func (r *Response) String() string {
```
Jakob3xD: This changes `String()` from a value receiver (`func (r Response) String()` in released v4.6.0) to a pointer receiver. On the v4 line that is a breaking change: a `Response` **value** no longer satisfies `fmt.Stringer`, so `fmt.Sprintf("%s", resp)` / `fmt.Print(resp)` on a value silently stops calling `String()` and prints the struct fields instead. `internal/version/version.go:30` targets `4.6.1` (a patch), and a backport to v4 should not introduce breaking changes.

The non-consuming `String()` fix inherently requires a pointer receiver — the body-restore (`r.Body = io.NopCloser(...)` on line 95) mutates the receiver — so it cannot be made non-breaking with a value receiver.

Recommended for the v4 backport: drop this `String()` receiver change and keep the value receiver (`String()` stays consuming on v4, as it already is in v4.6.0); defer the non-consuming fix to v5. The other six transport-correctness fixes don't depend on it. If the change must ship on v4, the release should be a minor/major (not `4.6.1`) and the CHANGELOG entry (line 202) labeled `**BREAKING**`, consistent with its peer entries (167/182/183/186/194/198).

Jakob3xD (approved):

sean-:
At this point in time I believe we can create a `4.6.1` tag and release.

PR #885, opened 2026-06-18
Rename `opensearchtransport.Client` to `opensearchtransport.Transport`

### LINKED ISSUE #853: v5 Transport Rename: `opensearchtransport.Client` -> `opensearchtransport.Transport`
## Summary

In the v5 stack we have three exported "client" types:

- `opensearchtransport.Client` - connection pool, retry, node selection/discovery
- `opensearch.Client` - `Perform` / `Do`, holds a transport
- `opensearchapi.Client` - typed API surface

Two are named `Client`, and the lowest-level "client" is not a client at all - it is the transport. Rename `opensearchtransport.Client` -> `opensearchtransport.Transport` so each layer's role reads correctly from its type name.

## Motivation

`opensearchtransport.Client` does no API work. It owns HTTP round-trip concerns: connection pool, retries, node selection, discovery, metrics. That is a transport. Calling it `Client` collides conceptually with the two genuine clients stacked above it and is a recurring source of "which client is this?" confusion. The v4 -> v5 boundary is the natural (and cheapest) place to make the rename.

## Proposed change

Rename the concrete type only; the constructor already returns it:

```go
// before
package opensearchtransport
type Client struct { /* ... */ }
func New(cfg Config) (*Client, error)

// after
package opensearchtransport
type Transport struct { /* ... */ }
func New(cfg Config) (*Transport, error)
```

## Blast radius

Small. Verified against the current v5 tree:

- **0** references to the concrete `opensearchtransport.Client` outside `*_test.go` / `*_gen.go`.
- 10 references total, all in tests: type assertions of the form `c.Transport.(*opensearchtransport.Client)` plus two test helpers.
- `opensearch.Client.Transport` is typed as `opensearchtransport.Interface`, not the concrete struct, so the interface-based wiring between the transport and the layers above it is untouched by the rename.

The internal test assertions and helpers update mechanically (`Client` -> `Transport`).

## Compatibility

v5 is a breaking boundary, so a hard rename is acceptable. For a soft landing, add a deprecated alias for one minor:

```go
// Deprecated: use Transport. Alias retained for v5.0 migration; removed later.
type Client = Transport
```

Because this is a type alias, existing `*opensearchtransport.Client` references keep compiling unchanged during the deprecation window.

## Upgrade path

Trivially scriptable. Within the package qualifier:

```sh
# BSD/macOS sed: use -i ''   |   GNU sed: use -i
grep -rl --include='*.go' 'opensearchtransport\.Client' . \
  | xargs sed -i '' -E 's/opensearchtransport\.Client/opensearchtransport.Transport/g'
```

For most users this matches nothing in their own code, since the concrete type is rarely referenced directly (the `Interface` is what flows through `opensearch.Client`). The alias above means even un-migrated code compiles.

## Non-goals

- **Not** renaming `New` -> `NewTransport`. `opensearchtransport.New` is already unambiguous within the package, and renaming it widens churn for no clarity gain.
- **Not** touching `opensearch*.Config` names. They only feed constructors, so there is no Client/Transport ambiguity there.
- **Not** moving the partial-failure error mask. It already lives on `opensearchapi.Client` (not on the transport or `opensearch.Client`), which is the correct layer; this proposal does not change that.

## End state

The transport layer's type name matches its job. Combined with the API-unification proposal, the library settles on exactly two names worth remembering: `opensearchtransport.Transport` and `opensearch.Client`.

sean-:
Resolves #853

## Summary

Renames the concrete `opensearchtransport.Client` type to `opensearchtransport.Transport`. The type owns HTTP transport concerns -- connection pooling, retries, node selection, and discovery -- not API concerns, so `Transport` reflects its role and stops it from colliding conceptually with the two genuine API clients above it (`opensearch.Client` and `opensearchapi.Client`).

This is a **breaking change**: the `Client` name is removed entirely, with no compatibility alias. A transitional `type Client = Transport` alias was considered but dropped because this lands in the `v5` line, where a clean break is appropriate and the migration is a mechanical find-and-replace.

The `New` constructor is intentionally **not** renamed. `opensearchtransport.New` is already unambiguous within the package, and `New` -> `NewTransport` would be churn without benefit.

## What changed

| Area | Change |
| --- | --- |
| Type | `opensearchtransport.Client` -> `opensearchtransport.Transport` (all methods, fields, and doc comments) |
| Constructor | `opensearchtransport.New` unchanged (still returns `*Transport`) |
| Generated code | `cmd/osgen` plugin template updated; all `plugins/*/client_gen.go` regenerated |
| Docs | `CHANGELOG.md`, `UPGRADING.md`, `guides/response_buffering.md` updated to reference `Transport` |

## Commits

This PR is structured as three reviewable commits:

1. **`refactor(transport)!: rename Client to Transport`** -- the rename itself (source, tests, docs, and the `osgen` template).
2. **`chore(plugins): regenerate client_gen.go for Transport rename`** -- mechanical `make regen` output, isolated so the generated churn is reviewable separately from the hand-written changes.
3. **`fix(response): buffer Do response body for every response`** -- a small follow-up cleanup to unify the `Do` body-buffering paths.

## Migration

Source-only, mechanical:

```go
// Before
var t *opensearchtransport.Client

// After
var t *opensearchtransport.Transport
```

`opensearchtransport.New(...)` callers are unaffected -- the constructor name and signature are unchanged.


opensearch.go
```diff
@@ -529,35 +529,28 @@ func (c *Client) Do(ctx context.Context, method string, req Request, dataPointer
...
-	if resp.Body != nil && response.IsError() {
-		// Buffer error-response bodies into rawBody so String renders via the
-		// rawBody fast-path (never draining Body) and a subsequent ParseError
-		// still reads an intact Body. Without this, a value-receiver String
-		// call (e.g. log.Printf("%s", resp)) would consume the single-use
-		// error Body and leave ParseError with an empty payload. In the
-		// default buffered mode Perform already returned an in-memory
-		// NopCloser, so this just copies bytes already resident in memory.
+	if resp.Body != nil {
+		// Buffer the response payload into rawBody for every Do response --
```
Jakob3xD, line 533: This bundled functional change buffers `Do()` for **every** response, including the nil-`dataPointer` success path that was previously left unbuffered. The behavior and `response.go` godoc are correct, but the existing (still unreleased) #859 CHANGELOG entry says `Do` "buffers error responses into `rawBody` like success responses" — which now understates the coverage (nil-`dataPointer` success was not buffered before this commit). Since it's the same release cycle, update that existing entry rather than adding a new one.

Separately, a heads-up on scope: this is a functional change living inside a PR titled as a pure rename. It's cleanly isolated in its own commit, so no objection — just flagging in case you'd prefer to split it out.

opensearchtransport/opensearchtransport.go
```diff
@@ -397,8 +397,8 @@ type Config struct {
 	AddressResolverRunner AddressResolverRunnerFunc
 }
 
-// Client represents the HTTP client.
-type Client struct {
+// Transport represents the HTTP client.
```
Jakob3xD: Lowest-confidence nit: the rename's rationale (per the CHANGELOG) is to stop the type name from colliding conceptually with `opensearch.Client`, yet this godoc still calls it "the HTTP client." Consider rewording to describe the role instead, e.g. "Transport manages HTTP round-trip concerns: connection pooling, retries, node selection, and discovery." Pre-existing prose and technically accurate, so take it or leave it — but this line is exactly where it'd be cheap to align.

Jakob3xD (commented):
### Code review

Clean, mechanically complete rename. `opensearch.Client`/`opensearchapi.Client` are correctly left untouched, every generated `client_gen.go` and the generator template are updated, and there are no compile-relevant stragglers. This PR also actively resolves the earlier #867-R5 doc concerns in `response.go` (the phantom `DisableResponseBuffering` reference is gone and the buffering coverage is now described accurately). No blockers.

Four findings, all docs/completeness — two posted inline. The two below have no specific added line:

1. `UPGRADING.md` has no dedicated migration section for this rename, even though it is marked `**BREAKING**` in the CHANGELOG. Every other v5 breaking change has its own subsection (`Partial Failure Errors`, `Default Router Injection in v5`, `DiscoverNodes() blocking semantics`, `opensearchtransport.Route` interface gained `OpID()`, `Response.RawBody()`). A short subsection with a before/after `*opensearchtransport.Client` -> `*opensearchtransport.Transport` snippet would keep the migration guide consistent.

https://github.com/opensearch-project/opensearch-go/blob/f8072b57369fbab7f13b368084e8e5cd77db6696/UPGRADING.md#L31-L39

2. The rename leaves stale `Client` references in comments of files the PR does not otherwise touch: `opensearchtransport/observer.go:18` ("must NOT call back into the Client") and `opensearchtransport/policy_index_router.go:373,380-381` ("when no Client is wired up", "without a full Client"). Comment-only, no compile impact — an optional completeness sweep so the codebase no longer refers to the old type name.

https://github.com/opensearch-project/opensearch-go/blob/f8072b57369fbab7f13b368084e8e5cd77db6696/opensearchtransport/observer.go#L16-L20

opensearchtransport/opensearchtransport.go
sean-: Reworded to describe the role rather than calling it "the HTTP client":

```go
// Transport manages HTTP round-trip concerns: connection pooling, retries,
// node selection, and discovery.
type Transport struct {
```

Aligns the godoc with the rename's rationale.

opensearch.go
sean-: Good catch. Updated the existing #859 entry rather than adding a new one, since it's the same unreleased cycle. Rewrote it in the affirmative and corrected the coverage so it no longer understates the change:

> Fix `Client.Do` to buffer every response body into `rawBody` -- decoded
> success, error, and no-decode (nil `dataPointer`) success alike (previously
> some paths, including the nil-`dataPointer` success path, were left
> unbuffered). ...

On the scope heads-up: agreed it's a functional change inside a rename-titled PR. It's isolated in its own commit (`fix(response): buffer Do response body for every response`) and is a small follow-up cleanup, so I've kept it here rather than splitting a one-commit change into a separate PR.

sean-:
Pushed an update addressing all four review findings:

- **`Transport` godoc** reworded to describe its role instead of "the HTTP client" (r3436427501)
- **#859 CHANGELOG entry** updated in place and corrected to reflect that `Do` now buffers every response body, including the nil-`dataPointer` success path (r3436427486)
- **`UPGRADING.md`** gained a dedicated rename migration subsection with a before/after snippet (review finding #1)
- **Stale `Client` comments** in `observer.go` and `policy_index_router.go` swept to `Transport` (review finding #2)

Changes are folded into the existing commits (rename-related into the rename commit, CHANGELOG wording into the response-buffering commit) to keep the three-commit structure, so the inline threads above will show as outdated after the force-push. `go build`, `go vet`, `gofmt`, and the unit suite pass.

Jakob3xD (approved):

PR #886, opened 2026-06-18
docs: reorganize guides, split UPGRADING.md, and add package documentation

sean-:
## Description

Reorganizes the project documentation for discoverability and removes duplicated content so each topic has a single source of truth. No code changes.

### Upgrade guides

- Split the 775-line `UPGRADING.md` into a version-history index plus per-major-version guides: `UPGRADING_V5.md`, `UPGRADING_V4.md`, `UPGRADING_V3.md`, `UPGRADING_V2.md`.
- Rename `opensearchapi/MIGRATING.md` to `opensearchapi/UPGRADING_V4_TO_V5.md` so the v4-to-v5 deep-dive sits in the same naming family.

### Guide and sample organization

- Group `guides/` and `_samples/` files by subsystem prefix so related docs sort together:
  - `transport-` connection pool, routing, discovery, health, metrics
  - `indexing-` documents, bulk, indices, templates, data streams
  - `usage-` search, raw JSON requests, tasks, error handling
  - `config-` environment variables, security, client construction
- Add `guides/README.md` as a topic index and a grouped sample catalog in `_samples/README.md`.

### Deduplication

- Make `guides/usage-error_handling.md` the single source for partial-error handling; `opensearchapi/README.md` keeps a short orientation and the Operation constants and links to the guide.
- Make `guides/transport-retry_backoff.md` the single source for resurrection-timeout configuration; `guides/transport-routing.md` links to it instead of duplicating the (drifted) config block, and gains the missing `ResurrectTimeoutFactorCutoff` reference row.

### Package documentation

- Add `doc.go` for `opensearchapi`, `plugins`, `signer`, and `signer/awsv2` (the last with a `Deprecated:` directive pointing to `signer/aws`).

### Tooling and incidental fixes

- The prettier markdown check ran `--check **/*.md` unquoted; with bash `globstar` off, root-level and deeply nested files were never linted. Quote the glob in the lint workflow and the `lint.markdown` Makefile target so every `.md` file is checked at any depth.
- Reformat `ADMINS.md`, `MAINTAINERS.md`, and `cmd/osgen/README.md` tables that the broken glob had been skipping, and fix a broken triage-issues link in `RELEASING.md`.


Jakob3xD (commented):
### Code review

Found 1 issue — posted as an inline comment.

The documentation reorganization itself is clean: spot-checked relative links and `#anchor` slugs after the renames/dedup resolve correctly, the four new `doc.go` package clauses and the `Deprecated:` convention are valid, and the prettier glob quoting fix is correct (it expands coverage rather than breaking it).

CHANGELOG.md
```diff
@@ -152,20 +152,21 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
...
+- Add `guides/config-envvars.md` as the canonical reference for every `OPENSEARCH_GO_*` environment variable — accepted values, defaults, parsing rules, and the exhaustive `OPENSEARCH_GO_ERROR_MASK` token list. Fix `OPENSEARCH_GO_ROUTER` default in `transport-routing.md` from incorrect `false` to correct `true`. ([#883](https://github.com/opensearch-project/opensearch-go/issues/883))
 
 ### Changed
 
+- Reorganize the documentation. Split `UPGRADING.md` into a version-history index plus per-major-version guides (`UPGRADING_V5.md` through `UPGRADING_V2.md`) and rename `opensearchapi/MIGRATING.md` to `opensearchapi/UPGRADING_V4_TO_V5.md`. Group the `guides/` and `_samples/` files by subsystem (`transport-`, `indexing-`, `usage-`, `config-`) and add a `guides/README.md` index. Make `guides/usage-error_handling.md` the single source for partial-error handling and `guides/transport-retry_backoff.md` the single source for resurrection-timeout config, replacing the duplicated copies in `opensearchapi/README.md` and `guides/transport-routing.md` with links. Add package documentation for `opensearchapi`, `plugins`, and `signer`. No code change
```
Jakob3xD: This entry says "No code change" and lists package docs for `opensearchapi`, `plugins`, and `signer`, but the PR adds four `doc.go` files — including `signer/awsv2/doc.go`, which is not in that list. That file also carries a `// Deprecated:` directive pointing to `signer/aws`, which is tooling-visible (staticcheck SA1019, `go vet`, gopls flag downstream `signer/awsv2` users). Consider adding `signer/awsv2` to the enumeration and dropping or qualifying "No code change," since shipping a deprecation marker is a source-level change with downstream effects.

sean-: Good catch on both points. Fixed:

- Added `signer/awsv2` to the `doc.go` enumeration in the Changed entry.
- Changed "No code change" to "No functional change," since you're right that a `Deprecated:` directive is a source-level, tooling-visible change.
- Added a dedicated `### Deprecated` entry for `signer/awsv2` (pointing to `signer/aws`), matching how the existing SA1019 deprecations in this changelog are recorded.

sean-:
Thanks for the review. Updated CHANGELOG and addressed in a follow-up: completed the enumeration, requalified the claim, and recorded the deprecation in the `### Deprecated` section.

CHANGELOG.md
```diff
@@ -152,20 +152,21 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
...
 
 ### Changed
 
+- Reorganize the documentation. Split `UPGRADING.md` into a version-history index plus per-major-version guides (`UPGRADING_V5.md` through `UPGRADING_V2.md`) and rename `opensearchapi/MIGRATING.md` to `opensearchapi/UPGRADING_V4_TO_V5.md`. Group the `guides/` and `_samples/` files by subsystem (`transport-`, `indexing-`, `usage-`, `config-`) and add a `guides/README.md` index. Make `guides/usage-error_handling.md` the single source for partial-error handling and `guides/transport-retry_backoff.md` the single source for resurrection-timeout config, replacing the duplicated copies in `opensearchapi/README.md` and `guides/transport-routing.md` with links. Add package documentation (`doc.go`) for `opensearchapi`, `plugins`, `signer`, and `signer/awsv2`. No functional change; the `signer/awsv2` `doc.go` carries a `Deprecated:` directive (see Deprecated below)
```
Jakob3xD: This PR now carries a real functional change: `aws.NewCredentialsCache(cfg.Credentials)` is newly added to `signer/awsv2/sdkv2signer.go` `NewSignerWithService` (absent at the base `bf544bc`). It eliminates a per-request `Credentials.Retrieve` — an STS call for assume-role/IRSA providers — on the signing hot path, which is exactly the kind of change this project records. So `No functional change` here is inaccurate, and the fix has no entry under `### Fixed`/`### Changed`. Suggest scoping this line to the doc reorganization and adding a dedicated Fixed/Changed entry for the `signer/awsv2` credential-caching fix, ideally noting the behavioral shift for any caller that passed a rotating raw credentials provider relying on per-request `Retrieve`.

Jakob3xD (commented):
### Code review

Found 3 issue(s) — all posted as inline comments.

signer/awsv2/doc.go
```diff
@@ -0,0 +1,19 @@
...
+// Modifications Copyright OpenSearch Contributors. See
+// GitHub history for details.
+
+/*
+Package awsv2 provides AWS Signature Version 4 request signing for OpenSearch using
+AWS SDK for Go v2. Construct a signer from an aws.Config with NewSigner (for Amazon
+OpenSearch Service) or NewSignerWithService, optionally passing SignerOptions to
+customize the underlying SigV4 signer.
+
+Deprecated: use the signer/aws package instead, which provides the same AWS SDK for Go
```
Jakob3xD: The deprecation rationale `enables credential caching by default` (mirrored in `CHANGELOG.md:214`) is undermined by this same PR: `signer/awsv2` now also wraps credentials in `aws.NewCredentialsCache` (`NewSigner` delegates to `NewSignerWithService`), so credential caching is no longer a distinguishing reason to migrate to `signer/aws`. Suggest rewording the rationale to a consolidation/maintenance motivation rather than caching, in both this godoc and the CHANGELOG Deprecated entry.

signer/awsv2/sdkv2signer.go
```diff
@@ -51,6 +51,16 @@ func NewSignerWithService(cfg aws.Config, service string, opts ...func(options *
 		return nil, errors.New("service cannot be empty")
 	}
 
+	// Wrap credentials in a cache. AWS SDK v2 does not cache credentials by
```
Jakob3xD: The claim `AWS SDK v2 does not cache credentials by default, so without this every signed request calls Credentials.Retrieve` overclaims for the documented `config.LoadDefaultConfig` path: `LoadDefaultConfig` already wraps the resolved provider in an `aws.CredentialsCache` (via `wrapWithCredentialsCache`, which is itself idempotent). The per-request STS concern only applies to a raw/manually-constructed credentials provider. The wrapping code itself is correct and harmless — for an already-cached provider the extra layer is a benign no-op. The same wording also appears in `signer/aws/aws.go:84` and `signer/doc.go`; consider scoping the claim to raw/manually-constructed credential providers.

CHANGELOG.md
sean-: Resolved, and the resolution went further than the original wording fix. Rather than keep two signer packages, the PR now removes `signer/aws` entirely and standardizes on `signer/awsv2` (see the `refactor(signer)!` commit). As a result:

- The `### Changed` doc-reorg line no longer claims "No functional change" — the functional change now has its own `**BREAKING**` entry recording the `signer/aws` removal.
- The credential-caching behavior lives in the surviving `signer/awsv2` constructor and is documented at the call site and in `UPGRADING_V5.md`.

Thanks for catching that the line was masking a real behavioral change.

signer/awsv2/doc.go
sean-: Good catch — the rationale was self-contradicting once this PR added caching to `awsv2` too. We took the opposite-and-simpler path: instead of deprecating `signer/awsv2`, we removed `signer/aws` and made `signer/awsv2` the single canonical signer. Its name matches AWS's own SDK-version nomenclature, and there's no longer a second package to deprecate-toward, so the contradictory "migrate for caching" rationale is gone entirely.

signer/awsv2/sdkv2signer.go
sean-: You're right on both counts: `config.LoadDefaultConfig` already wraps the provider in an `aws.CredentialsCache` (via `wrapWithCredentialsCache`, which is idempotent), and `aws.NewCredentialsCache` does not early-return, so unconditional wrapping would double-wrap that path.

Fixed in `signer/awsv2`:
- The comment is now scoped to a raw/manually-constructed `CredentialsProvider`, where the per-request `Retrieve` (and the STS hot-path concern) actually applies.
- The wrap is now guarded — `if _, ok := cfg.Credentials.(*aws.CredentialsCache); !ok` — so an already-cached provider (e.g. from `LoadDefaultConfig`) is left untouched rather than double-wrapped.

We default to wrapping rather than trusting the caller to have done it; a caller who wants different behavior can supply a pre-wrapped provider.

sean-:
Thanks. All three were valid items and have been addressed. They also surfaced a package deprecation reversal: the PR had `signer/awsv2` being deprecated *toward* `signer/aws` while both used AWS SDK v2 and this same PR added caching to both. That direction was backwards.

The branch now removes `signer/aws` and standardizes on `signer/awsv2` as the single signer, so there's one coherent path for AWS request signing. The credential-cache fix is scoped to raw providers and guarded against double-wrapping. Branch history was tidied (fixups squashed, messages reworded) so each commit reads correctly.

UPGRADING_V5.md
```diff
@@ -0,0 +1,156 @@
...
+
+For responses decoded by `Client.Do`, the buffered bytes are also available without consuming the body reader via the `RawBody() []byte` method (useful for inspection or comparison testing):
+
+```go
+raw := resp.RawBody() // nil for streamed or error responses; read resp.Body directly there
+```
+
+## `signer/aws` removed in favor of `signer/awsv2`
+
+The `signer/aws` package is removed. Use `signer/awsv2`, whose name mirrors AWS's own SDK-version nomenclature. Both packages signed with AWS SDK for Go v2 and share the same constructor shape, so the migration is an import-path change:
```
Jakob3xD: "the migration is an import-path change" (and "keep the same signatures" on line 156) understates the migration for the audience this guide serves — users on released v4. In released v4 (verified at tag v4.6.0) `signer/aws` is built on AWS SDK for Go **v1**: `NewSigner(opts session.Options) (*Signer, error)`. The v5 target `signer/awsv2` is SDK v2: `NewSigner(cfg aws.Config, ...) (signer.Signer, error)`. So a real v4 upgrader changes the SDK major, the constructor parameter type (`session.Options` -> `aws.Config` built via `config.LoadDefaultConfig`), and the return type (`*Signer` -> the `signer.Signer` interface) — not just the import path. Additionally, `signer/aws` exported `OpenSearchService`/`OpenSearchServerless` constants that `signer/awsv2` does not, so callers using `requestsigner.OpenSearchServerless` must switch to the `"aoss"` literal. Suggest rewording to describe this as a full SDK v1->v2 signer migration (the USER_GUIDE example already shows the `aws.Config` form).

CHANGELOG.md
```diff
@@ -198,10 +199,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
   - Enables proper context propagation for timeouts, cancellation, and graceful shutdown
   - Role compatibility validation prevents conflicting role assignments (master+cluster_manager, warm+search)
   - OpenSearch 3.0+ searchable snapshots now use `warm` role instead of deprecated `search` role
-- **BREAKING**: Migrate `signer/aws` package from AWS SDK v1 to AWS SDK v2 due to AWS SDK v1 reaching end-of-support on July 31, 2025
-  - Constructor now takes `aws.Config` instead of `session.Options`
-  - See USER_GUIDE.md for details required to migrate
-  - Users who need access to the existing `signer/awsv2` API can still use it, however they are encouraged to migrate to `signer/aws`
+- **BREAKING**: Remove the `signer/aws` package. Use `signer/awsv2`, whose name mirrors AWS's own SDK-version nomenclature; both signed with AWS SDK for Go v2 and share the same constructor shape (`NewSigner(cfg)` / `NewSignerWithService(cfg, service)`), so migration is an import-path change from `signer/aws` to `signer/awsv2`. See USER_GUIDE.md for an example.
```
Jakob3xD: Same as the UPGRADING_V5.md note: "migration is an import-path change from `signer/aws` to `signer/awsv2`" understates it for released-v4 users. Released v4 (v4.6.0) `signer/aws` uses AWS SDK v1 (`NewSigner(session.Options) (*Signer, error)`), so the real migration is an SDK v1->v2 change (new `aws.Config` parameter, interface return) plus dropping the removed `OpenSearchService`/`OpenSearchServerless` constants — not just an import swap.

signer/doc.go
```diff
@@ -0,0 +1,35 @@
...
+Service) or NewSignerWithService (for Amazon OpenSearch Serverless and other services):
+
+	cfg, err := config.LoadDefaultConfig(context.TODO())
+	if err != nil {
+	  log.Fatal(err)
+	}
+
+	signer, err := awsv2.NewSigner(cfg)
+
+The constructor caches the config's credentials. AWS SDK v2 does not cache a raw
```
Jakob3xD: This counterfactual is inaccurate for the `config.LoadDefaultConfig` example shown directly above. `LoadDefaultConfig` already returns credentials wrapped in an `*aws.CredentialsCache`, and the constructor's idempotent guard (`if _, ok := cfg.Credentials.(*aws.CredentialsCache); !ok { ... }` in sdkv2signer.go) skips wrapping in that case — so for this example "The constructor caches the config's credentials" is false, and "without this every signed request would call Credentials.Retrieve" also doesn't hold (the SDK's own cache already prevents that). The sibling comment in `signer/awsv2/sdkv2signer.go` documents the skip correctly; suggest aligning this package doc to the net-effect wording (e.g. "signing reuses resolved credentials rather than calling Retrieve per request; raw providers are wrapped in a cache, already-cached providers such as those from config.LoadDefaultConfig are left as-is").

Jakob3xD (commented):
### Code review

Re-review on the force-pushed head (`refactor(signer)!: remove signer/aws, standardize on signer/awsv2`). The prior round's findings on the awsv2 deprecation/credential-cache wording are all resolved here, and the removal is clean (no dangling imports, the two `doc.go` files consistently point at `signer/awsv2`). 3 issues posted as inline comments, plus one minor note below.

Minor (no specific line): the `signer/awsv2` credential-caching change (commit `8b9ac5f1`, adds `aws.NewCredentialsCache`) is not recorded in the Unreleased CHANGELOG. `signer/awsv2` shipped in v4.6.0 without caching, so this is an observable behavioral change to the now-canonical signer and is worth a `Changed`/`Fixed` line. Related: the PR description still says "No code changes" and references a `Deprecated:` directive on `signer/awsv2/doc.go`, both of which are now stale (the PR removes `signer/aws` and `awsv2/doc.go` now reads "the recommended signer").

UPGRADING_V5.md
sean-: Fixed. You're right that "import-path change" understated it for the audience this guide serves — released-v4 users on `signer/aws`, which is AWS SDK **v1** at v4.6.0. The section now describes it as a full SDK v1 -> v2 migration and spells out each change:

- constructor input `session.Options` -> `aws.Config` (via `config.LoadDefaultConfig`)
- return type `*Signer` -> the `signer.Signer` interface
- removed `OpenSearchService`/`OpenSearchServerless` constants -> the `"es"`/`"aoss"` literals
- the additional optional `SignerOptions`

The before/after code block shows both forms. Thanks for catching it.

CHANGELOG.md
sean-: Fixed alongside the UPGRADING_V5.md note. The CHANGELOG entry now describes the SDK v1 -> v2 change (the `session.Options` -> `aws.Config` parameter, the `signer.Signer` interface return, and dropping the removed `OpenSearchService`/`OpenSearchServerless` constants for the `"es"`/`"aoss"` literals) rather than calling it an import swap, and points at UPGRADING_V5.md and USER_GUIDE.md for the details.

signer/doc.go
sean-: Fixed. You're right that "the constructor caches the config's credentials" / "without this every request would call Retrieve" was false for the `config.LoadDefaultConfig` example shown directly above, since that provider is already an `*aws.CredentialsCache` and the constructor's guard skips it. The package doc now uses net-effect wording aligned with the `sdkv2signer.go` comment:

> The constructor ensures credentials are cached so signing reuses resolved credentials rather than calling `Credentials.Retrieve` on every request. A raw `CredentialsProvider` is wrapped in an `aws.CredentialsCache`; an already-cached provider (such as one from `config.LoadDefaultConfig`) is left as-is.

sean-:
Thanks for the re-review. Addressed the three inline points (the v4 -> v5 signer migration is now described as a full SDK v1 -> v2 change in both UPGRADING_V5.md and the CHANGELOG; the `signer/doc.go` cache wording is corrected to net-effect), and your minor note:

- Added a `### Fixed` CHANGELOG entry for the `signer/awsv2` credential-caching change, noting `signer/awsv2` shipped without it in v4.6.0.
- Refreshed the PR description — the stale "No code changes" line and the reference to the removed `Deprecated:` directive are gone; it now reflects the `signer/aws` removal and `awsv2` as the canonical signer.

Jakob3xD (approved):
### Code review

Approving. All findings from the prior review rounds are addressed on this head:

- The `signer/aws` -> `signer/awsv2` migration in UPGRADING_V5.md and CHANGELOG is now described as a full AWS SDK v1 -> v2 migration (constructor input `session.Options` -> `aws.Config`, return type -> the `signer.Signer` interface, removed `OpenSearchService`/`OpenSearchServerless` constants -> `"es"`/`"aoss"` literals). The Before/After snippet matches the released v4.6.0 API.
- `signer/doc.go` replaces the earlier counterfactual with accurate net-effect wording, including the `config.LoadDefaultConfig` already-cached skip.
- The `signer/awsv2` credential-caching change now has a dedicated `### Fixed` CHANGELOG entry that notes it shipped without caching in v4.6.0.

The `signer/aws` removal itself is clean: no dangling imports, build is green, and both package docs consistently point at `signer/awsv2` as the recommended signer.

PR #890, opened 2026-06-23
fix(osgen): json.RawMessage guard + collision/nullable/alias/enum fixes

ryanyuan:
### Description

• **Problem**: The generator emitted `json.RawMessage` for many response/request fields that should have been typed structs or enums. Causes:
(1) a pure `oneOf`/`anyOf` union whose parent-scoped Go name collided with its parent struct silently dropped the parent → raw;
(2) OpenAPI 3.1 `["null","<primitive>"]` scalars and bare-`$ref` alias responses fell through to raw;
(3) security `status` was typed as a `[number, string]` union (the server only ever returns a `RestStatus` name);
(4) `_sql/stats`/`_ppl/stats` declared a request body the server ignores, spawning a dead `SQLStats` raw type. There was also no guard to stop new raw uses from creeping in — and the initial guard missed raw leaves under nested wrappers (e.g. `[][]json.RawMessage`).

• **Fix**:
(1) re-key the colliding union by its referenced schema so the parent struct survives (narrow; non-colliding unions unchanged) + a non-fatal collision reporter;
(2) a checked-in `json.RawMessage` allowlist that fails generation on unlisted uses, detecting a raw leaf at **any** wrapper depth (pointer/slice/`map[string]`), reported by its outermost form;
(3) resolve nullable scalars and bare-`$ref` aliases to their typed forms;
(4) generate an int-backed `RestStatus` enum (closed-set `UnmarshalJSON`, unknown value → typed `*UnknownRestStatusError` recoverable via `errors.As`); shared enums keyed by `x-enum-name` panic on conflicting value sets; patch security `status` to use it;
(5) remove the phantom `_sql/stats`/`_ppl/stats` requestBody.

• **Impact**: `json.RawMessage` allowlist shrinks substantially. `tasks.list`/`tasks.cancel`/`delete_by_query_rethrottle`, `_common.mapping`, CAT `*Record`, ISM/`ml.search_*`, and security responses now expose typed structs/enums. `_samples/tasks.go` simplified to use the now-typed response. Tests are table-driven and mutation-verified, including `NewEnumTypesFile`, nested raw-form classification, and the enum unknown-value / conflict paths.

• **Caveat to call out explicitly**: The collision reporter surfaces **5 pre-existing collisions** in unrelated groups (`SearchProfile`, `FlowFrameworkCommonResp` ×3, `SecurityAnalyticsFindingsResp`) — long-standing, intentionally out of scope, reporter is non-fatal by design. `x-enum-name` is a **local-only vendor extension**; the `RestStatus` spec fix is upstream-pending for `opensearch-api-specification` (named `$ref` enum there). `MarshalJSON` errors on a zero-value (`Unknown`) `RestStatus` — unreachable today (response-only pointers w/ `omitempty`). `RestStatusOk` casing is intentional (matches existing `SecurityOk`).

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license. For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).


cmd/osgen/rawmessage_guard.go
```diff
@@ -0,0 +1,381 @@
...
+const (
+	rawBare  rawForm = iota // json.RawMessage
+	rawSlice                // []json.RawMessage
+	rawMap                  // map[string]json.RawMessage
+)
+
+func (f rawForm) String() string {
+	switch f {
+	case rawBare:
+		return "bare"
```
sean-: These magic strings should be `const`'s

cmd/osgen/resp_walk.go
```diff
@@ -151,6 +180,9 @@ func (w *walker) resolveInlineSchema(schema *openapi3.Schema, schemaKey, group s
 
 	// Primitive types.
 	if schema.Type.Is("string") {
```
sean-: I think these types have constants in one of the openapi packages

opensearch-openapi.yaml
```diff
@@ -70196,9 +70188,57 @@ components:
       type: object
       properties:
         status:
-          type:
-            - number
-            - string
+          type: string
```
sean-, line 70191: Let's get this pushed upstream to `opensearch-api-specification` next.

opensearch-openapi.yaml
```diff
@@ -70196,9 +70188,57 @@ components:
...
+          description: The HTTP status of the response, as a RestStatus enum name (e.g. "OK", "CREATED").
+          x-enum-name: RestStatus
+          enum:
+            - CONTINUE
```
sean-: I think this will fail a linter upstream in opensearch-api-specification because the linter will alpha-sort this, and we need to setup an ignore.  It would also be good to add the status code as a comment (e.g. `- OK # 200`) to anchor the enum values.

sean- (dismissed):

opensearch-openapi.yaml
ryanyuan: Done. Status codes and eslint ignore blocks have been added.

cmd/osgen/resp_walk.go
ryanyuan: Done. Swapped them with the openapi types.

cmd/osgen/rawmessage_guard.go
ryanyuan: Done. The hardcoded values have been extracted to const.

opensearch-openapi.yaml
ryanyuan: I will raise an upstream PR for this soon.

sean- (dismissed):

cmd/osgen/rawmessage_guard.go
```diff
@@ -0,0 +1,390 @@
...
+	return allowed, nil
+}
+
+// writeRawMessageAllowlist rewrites the allowlist file from uses, grouped by
+// group with sorted keys for minimal diffs. uses is assumed pre-sorted by
+// sortRawUses (collectRawMessageUses guarantees this).
+func writeRawMessageAllowlist(path string, uses []rawUse) (bool, error) {
+	var b strings.Builder
+	b.WriteString("# osgen json.RawMessage allowlist - DO NOT EDIT BY HAND.\n")
+	b.WriteString("# Regenerate with: make gen-api ARGS=-update-raw-message-allowlist\n")
```
Jakob3xD: This header tells contributors to run `make gen-api ARGS=-update-raw-message-allowlist`, but the `gen-api` Makefile target does not thread `$(ARGS)` into its `go run` invocation — the flag is silently dropped and the allowlist is never refreshed. The dedicated target added in this PR is `make gen-api-update-rawlist`. The same string is mirrored into the checked-in `cmd/osgen/rawmessage_allowlist.txt:2`, so fixing it here and regenerating updates both. Suggest pointing the header at `make gen-api-update-rawlist`.

cmd/osgen/ir/types.go
```diff
@@ -261,6 +273,7 @@ const (
 	TypeStruct    TypeKind = iota // plain struct with fields
 	TypeUnion                     // byte-prefix discriminated union (token class dispatch)
 	TypeLazyUnion                 // lazy-decode union (stores raw JSON, decodes on accessor)
+	TypeEnum                      // string-backed enum (named string type + const block)
```
Jakob3xD: This comment (and the related ones) describes the enum as "string-backed"/"named string type", but the generated type is int-backed (`type <Name> int` with `iota`) — `frag_enum.go` itself correctly says "int-backed (const iota)". The same inaccurate wording recurs at `resp_types.go:49` (the `IsEnum` field comment), `spec.go:51` (the `extEnumName` doc), and the `resolveStringEnum` doc in `resp_walk.go`. Suggest updating these to "int-backed (iota)" so the internal docs match the emitted code.

Jakob3xD (commented):
### Code review

Found 4 issue(s) — 2 posted as inline comments, 2 below. All low-severity; nothing blocking.

1. `x-enum-name` is only honored on the inline-string branch. `resolveStringEnum` is reached from `resolveInlineSchema`, but a string property typed via `$ref` to a component schema flows through `resolvePropertylessSchema` → `primitiveGoType`, which does not consult `x-enum-name`. A future spec that puts `x-enum-name` on a component-level string schema referenced by `$ref` would silently lose the enum marker and emit a plain `string`. Latent today (the current spec defines the status enum inline), so this is a low-priority note rather than a present bug.

https://github.com/opensearch-project/opensearch-go/blob/f1e01d01463b3f3c2f2083db51905b96810ce248/cmd/osgen/resp_walk.go#L760-L767

2. The five new CHANGELOG entries omit a `([#890](...))` PR link, which CONTRIBUTING.md asks for. Minor and pre-existing — roughly half of the current Unreleased entries also omit it — so flagging for consistency only.

https://github.com/opensearch-project/opensearch-go/blob/f1e01d01463b3f3c2f2083db51905b96810ce248/CHANGELOG.md#L7-L11

cmd/osgen/rawmessage_guard.go
sean-: Good catch: the `gen-api` target doesn't thread `$(ARGS)` into its `go run`, so that flag really was dead. Rather than point at the `gen-api-update-rawlist` target, I coupled the header to the generator flag itself, since the allowlist is a generator artifact and shouldn't assume a particular Makefile wrapper:

```go
b.WriteString("# Regenerate by re-running `cmd/osgen` with `-update-raw-message-allowlist`.\n")
```

The checked-in `rawmessage_allowlist.txt:2` picks up the new text on regeneration. Fixed in 75750d4a.

cmd/osgen/ir/types.go
sean-: Agreed.  The generated type is `type <Name> int` with `iota`, so "string-backed"/"named string type" was just stale wording. Corrected all four sites (`ir/types.go` `EnumMembers`/`EnumMember`/`TypeEnum`, `spec.go` `extEnumName`, `resp_types.go` `IsEnum`, and the `resolveStringEnum` doc in `resp_walk.go`) to "int-backed (iota)", matching what `frag_enum.go` already says. Fixed in 75750d4a.

sean-:
@ryanyuan Thank you for working this issue and pushing this up!

Thanks for the pass, @Jakob3xD.  All four addressed in 75750d4a (regeneration produces no `*_gen.go` changes; only the allowlist header text updates):

1. **`x-enum-name` only honored on the inline-string branch**: fixed. `resolvePropertylessSchema` now consults `resolveStringEnum` before the plain-primitive fallback, so a string schema carrying the marker is typed as an enum whether it arrives inline or via a component `$ref`. Latent today as you noted (the status enum is declared inline), but the guard's contract should hold for any future `$ref`'d enum schema. Added `TestWalkerStringEnumViaRef` to lock it in (verified it fails without the fix).

2. **CHANGELOG entries missing the `([#890](...))` link**: added to all five entries from this PR, matching the file's per-entry link convention.

3. **Allowlist header pointed at a flag the `gen-api` target drops**: fixed; header now references the `cmd/osgen` `-update-raw-message-allowlist` flag directly (see inline reply).

4. **"string-backed" enum docs**: corrected to "int-backed (iota)" across all four sites (see inline reply).

sean- (approved):

PR #896, opened 2026-06-26
feat(opensearchtransport): enable atomic per-request metrics by default

### LINKED ISSUE #891: Enable atomic per-request metrics by default (v4 and v5)
## Background

The transport's metrics subsystem is gated behind a single `EnableMetrics` flag (default off). That one flag covers two tiers of work with very different costs:

- **Per-request counters** -- `requests`/`failures` (`atomic.Int64`, `opensearchtransport/metrics.go:199`) and the responses-by-status tally. These are atomic-based and lock-free (the prior lock-free PR replaced the mutex-guarded `responses map[int]int`). The per-request cost is a handful of atomic adds.
- **Deep metrics** -- the callback-augmented snapshot: connection-state enumeration plus the per-connection / per-policy / per-snapshot callbacks (`connMetricCallbacks`, `policyCallbacks`, `snapshotCallbacks`). These run only when `Metrics()` is called (`metrics.go:330/344/365`), never on the request path, and are the genuinely expensive instrumentation.

Gating the cheap per-request counters behind the same flag as the expensive deep path means callers who want basic request/response counts pay for an opt-in they shouldn't need, and any internal consumer of those counters (see #C) cannot rely on them being populated.

## Proposal

Make the atomic per-request counters **always on** in both v4 and v5, and narrow `EnableMetrics` so it gates only the deep-metrics path.

- The per-request counters (`requests`, `failures`, responses-by-status) are populated unconditionally. They are lock-free atomics, so the per-request cost is negligible.
- `EnableMetrics` continues to gate the deep-metrics path (the callback-augmented `Metrics()` snapshot).
- `Metrics()` reports the always-on counters regardless of `EnableMetrics`; it adds the deep, callback-augmented fields only when `EnableMetrics` is set.

This is an **additive behavior change** in v4: counters that were previously gated are now always populated. That is intended and acceptable -- no caller loses data, and the per-request cost is atomic-only.

## Prerequisite

Confirm the lock-free counter work is present in the base branch. On the current branch `metrics.go:236-237` still shows the `m.mu.Lock()` -> `responses[status]++` -> `Unlock()` form. If the prior lock-free PR is not yet merged/rebased in, land it first so the per-request counters are genuinely lock-free before they are made unconditional.

## Scope of change

- `opensearchtransport/` -- populate per-request counters unconditionally; narrow `EnableMetrics` to gate only the deep-metrics path; `Metrics()` returns the always-on counters regardless of the flag.
- Tests: counters populate with `EnableMetrics` unset; deep-metrics fields appear only when set; concurrent request load updates counters correctly (no lock on the request path).
- `CHANGELOG.md` -- entry under Changed (per-request metrics now always on; `EnableMetrics` gates only deep metrics).

## Acceptance Criteria

- [ ] Per-request counters (`requests`, `failures`, responses-by-status) populate with `EnableMetrics` unset, in both v4 and v5.
- [ ] The per-request path takes no lock for metrics (atomics only).
- [ ] `EnableMetrics` gates only the deep-metrics (callback-augmented snapshot) path.
- [ ] `Metrics()` returns the always-on counters regardless of `EnableMetrics`; deep fields appear only when enabled.
- [ ] Lock-free counter work confirmed present in the base before counters are made unconditional.
- [ ] Full test suite (unit + integration) passes.

ryanyuan:
#### Problem

The transport's metrics subsystem is gated behind a single `EnableMetrics` flag (default off), which covers two tiers of work with very different costs:

- **Per-request counters** — `requests`, `failures`, and responses-by-status. Lock-free atomics; cost is a handful of atomic adds.
- **Detailed metrics** — the callback-augmented `Metrics()` snapshot: connection-state enumeration plus the per-connection / per-policy / per-snapshot callbacks. Genuinely expensive, but only runs when `Metrics()` is called.

Gating the cheap counters behind the same flag as the expensive detailed path means callers who want basic request/response counts pay for an opt-in they shouldn't need, and internal consumers cannot rely on those counters being populated.

#### Fix

- The per-request counters are now **always collected** via lock-free atomics, independent of `EnableMetrics`. The `metrics` struct is always allocated and wired into the connection pools.
- `EnableMetrics` now sets only a `detailed` flag that gates the expensive detailed-metrics path. Detailed-callback registration at all 7 policy sites is guarded by a nil-safe `metrics.detailedEnabled()` helper, so the counters-only path does zero per-snapshot work.
- `Metrics()` returns the per-request counters unconditionally and builds the detailed, callback-augmented fields only when `detailed` is set.
- The responses-by-status counter moved from a mutex-guarded `map[int]int` to a lock-free `atomic.Int64` array indexed by status code, with a single overflow bucket for out-of-range codes (committed as a separate `perf` prerequisite).

#### Impact

- **Additive behavior change:** counters previously gated are now always populated. No caller loses data; per-request cost stays atomic-only.
- **`Metrics()` no longer returns an error when metrics are disabled** — it always returns the per-request counters. Callers that branched on `if err != nil` for the disabled case should drop that check.
- A custom `ConnectionPoolFunc` returning a single-URL pool no longer errors during construction. Metrics wiring is now a no-op on a pool type mismatch (mirroring the multi-node branch), so custom pools that are neither built-in type construct cleanly.
- The public `Metrics.Responses` map keys out-of-range status codes under `-1` (`statusOverflow`); documented on the field.

### Issues Resolved
Closes [#891]


CHANGELOG.md
```diff
@@ -163,6 +163,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ### Changed
 
+- Per-request transport metrics (`requests`, `failures`, responses-by-status) are now always collected via lock-free atomics, independent of `EnableMetrics`. `EnableMetrics` now gates only the detailed-metrics snapshot (per-connection, per-policy, and router state returned by `Metrics()`). The responses-by-status counter moved from a mutex-guarded map to a lock-free atomic array. `Metrics()` no longer returns an error when metrics are disabled — it always returns the per-request counters (callers that branched on `if err != nil` for the disabled case should drop that check). ([#891](https://github.com/opensearch-project/opensearch-go/issues/891))
```
Jakob3xD: This entry documents a silent runtime contract change — `Metrics()` no longer returns the error that callers used to detect the disabled state — but it carries no `**BREAKING**` prefix. The runtime-only default changes directly below it (#167 router-on-by-default, #168 partial-failure-reported-by-default) and #188 (node-discovery behavior) are all marked `**BREAKING**` despite being non-compile changes, so this one is the same class. Two suggestions: (a) mark this entry `**BREAKING**` for consistency, and (b) add an UPGRADING_V5.md migration note (with a before/after for the dropped `if err != nil` check), consistent with other behavioral changes there such as "DiscoverNodes() blocking semantics". The entry text itself is clear — this is about discoverability of the break.

Jakob3xD (commented):
### Code review

Found 3 issue(s) — 1 inline, 2 below. No code bug: the atomic increments, the `statusCode-100` indexing over `[100, 600)`, the overflow bucket, and nil-safety are all correct. This is a documentation-completeness review around the silent runtime behavior change in `Metrics()`.

1. `Policies` and `Router` field godoc says "only populated when router/scoring is active", but this PR also gates these fields behind the detailed path — they stay nil when `EnableMetrics` is false even if the router/scoring is active. The PR's own new in-function comment in `Metrics()` (metrics.go:318-320) lists `Policies`/`Router` as staying zero/nil when the detailed path is off, which contradicts the field godoc. Suggest adding the `EnableMetrics`/detailed-path condition to both field comments.

https://github.com/opensearch-project/opensearch-go/blob/5dad97323341163258155b9ba8c7b31580f6fccc/opensearchtransport/metrics.go#L92-L98

2. This behavior change leaves two docs outside the PR's diff stale and worth updating here or in a follow-up: `policy.go:62` still says "Metrics (nil when EnableMetrics is false)" although `config.metrics` is now always non-nil (gating moved to `detailedEnabled()`); and `guides/transport-metrics.md:7,27` still tells users `Metrics()` returns `"transport metrics not enabled"` when disabled, which is no longer true for a `New()`-constructed transport.

https://github.com/opensearch-project/opensearch-go/blob/5dad97323341163258155b9ba8c7b31580f6fccc/opensearchtransport/policy.go#L61-L63

https://github.com/opensearch-project/opensearch-go/blob/5dad97323341163258155b9ba8c7b31580f6fccc/guides/transport-metrics.md#L6-L8

CHANGELOG.md
sean-: Agreed — it's the same class of runtime-only contract change as the router-on-by-default and partial-failure by-default entries right below it, so the inconsistency was a discoverability gap. Done in df7ed94a:

- (a) Marked the CHANGELOG entry **BREAKING**.
- (b) Added an `UPGRADING_V5.md` section ("Metrics error on disabled removed") with a before/after showing the dropped `if err != nil` disabled-state check, alongside the existing behavioral notes like "DiscoverNodes() blocking semantics". The CHANGELOG entry now links to it.

sean-:
Thanks @Jakob3xD.  The field godoc and the surrounding docs hadn't caught up with the two-tier split. Fixed in df7ed94a:

1. **`Policies` / `Router` field godoc** (`metrics.go`): reworded to state *both* preconditions — `EnableMetrics` set **and** an active router/scoring — so it no longer reads as if an active router alone populates them. They stay nil when `EnableMetrics` is false, matching the in-function comment in `Metrics()`.

2. **Stale out-of-diff docs:**
   - `policy.go` `policyConfig.metrics`: the "nil when EnableMetrics is false" comment is now wrong since the struct is always allocated. Updated to note it's always non-nil for a `New()`-constructed transport and that detailed callbacks are gated by `metrics.detailedEnabled()`, not nil-ness.
   - `guides/transport-metrics.md`: dropped the "`Metrics()` returns `transport metrics not enabled`" claim, documented the two-tier model (always-on counters vs. opt-in detailed snapshot), and corrected the error description (non-nil only on a detailed-snapshot callback failure).

Thanks for pushing this up, @ryanyuan!

sean- (approved):

Issue #892, opened 2026-06-24
Remove EnableMetrics in v5 and reimplement deep metrics with lock-free structures


sean-:
## Background

After #A, the per-request counters are always-on lock-free atomics, and `EnableMetrics` gates only the **deep-metrics** path: the callback-augmented `Metrics()` snapshot, which enumerates connection state and invokes the per-connection / per-policy / per-snapshot callbacks (`connMetricCallbacks`, `policyCallbacks`, `snapshotCallbacks`, `metrics.go:330/344/365`).

In v5 we can go further: remove the `EnableMetrics` flag entirely. With the per-request counters already unconditional, the only remaining reason for the flag is the cost of the deep path. If the deep path is reimplemented so it is cheap enough to leave on (or only does work when `Metrics()` is actually called), the flag has no purpose and becomes API surface to drop at the major-version boundary.

## Proposal (v5 only)

1. **Reimplement the deep-metrics path with lock-free structures** so it carries no per-request cost and no shared-lock contention. The deep snapshot should do work only when `Metrics()` is invoked, reading from lock-free state rather than lock-protected maps/slices.
2. **Remove `EnableMetrics` entirely** from the v5 config surface. All metrics (counters + deep snapshot) are always available; `Metrics()` always returns a full snapshot.
3. Update any v5 call sites and docs that reference `EnableMetrics`.

## Notes

- This is a breaking API change, appropriate only for v5 (the major version). v4 retains `EnableMetrics` (gating only the deep path, per #A).
- Benchmark the reimplemented deep path (single-threaded and highly concurrent) to confirm it is cheap enough to be unconditional. Do not assume; measure.

## Scope of change

- `opensearchtransport/` -- reimplement the deep-metrics snapshot with lock-free structures; remove `EnableMetrics` and its gating.
- `opensearch.go` / `opensearchapi/` -- remove `EnableMetrics` from the v5 config; update construction paths.
- Benchmarks for the deep-metrics snapshot path.
- Tests: `Metrics()` returns a full snapshot unconditionally; no `EnableMetrics` references remain.
- `CHANGELOG.md` -- entry under Changed/Removed (v5: `EnableMetrics` removed; deep metrics always available).

## Acceptance Criteria

- [ ] `EnableMetrics` is removed from the v5 config surface; no references remain.
- [ ] Deep-metrics snapshot uses lock-free structures and does work only at `Metrics()` call time.
[...]

PR #901, opened 2026-06-30
feat(opensearchtransport)!: remove EnableMetrics, reimplement deep metrics with lock-free structures

ryanyuan:
### Problem

`EnableMetrics` gated the *detailed* metrics path (per-connection enumeration, per-policy breakdowns, router-cache state) behind an opt-in flag. After #891, the per-request counters were already always-on lock-free atomics, so the flag only gated **call-time** work in `Metrics()` — leftover API surface to drop at the v5 major boundary.

The issue's mandate was *"do not assume; measure."* A mutex profile under concurrent `Perform()` load showed the real cost: `buildConnectionMetric` took a full `Connection.mu.Lock()` **just to read** two `time.Time` fields (`deadSince`, `overloadedAt`). That was the **number 1 explicit-lock contention site — 412 ms cumulative, 3.85% of total delay** — because the snapshot reader contended with the per-request `OnSuccess`/`OnFailure` writers.

### Improvement

- **Removed `EnableMetrics`** from the client and transport `Config`. Detailed metrics are now always collected; `Metrics()` always returns the full snapshot. **(BREAKING)**
- **Converted `deadSince`/`overloadedAt`** from `mu`-guarded `time.Time` to lock-free `atomic.Int64` (UnixNano, `0` = unset). Writes still occur under `c.mu` so the resurrection/standby read-modify-write decisions stay serialized; only the reads went lock-free, so `buildConnectionMetric` no longer takes each connection's mutex.

### Impact

Benchmarks: `go test ./opensearchtransport/ -run='^$' -bench='Metrics' -benchmem -count=64` (Apple M4 Max, `darwin/arm64`); medians of 64 runs. *Before* = reader lock reintroduced at the contention site; *after* = this PR. The delta isolates exactly the lock removal (the atomic migration is shared).

| Benchmark | Before (ns/op) | After (ns/op) | Δ |
|---|---|---|---|
| `BenchmarkMetrics` (single-threaded) | 2,126 | 2,064 | −2.9% |
| `BenchmarkMetricsParallel` (reader-vs-reader) | 1,994 | 1,956 | ~flat |
| **`BenchmarkMetricsUnderLoad`** (reader-vs-writer) | **13,390** | **12,448** | **−7.0%** |

- **Mutex contention at the `buildConnectionMetric` lock site: 412 ms → ~10.5 ms (97% reduction).**
- The single-threaded and reader-only paths are unchanged — the lock was never contended reader-vs-reader (`Parallel ≈ single-threaded` confirms this). The entire win is **reader-vs-writer**: snapshotting no longer blocks on the per-request writers.

### Issues Resolved
Closes [#892]


ryanyuan:
### Benchmarking

**Command:** `go test ./opensearchtransport/ -run='^$' -bench='Metrics' -benchmem -count=64`
**Environment:** Apple M4 Max (`darwin/arm64`), Go default GOMAXPROCS=14

**Methodology:** *Before* = per-connection `c.mu.Lock()` reintroduced at the
`buildConnectionMetric` read site (the original mutex-guarded path). *After* =
lock-free atomic reads (this PR). Only the reader-side lock differs between the
two runs; the atomic field migration is shared, so the delta isolates exactly
the lock removal. n = 64 runs per benchmark.

#### Medians (ns/op), n=64

| Benchmark | Before | After | Δ |
|---|---|---|---|
| `BenchmarkMetrics` (single-threaded) | 2,126 | 2,064 | −2.9% |
| `BenchmarkMetricsParallel` (reader-vs-reader) | 1,994 | 1,956 | ~flat |
| **`BenchmarkMetricsUnderLoad`** (reader-vs-writer) | **13,390** | **12,448** | **−7.0%** |

Min/max spread (ns/op):

| Benchmark | Before min–max | After min–max |
|---|---|---|
| `BenchmarkMetrics` | 2,048–2,164 | 1,988–2,096 |
| `BenchmarkMetricsParallel` | 1,877–2,172 | 1,798–2,174 |
| `BenchmarkMetricsUnderLoad` | 12,647–13,994 | 11,839–13,243 |

`BenchmarkMetricsUnderLoad` median allocs/op: 280 → 262. (Allocs here track the
concurrent `Perform()` traffic, not the lock change directly.) The two
un-contended benchmarks are 43 allocs/op in both runs.

#### Mutex-contention profile (primary evidence)

`go test -mutexprofile` under concurrent `Perform` load: the `c.mu.Lock()` in
`buildConnectionMetric` was the number 1 explicit-lock contention site —
**412 ms cumulative (3.85% of total delay)** before, **~10.5 ms (~0.1%)** after:
a **97% reduction**.

#### Takeaways

- The single-threaded and reader-only paths are essentially unchanged — the lock
  was never contended reader-vs-reader (`Parallel ≈ single-threaded` confirms it).
- The win is **reader-vs-writer**: `Metrics()` snapshotting no longer blocks on the
  per-request `OnSuccess`/`OnFailure` writers. `UnderLoad` ns/op improves and the
  mutex profile shows the contention essentially eliminated.
- `go test -race ./opensearchtransport/` stays clean with concurrent
  `Perform()` + `Metrics()`.

sean- (dismissed):

opensearchtransport/metrics.go
```diff
@@ -285,9 +271,11 @@ func (m *metrics) responsesSnapshot() map[int]int {
...
 }
 
-// Metrics returns the transport metrics. The detailed fields -- per-connection
-// enumeration, per-policy snapshots, and the router snapshot -- are populated
-// only when Config.EnableMetrics is set.
+// Metrics returns the transport metrics. It always returns the full snapshot:
+// the per-request counters (lock-free atomics on the hot path) plus the detailed
+// fields -- per-connection enumeration, per-policy snapshots, and the router
+// snapshot. The detailed fields are assembled lazily and lock-free at call time.
+// The returned error is non-nil only when a snapshot callback fails.
```
Jakob3xD: The godoc says the error is non-nil *only* when a snapshot callback fails, but the defensive `c.metrics == nil` branch just below (line 283) also returns an error — with the text `"transport metrics not enabled"`. That wording recycles the term for the `EnableMetrics` concept this PR removes, so it no longer maps to any real state. Suggest rewording line 283 to e.g. `"transport metrics not initialized"` (or `return Metrics{}, nil`); that also makes this `only when a snapshot callback fails` sentence accurate.

opensearchtransport/metrics_internal_test.go
```diff
@@ -211,23 +215,167 @@ func TestMetrics(t *testing.T) {
...
+					}
+				}
+			}()
+		}
+		wg.Wait()
+	})
+
+	t.Run("Metrics() surfaces callback errors via errors.Join", func(t *testing.T) {
+		// A failing callback is now the ONLY way Metrics() returns a non-nil
+		// error (the "not enabled" error was removed in #891). Each callback
```
Jakob3xD: This comment says the "not enabled" error was removed, but `metrics.go:283` still returns that exact string for the defensive nil-metrics path. The wording collides with the removed disabled-metrics concept, so this comment and the `Metrics()` godoc end up reading as contradictory. Disambiguating line 283 (e.g. `"transport metrics not initialized"`) resolves the collision.

Jakob3xD (commented):
### Code review

Found 2 issue(s) — all posted as inline comments. Both are non-blocking documentation/comment-accuracy nits; the lock-free timestamp migration itself looks correct (writes stay under `c.mu`, reads are lock-free).

sean-:
This should also be backported to v4, too.

sean- (approved):

PR #902, opened 2026-06-30
feat(opensearchtransport): add default-on client-side DNS caching

ryanyuan:
#### Problem

Transient resolver outages (e.g. a node-local DNS blip producing `dial tcp: lookup ...: i/o timeout`) fail requests even to nodes whose addresses were already resolved, because the default transport re-resolves every dial through the system resolver with no cache to fall back on. DNS-timeout errors are also not retried under the default config.

#### Fix

Install a process-local DNS cache (`github.com/rs/dnscache`) on the built-in transport, enabled by default:

1. Resolved addresses are cached and re-resolved on an interval (default `60s`, mirroring the TTL AWS publishes for managed OpenSearch Service endpoints).
2. When the resolver is briefly unreachable, the last-known-good address continues to be served until it recovers, so a transient DNS outage no longer fails requests to already-resolved hosts.
3. A node that genuinely moved self-corrects: the stale IP fails to dial, the connection is marked dead, and routing skips it.
4. Tunable/disable via the `DNSCacheRefresh` config field or `OPENSEARCH_GO_DNS_CACHE_REFRESH` (`0` = default, `<0` = disable, `>0` = explicit interval).
5. New `DNSLookups`, `DNSCacheMisses`, and `DNSLookupErrors` counters exposed via `Transport.Metrics()`, mirroring the existing `AddressResolver*` counters.

#### Impact

Caching is installed only when no custom `Transport` is supplied; a caller-provided `Transport` is never modified. The refresh goroutine is bound to the transport root context, so `Close()` reclaims it. Because Go's resolver does not expose record TTLs, the interval is a re-resolution cadence, not a per-record TTL.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license. For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).


opensearchtransport/dnscache.go
```diff
@@ -0,0 +1,145 @@
...
+	"github.com/rs/dnscache"
+
+	"github.com/opensearch-project/opensearch-go/v5/internal/envvars"
+)
+
+// defaultDNSCacheRefresh is the default DNS cache refresh interval. 60s mirrors
+// the TTL AWS publishes for managed OpenSearch Service domain endpoints (and the
+// JVM networkaddress.cache.ttl value AWS recommends to match it), so the client
+// re-resolves about as often as those records are expected to rotate.
+const defaultDNSCacheRefresh = 60 * time.Second
```
sean-, line 37: Can you thread this through the `Client`'s `Config` so we can twiddle this by more than just an envvar?

sean- (dismissed):

ryanyuan:
## Benchmark: the DNS cache is free when DNS is healthy, and decisive when it isn't

### Test configuration

```
go test ./opensearchtransport/ -run='^$' -bench='DNSResolve' -benchmem -count=64
```

- **Environment:** Apple M4 Max (`darwin/arm64`), Go default `GOMAXPROCS=14`
- **Iterations:** n = 64 per benchmark; results reported as medians
- **Compared:** `Cached` = this PR's resolver, primed (every measured call is a cache hit). `Uncached` = pre-PR behavior, resolving on every call. Resolver latency is injected via a controllable test resolver to model healthy → degraded → down.

### Resolution cost vs resolver latency (ns/op, n=64)

| Resolver state | Uncached (resolve every dial) | Cached (this PR) | Cache saves |
|---|---|---|---|
| **Healthy** (0 latency) | 3.3 ns | 21.7 ns | −18 ns *(cache slightly slower, both negligible)* |
| **Slow** (1 ms lookup) | 1.26 ms | 22.0 ns | **~57,000× faster** |
| **Degraded** (10 ms lookup) | 11.81 ms | 21.9 ns | **~539,000× faster** |

### Resolver outage (ns/op, n=64)

| Scenario | Result | Cost |
|---|---|---|
| **Cached** — serves last-known-good | dial proceeds ✅ | 21.9 ns |
| **Uncached** — nothing to fall back on | request fails ❌ | 5.62 ms, then error |

(Outage modeled as a resolver that times out before failing — the realistic failure mode; an unreachable resolver hangs to the deadline, it does not reject instantly. `outageDelay` stands in for that timeout, scaled down to keep the benchmark fast.)

### Min/max spread (ns/op, n=64)

| Benchmark | median | min – max |
|---|---|---|
| `DNSResolveStep/Cached/HealthyDNS` | 21.7 ns | 20.8 – 22.3 ns |
| `DNSResolveStep/Uncached/HealthyDNS` | 3.3 ns | 3.2 – 3.4 ns |
| `DNSResolveStep/Cached/SlowDNS_1ms` | 22.0 ns | 21.7 – 22.8 ns |
| `DNSResolveStep/Uncached/SlowDNS_1ms` | 1.26 ms | 1.14 – 1.28 ms |
| `DNSResolveStep/Cached/DegradedDNS_10ms` | 21.9 ns | 21.0 – 22.6 ns |
| `DNSResolveStep/Uncached/DegradedDNS_10ms` | 11.81 ms | 11.64 – 11.94 ms |
| `DNSResolveOutage/Cached_ServesStale` | 21.9 ns | 21.1 – 22.6 ns |
| `DNSResolveOutage/Uncached_FailsAfterTimeout` | 5.62 ms | 5.58 – 6.22 ms |

### Allocations

The cache adds exactly **one 16 B allocation per resolve** (the returned address slice) and nothing more — flat across every state. The uncached path allocates nothing but pays the full resolver round-trip on every dial.

### Conclusions

- **Healthy DNS: no meaningful cost.** A cache hit is ~22 ns — both paths are single- to double-digit nanoseconds against a request that takes microseconds-to-milliseconds end to end. The +18 ns is one map lookup and one slice alloc, lost in the noise. This is the "no difference in normal operation" property required to default the cache on.
- **The win is entirely conditional on the resolver, exactly as expected.** As soon as DNS is slow (1 ms) or degraded (10 ms), the cache turns a per-request resolver round-trip into a flat ~22 ns hit — orders of magnitude saved, and the gap widens linearly with resolver latency.
- **Under a full outage the cache is the difference between success and failure**, not just speed: serving the last-known-good address lets in-flight requests to already-resolved nodes keep working while the resolver is down, which is the transient-failure case this PR exists to fix.

opensearchtransport/opensearchtransport.go
```diff
@@ -166,6 +167,28 @@ type Config struct {
...
 
+	// DNSCacheRefresh controls how often the client-side DNS cache re-resolves
+	// cached hostnames. When the resolver is briefly unreachable, the
+	// last-known-good address continues to be served until the resolver recovers.
+	// The cache is only installed when no custom Transport is provided; a
+	// caller-supplied Transport is never modified. Because Go's resolver does not
+	// expose record TTLs, the interval is a re-resolution cadence rather than a
+	// per-record TTL.
+	// 0 = default (60s), <0 = disable caching, >0 = explicit interval.
+	DNSCacheRefresh time.Duration
```
Jakob3xD, line 178: These three DNS knobs exist only on `opensearchtransport.Config`. The top-level `opensearch.Config` has no DNS fields, and the mapping block in `opensearch.go` (the `opensearchtransport.New(opensearchtransport.Config{...})` call, ~lines 302–356) does not propagate them. So a user configuring via the documented path — `opensearch.NewClient(opensearch.Config{...})` — cannot set these programmatically; only the env vars work. `doc.go` is self-contradictory on this: line 31 says to configure via `opensearch.Config`, while lines 38–39 say "Configure the interval with DNSCacheRefresh". This looks like the config-threading requested in review ("thread this through the Client's Config so we can twiddle this by more than just an envvar") only half-completed. The fix lives in `opensearch.go`: add `DNSCacheRefresh`/`DNSDialTimeout`/`DNSKeepAlive` to `opensearch.Config` and pass them through the mapping block.

opensearchtransport/metrics.go
```diff
@@ -88,6 +88,12 @@ type Metrics struct {
 	AddressResolverRewrites int `json:"address_resolver_rewrites"` // URL actually changed
 	AddressResolverErrors   int `json:"address_resolver_errors"`   // Resolver returned an error
 
+	// DNS cache counters. Populated by the caching DialContext on the built-in
+	// transport; zero when a custom Transport is supplied or caching is disabled.
+	DNSLookups      int `json:"dns_lookups"`       // Dials that consulted the DNS cache
+	DNSCacheMisses  int `json:"dns_cache_misses"`  // Lookups not served from cache (cold or refreshed)
```
Jakob3xD: The `(cold or refreshed)` note appears inaccurate. A routine refresh re-stores the entry, so a continuously-used host stays a cache hit after the initial cold lookup — `dnsCacheMisses` would not increment on each refresh. This also contradicts the PR's own hook-site godoc in `dnscache.go` ("a lookup not served from cache", no "refreshed") and `TestCachingDialContextMetrics`, which only exercises the cold-miss path. Suggest e.g. "cold, or re-resolved after a ClearUnused eviction". The same wording is also on the internal struct (line 257) and in `guides/transport-metrics.md:80`.

opensearchtransport/dnscache.go
```diff
@@ -0,0 +1,208 @@
...
+		}
+		if len(ips) == 0 {
+			if m != nil {
+				m.dnsLookupErrors.Add(1)
+			}
+			return nil, &net.DNSError{Err: "no such host", Name: host, IsNotFound: true}
+		}
+
+		var conn net.Conn
+		for _, ip := range ips {
```
Jakob3xD: Optional / design note (not a bug). This custom dial loop gives each resolved IP the full `dialer.Timeout` (default 30s), so a dead or blackholed primary IP stalls for up to that timeout before the next IP is tried — whereas the stock `net.Dialer` splits the deadline across the resolved addresses (`partialDeadline`) and does dual-stack/Happy-Eyeballs. Secondary, conditional point: the cache freezes the A-record order for the refresh interval and the loop iterates without shuffling, so new connections to a multi-A host can pin to the first IP for ~60s, which works against the multi-IP AWS-endpoint motivation. If worth addressing, shuffling `ips` before the loop covers the ordering part.

Jakob3xD (commented):
### Code review

Found 5 issue(s) — 3 posted as inline comments, 2 below. None blocking; the goroutine lifecycle (`defer cancel()` + success-flag), `http.DefaultTransport.Clone()` (not mutated), the 0/<0/>0 config semantics + env precedence, and the new metrics atomics all check out.

1. The default-on behavior change is not surfaced in `UPGRADING_V5.md`. Enabling DNS caching by default changes runtime networking for every existing user: with serve-stale, a node whose IP changes (failover / blue-green) can keep receiving the last-known-good address for up to the refresh interval (~60s). That is the same class of default-on change that, in #896, was documented in `UPGRADING_V5.md` with a migration note and an explicit opt-out. Suggest adding one here, including the opt-out (`DNSCacheRefresh < 0` / `OPENSEARCH_GO_DNS_CACHE_REFRESH=-1`).

https://github.com/opensearch-project/opensearch-go/blob/0bc0a3e3094ce7f65651f3f06fa4a1de8da86fab/CHANGELOG.md#L8-L9

2. Maintainer awareness only (not blocking): the new dependency `github.com/rs/dnscache` is MIT (fine), but it is pinned to an untagged pseudo-version with its last upstream commit in 2023 and a single maintainer — and it now sits in the default code path of every client. Worth a conscious decision on taking it on (vs. vendoring the small amount of cache logic) given how sparingly this library otherwise adds dependencies.

https://github.com/opensearch-project/opensearch-go/blob/0bc0a3e3094ce7f65651f3f06fa4a1de8da86fab/go.mod#L9

opensearchtransport/opensearchtransport.go
sean-: Fixed. Threaded all the DNS knobs through `opensearch.Config` so they're settable via the documented `opensearch.NewClient(opensearch.Config{...})` path, not just env vars:

- Added `DNSCacheRefresh`, `DNSDialTimeout`, `DNSKeepAlive` (and a new `DNSTimeout`, see below) to `opensearch.Config`.
- Propagated each through the mapping block in `opensearch.go`.

`doc.go` is no longer self-contradictory: it points at `opensearch.Config` and the fields it names now actually live there.

opensearchtransport/metrics.go
sean-: Fixed, using your suggested phrasing. Confirmed against the library source: a routine refresh re-stores a used entry in place, so a continuously-used host stays a cache hit after the cold looku; `OnCacheMiss` only fires on a genuine cold lookup or after a `ClearUnused` eviction.

Updated to "Lookups not served from cache (cold, or re-resolved after a ClearUnused eviction)" on both `Metrics.DNSCacheMisses` and the internal atomic, and to "(cold, or re-resolved after eviction)" in `guides/transport-metrics.md`. The hook-site godoc in `dnscache.go` already read "a lookup not served from cache", so it was left.

opensearchtransport/dnscache.go
sean-: Took this past the optional note, since it surfaced a couple of real issues. Reworked the multi-address dial path:

- **Head-of-line blocking**: instead of a sequential loop giving each IP the full `dialer.Timeout`, a multi-A host now races up to `maxParallelDials` (3) addresses concurrently and takes the first to connect, cancelling and closing the losers. A dead/blackholed IP no longer stalls the others.
- **Ordering bias**: rather than shuffling `ips` in place, selection uses a random start offset with wraparound. `dnscache.LookupHost` returns the resolver's cached backing slice. `http.Transport` dials concurrently per pending request, so an in-place shuffle would race other dials and corrupt the shared cache entry. The random-offset read avoids mutating it. (The library only ever swaps the slice pointer under lock on refresh, never mutates elements, so a concurrent refresh can at most leave us racing a stale-but-valid address set, the intended serve-stale behavior.)
- **Per-attempt bound**: each attempt is bounded by `net.Dialer.Timeout`, which applies per `DialContext` call, so a slow address can't consume the whole budget.
- **Zero addresses**: a lookup that yields no addresses now returns an `ErrDNSNoAddresses` sentinel (wrapping the host) instead of a synthesized `net.DNSError`; all-fail returns an `errors.Join` of the per-address errors.

While in here I also added a `DNSTimeout` knob (config + `OPENSEARCH_GO_DNS_TIMEOUT`, default 10s) that sets `dnscache.Resolver.Timeout` so one hung resolution could stall an entire refresh tick.

I did not pursue the stock `net.Dialer` `partialDeadline`/Happy-Eyeballs deadline-splitting because the parallel race makes it moot for the dead-IP case, and full dual-stack ordering felt out of scope here.

sean-:
Thanks for the thorough pass. Status on the two non-inline items:

**1. UPGRADING_V5.md default-on note**: Added. New "Client-side DNS caching on by default" section documents the runtime networking change, calls out the serve-stale implication during failover/blue-green (last-known-good served for up to the refresh interval), and gives the opt-out both ways (`DNSCacheRefresh: -1` / `OPENSEARCH_GO_DNS_CACHE_REFRESH=-1`), plus the implicit opt-out of supplying a custom `Transport`. Modeled on the #896 metrics note.

**2. `github.com/rs/dnscache` dependency**: Acknowledged as a conscious maintainer decision rather than something this PR changes. The surface we depend on is small (cache + `RefreshWithOptions` serve-stale semantics), so vendoring the cache logic stays a viable fallback if the upstream staleness becomes a concern.

sean- (approved):

PR #906, opened 2026-07-02
perf(opensearchtransport): make connection dead/overloaded timestamps lock-free for v4

ryanyuan:
#### Problem

A mutex profile found `buildConnectionMetric` locking a connection mutex just to read two `time.Time` fields (`deadSince`, `overloadedAt`). It was the top contention site — **412 ms cumulative, 3.85% of total delay** — because every metrics snapshot serialized against the per-request `OnSuccess`/`OnFailure` writers.

This is the non-breaking perf subset of #901, backported to v4 per maintainer request. Unlike #901, this change **retains `EnableMetrics`** (same surface area and behavior); only the lock-free timestamp conversion is ported.

#### Improvement

- Converted `deadSince`/`overloadedAt` from mutex-guarded `time.Time` to lock-free `atomic.Int64` (UnixNano encoding, `0` = unset). Writes stay under `c.mu`; reads go lock-free.
- `buildConnectionMetric` now reads both timestamps without acquiring the connection mutex, eliminating the dominant contention against per-request writers.
- No public API change — `EnableMetrics` is unchanged.
- Also fixes an unrelated pre-existing race in `RolePolicy.DiscoveryUpdate`: `recalculateWarmupParams` is now called under the pool write lock.

#### Impact
Measured on Apple M4 Max, `BenchmarkMetricsUnderLoad` (8 concurrent writers + snapshot reader):
- Contention at the `buildConnectionMetric` lock site: **1.07 s → ~14.9 ms (~98.6% reduction)** — drops from the top app-level contention site to negligible.
- `BenchmarkMetricsUnderLoad`: **12.90 µs → 10.82 µs (−16.2%, p=0.000)**, allocs 262 → 232 (−11.4%).
- `BenchmarkMetricsParallel`: 1.99 µs → 1.89 µs (−4.9%).
- `BenchmarkMetrics` (single-threaded): 2.02 µs → 2.09 µs (+3.3%) — two atomic loads vs a direct read; immaterial and offset by the contended-path win.
- `go test -race ./...` clean.

### Issues Resolved
Relates to [#892] 


ryanyuan:
### Benchmark results (v4)

Re-profiled on this branch vs its parent (`upstream/v4`, `ca3c1c5b`) so the numbers are v4-specific, using the same methodology as #901. Hardware: **Apple M4 Max**. Benchmark added in `opensearchtransport/metrics_benchmark_test.go`.

#### Mutex contention

`BenchmarkMetricsUnderLoad` — 8 concurrent request writers + a `Metrics()` snapshot reader; `-mutexprofile`, `-benchtime=3s`.

| Site | Before (`upstream/v4`) | After (this branch) |
| --- | --- | --- |
| `buildConnectionMetric` `c.mu` lock | **1.07 s cumulative (7.33% of total mutex delay)** | **14.9 ms (0.10%)** — lock removed |

Contention at that site drops **~98.6%** (1.07 s → ~14.9 ms). It goes from the top application-level contention site to negligible; the remaining mutex delay is runtime-scheduler noise from the fake-transport goroutine churn, not this code.

#### Throughput / allocations

`benchstat`, `n=10`, `-benchtime=2s`:
| Benchmark | Before | After | Δ |
| --- | --- | --- | --- |
| `BenchmarkMetrics` (single-threaded) | 2.024 µs | 2.090 µs | +3.3% (p=0.000) |
| `BenchmarkMetricsParallel` | 1.990 µs | 1.892 µs | −4.9% (p=0.022) |
| `BenchmarkMetricsUnderLoad` | 12.90 µs | 10.82 µs | **−16.2% (p=0.000)** |
| `BenchmarkMetricsUnderLoad` allocs/op | 262.5 | 232.5 | −11.4% |

sean- (approved):

Issue #893, opened 2026-06-24
Add Client.Close() and cache default-constructed Clients to bound goroutine/memory resource leak


sean-:
## Background

In earlier releases the client was effectively a thin data structure: constructing one allocated no background resources, so discarding it without an explicit shutdown was safe. In v4 the client became stateful -- construction now starts background goroutines (node discovery, router pollers, transport workers) and a `cancelFunc` that must be invoked to release them. Existing caller code written against the thin client does not close anything, so any path that constructs and discards clients now leaks goroutines and the associated resources.

This change addresses that in two parts: a supported `Client.Close()` so callers can shut down explicitly, and a singleton cache that masks the new resource statefulness for the implicit default-construction path while callers migrate their code to close clients.

## Summary

Two related changes that address a goroutine/memory leak introduced in v4 and the client-shutdown ergonomics it exposes in v5:

1. **`Client.Close()`** on both the root `opensearch.Client` and the wrapper `opensearchapi.Client`, so callers can shut a client down without referencing transport internals.
2. **A process-wide singleton cache** for _implicitly_ default-constructed clients, keyed by a hash of the resolved construction config, with idle reaping governed by a new `OPENSEARCH_GO_DEFAULT_CLIENT_TTL` env var.

User-constructed clients (`NewClient(cfg)`) do **not** enter the cache. Only the implicit default-construction path (`NewDefaultClient`, and `NewBulkIndexer` with no client) is affected.

- **v4**: backport with a default TTL of `6m`.
- **v5**: same default TTL of `6m`. The cache stays armed; there is no reason to change its behavior across the major version. `Client.Close()` gives callers a supported way to shut down explicitly, and the cache plus TTL bounds the leak for code that does not.

## Problem 1: client shutdown requires referencing transport internals

There is currently no `Close()` on either client type. The root `Client.Transport` field is typed as the bare `opensearchtransport.Interface`, which only declares `Perform`:

```go
// opensearch.go:253
type Client struct {
    Transport opensearchtransport.Interface // only has Perform(...)
    config    *Config
}

// opensearchtransport/opensearchtransport.go
type Interface interface {
    Perform(*http.Request) (*http.Response, error)
}
```

`Close()` exists only on the concrete `*opensearchtransport.Transport` (`opensearchtransport.go:1160`). To release the transport's goroutines and pollers, a v5 caller must type-assert to the concrete type:

```go
[...]

PR #926, opened 2026-07-08
fix(client): Add `Close()` and cache implicitly-created default clients

ryanyuan:
### Description

Adds `Close()` to `opensearch.Client` and `opensearchapi.Client` so callers can explicitly tear down background goroutines (node discovery, health/stats pollers, DNS refresh) and idle connections. No transport type-assertion required.

Implicitly-created default clients now share one transport through a process-wide, refcounted, idle-TTL cache keyed by a hash of the config. This covers `opensearch.NewDefaultClient`, `opensearchapi.NewDefaultClient`, and the client `opensearchutil.NewBulkIndexer` builds when the caller supplies none. Each construction used to spin up its own goroutines and connection pool and leak them; now identical default configs resolve to the same transport. User-built `NewClient` clients never enter the cache. `NewBulkIndexer` closes the client it implicitly creates when the indexer is closed. Tune the idle eviction window with `OPENSEARCH_GO_DEFAULT_CLIENT_TTL` (default `16m`; `0` = never evict; a negative value disables caching).

The cache lives in a new transport-agnostic `internal/ttlcache` package. Config hashing uses a `KeyBuilder` with length-prefix framing so distinct configs cannot collide on one key.

### Issues Resolved

Closes [#893]


CHANGELOG.md
```diff
@@ -6,15 +6,16 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ### Added
 
+- Add `Close()` to `opensearch.Client` and `opensearchapi.Client` for explicit teardown of background goroutines (node discovery, health/stats pollers, DNS refresh) and idle connections, without type-asserting the transport. Cache implicitly-constructed default clients (`opensearch.NewDefaultClient`, `opensearchapi.NewDefaultClient`, and the client `opensearchutil.NewBulkIndexer` builds when none is supplied) in a process-wide, refcounted, idle-TTL cache keyed by config hash, so identical default clients share one transport instead of leaking one set of goroutines and its connection pool per construction. User-built `opensearch.NewClient`/`opensearchapi.NewClient` clients never enter the cache. `opensearchutil.NewBulkIndexer` now closes the client it implicitly creates when the indexer is closed. Tune the idle eviction window with `OPENSEARCH_GO_DEFAULT_CLIENT_TTL` (default `16m`; `0` = never evict; a negative value disables caching so every call builds a fresh client) ([#893](https://github.com/opensearch-project/opensearch-go/issues/893))
```
Jakob3xD, line 9: This entry sits under `### Added`, but caching implicitly-created default clients is a runtime behavior change: two `NewDefaultClient()` calls with identical config now share one transport, so their `Metrics()` counters are aggregated rather than per-client, and they share goroutine/connection lifecycle. Consider a `**BREAKING**` marker plus an `UPGRADING_V5.md` note (as done for the analogous default-behavior changes in #896), primarily so the shared-`Metrics()` observability change is discoverable for callers that build multiple default clients.

opensearch.go
```diff
@@ -423,6 +526,107 @@ func getAddressFromEnvironment() []string {
...
+	return nil
+}
+
+// configKey returns a stable cache key for cfg's hashable fields and true, or
+// (0, false) when cfg carries a field that cannot be compared by value (funcs,
+// interfaces, a context, or a custom transport); such a client is never cached.
+// The field selection lives here because it reads Config; the hashing primitive
+// is [ttlcache.KeyBuilder]. It must stay in sync with Config;
+// TestConfigKey_FieldGuard fails loudly when Config grows a field.
+func configKey(cfg Config) (ttlcache.Key, bool) {
```
Jakob3xD, line 560: `configKey` hashes raw, pre-normalization values, whereas `NewClient` normalizes before building the transport: addresses get a trailing `/` trimmed, in-URL `user:pass@` credentials are folded into Username/Password, and the scheme is lowercased. Header *values* are also sorted, which hides multi-value ordering. This is latent today — only empty `Config{}` from `NewDefaultClient()` reaches this path, so there is no false cache hit and no credential bleed. But if the cache is ever extended to caller-supplied configs, these quirks would silently duplicate or over-merge transports. Worth either mirroring `NewClient`'s normalization here, or a short doc comment noting the key is derived from raw pre-normalized values and only `Config{}` is cached.

internal/ttlcache/ttlcache.go
```diff
@@ -0,0 +1,286 @@
...
+		return zero, nil, err
+	}
+
+	// Lock-free hit path.
+	if v, ok := c.cache.Load(key); ok {
+		e := v.(*Value[T])
+		if e.incIfLive() {
+			return e.Obj, releaseFn(e), nil
+		}
+		// Claimed for eviction; fall through to the locked slow path, which
```
Jakob3xD: This comment understates the slow path. After a failed `incIfLive()` the code first calls `item.New(ctx)` (a potentially network-bound construction) *before* taking `c.mu`, and the locked path does not only block until the sweep removes the entry — it also reacquires an entry another goroutine may have re-inserted under the same key and discards the freshly built value. Consider noting the pre-lock construction and the reacquire-and-discard case so the comment matches the code.

internal/ttlcache/straykey.go
```diff
@@ -0,0 +1,27 @@
...
+//
+// The OpenSearch Contributors require contributions made to
+// this file be licensed under the Apache-2.0 license or a
+// compatible open source license.
+
+package ttlcache
+
+import (
+	"fmt"
+	"testing"
```
Jakob3xD: This pulls the stdlib `testing` package into a shipped production file (`internal/ttlcache`) to gate panic-vs-log via `testing.Testing()`. It is legitimate (Go 1.21+, no flag pollution) but unusual for this repo, where `testing` is otherwise confined to `_test.go` files. Is importing `testing` into the shipped package intentional? If so, no change needed — flagging only so it is a conscious choice.

Jakob3xD (commented):
### Code review

Found 5 issue(s) — 4 posted as inline comments, 1 below:

1. The new env var `OPENSEARCH_GO_DEFAULT_CLIENT_TTL` is documented in the CHANGELOG and in the `internal/envvars` godoc, but is missing from `guides/config-envvars.md`, which states it is the canonical reference for every `OPENSEARCH_GO_*` runtime variable. The four DNS vars from #902 each have both a quick-reference table row and a detail entry there; this var should be added analogously (table row + detail section).

https://github.com/opensearch-project/opensearch-go/blob/39b70151d18c3a5be9fffcb5b8b58bcefccde217/guides/config-envvars.md#L13-L20

internal/ttlcache/straykey.go
ryanyuan: Good catch. 

Switching testing.Testing() to a package var, `panicOnInvariantViolation`, which is false in a shipped binary and set true by an init in `straykey_internal_test.go`. Only ttlcache internals can reach the stray-key path, so flipping it in ttlcache's own test binary covers everything that can trigger it, and `TestOnStrayKey_PanicsUnderTest` still passes.

internal/ttlcache/ttlcache.go
ryanyuan: Updated the comment with the pre-lock `item.New(ctx)` construction and the reacquire-and-discard case

opensearch.go
ryanyuan: Updated the docstring to cover those cases.

CHANGELOG.md
ryanyuan: Updated.

sean- (approved):

Issue #464, opened 2024-01-24
[FEATURE] BulkIndexer - ensure that subsequent requests on the same documentID are consistently sent to the same worker


mikesmitharoo:
### Is your feature request related to a problem?

The BulkIndexer allows the configuration of multiple workers via `NumWorkers`. Which `fans out` bulk requests to N go routines which manage bulk buffers and API requests separately. 

There is no allocation logic for the `fan out` so there is no guarantee that multiple actions to the same document will exist in the same request. And more importantly it could break ordering as subsequent actions may be flushed before previous ones depending on the state of the what buffer is for that particular worker.

This can cause issues with systems processing a lot of messages for the same items such as streams from kafka where many item updates could occur for the same item. It makes it possible to trigger document not found exceptions in scenarios where 1 message creates and the next message is an update, because the update worker flushes but the create message is still on the buffer of another worker. 

If document IDs can be guaranteed to be on the same worker they will either be in the same bulk request or at least in the correct order of bulk requests.

### What solution would you like?

- Replace the existing internal queue channel with an array of channels up to the `NumWorkers`. 
- Create a numerical hash of the `DocumentID`.
- Use that numerical hash to determine what `queue` the request should be added to on the `Add` function.

### What alternatives have you considered?

- Consuming code could handle this by setting `NumWorkers` to 1, but this remove the horizontal scaling of this library and can cause throughput issues.
- Additionally consuming code could create `N` `BulkIndexer` instances all with 1 `NumWorkers` and handle this hashing process there to add messages to different bulk indexers based on the hash. But it would be better if this was wrapped up in this library as it's generic.

### Do you have any additional context?

N/A

dblock:
This is a good idea. Want to try to implement this @mikesmitharoo?

Catch All Triage - [1](https://github.com/dblock) [2](https://github.com/krisfreedain) [3](https://github.com/gkamat) [4](https://github.com/andrross) [5](https://github.com/getsaurabh02)

yanaix10:
hey @sean-  
If the issue has been reviewed can I have a look on to ressolve it ?
If you could please assign this to me.

sean-:
@naitikdevx10 : feel free to run with this.  Be sure to look at https://github.com/opensearch-project/opensearch-go/blob/main/guides/transport-routing.md#document-path-docrouter

Ashwinnbr007:
Is anyone looking into this issue.. Can i take a peek at this if no one is actively looking into this? 

cc @sean-

PR #930, opened 2026-07-08
feat(opensearchutil): Add Flush(context.Context) to BulkIndexer

LeeFred3042U:
### Description

Adds `Flush(context.Context) error` to the `BulkIndexer` interface, so callers can drain every submitted item without closing the indexer.

`Close` was the only way to force a drain, and it closes each worker queue, so anything that needs to drain on demand had to build a fresh indexer per unit of work: a warm Lambda handler holding its dependencies across invocations, or a test asserting on indexed documents. Neither threshold substitutes, as the issue notes: `FlushInterval` means waiting on a timer, `FlushBytes` means shrinking batches until `/bulk` stops paying off.

**How it works.** `Flush` sends a barrier down each worker's queue and waits for that worker to report back. A queue is FIFO, so by the time a worker takes the barrier it has already buffered every item queued ahead of it, which is what makes the drain a barrier rather than a sample. The barrier carries the caller's context, so a deadline passed to `Flush` bounds the bulk requests it drives.

**What changed since the first revision of this PR.** The original implementation replaced the channel queue with a lock-free ring buffer and a single sequencer goroutine. In the meantime #950 landed per-document routing on main in the same code: each worker owns its item channel, and `Add` picks one by hashing `DocumentID`, so every action on a document reaches one worker in submission order (#464). A single sequencer cannot offer that guarantee and serializes work the per-worker channels do in parallel, so this branch was brought up to main and `Flush` reimplemented on top of the channel design. `QueueSize` went with the ring buffer, so no new configuration knob is added.

### Semantics

- Items added concurrently with `Flush` may land in that drain or the next one. The guarantee covers what each worker had already been handed when `Flush` reached it.
- A non-nil error means a drain did not go through (transport failure, error status, unreadable response). Documents the cluster rejected individually reach their `OnFailure` callback and do not make `Flush` return an error.
- Safe to call repeatedly and concurrently with `Add` and other `Flush` calls. Like `Add`, it must not be called after `Close`: `Close` closes the worker queues, and a `Flush` reaching a closed queue panics.

### Breaking change

The exported `BulkIndexer` interface gains a method, so external implementations — usually a hand-written fake standing in for the indexer in tests — must add `Flush`. Indexers built via `NewBulkIndexer` gain it automatically. `UPGRADING_V5.md` carries the before/after, alongside the existing `opensearchtransport.Route`/`OpID()` precedent.

### Issues Resolved
Closes [#336]

For more information on following Developer Certificate of Origin and signing off your commits, please check here.


sean-:
@ryanyuan , can you look at this relative to #950 and fix this up accordingly?

iprithv (commented):
sorry but this seems to be fully superseded by what's already on main. the ring buffer, Flush() method, closing flag, sequencer goroutine, all of it landed through #950 and related commits. branch is many commits behind main, and if you rebase there would be no meaningful diff left. I think this can be closed.

ryanyuan:
@iprithv `Flush()` doesn't seem to be on main. I merged the `main` onto this branch so it should show the live diff now.

opensearchutil/bulk_indexer.go
```diff
@@ -455,12 +539,46 @@ func (w *worker) run(ctx context.Context) {
...
-			case item, ok := <-w.ch:
+			case entry, ok := <-w.ch:
 				if !ok {
 					// Channel closed, exit worker
 					return
 				}
 
+				if entry.flush != nil {
+					if w.bi.config.DebugLogger != nil {
+						w.bi.config.DebugLogger.Printf("[worker-%03d] Received flush barrier\n", w.id)
```
sean-, line 551: Why does this require a trailing `\n`?

CHANGELOG.md
```diff
@@ -180,6 +180,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
...
...t.Context) error` to `opensearchutil.BulkIndexer`, which sends every item added before the call and leaves the indexer open for further `Add` calls. `Close` was the only way to force a drain, and it closes each worker queue, so a caller that has to drain on demand -- a warm AWS Lambda handler holding its dependencies across invocations, or a test asserting on indexed documents -- had to build a new indexer per unit of work instead of flushing one. Neither existing threshold substitutes: `FlushInterval` means waiting on a timer, and `FlushBytes` means shrinking batches until the bulk API stops paying off. `Flush` sends a barrier down each worker queue and waits for the worker to report back. Each queue is FIFO, so by the time a worker takes the barrier it has already buffered every item queued ahead of it. Items added concurrently with `Flush` may land in that drain or the next one. Breaking only for code implementing the `BulkIndexer` interface, usually a hand-written test fake, which must add the method; see [`UPGRADING_V5.md`](UPGRADING_V5.md#opensearchutilbulkindexer-interface-gained-flush) for before/after ([#336](https://github.com/opensearch-project/opensearch-go/issues/336))
```
sean-: Why is this a breaking change?  It's an addition since `Flush(ctx)` is additive and net-new.

UPGRADING_V5.md
```diff
@@ -398,3 +398,43 @@ MatchPhrase: map[string]opensearchapi.CommonQueryDSLMatchPhraseQuery{"title": {}
...
+```go
+type BulkIndexer interface {
+    Add(context.Context, BulkIndexerItem) error
+    Flush(context.Context) error // new in v5
+    Close(context.Context) error
+    Stats() BulkIndexerStats
+}
+```
+
+External code that implements `BulkIndexer` must add a `Flush(context.Context) error` method. Indexers built via `NewBulkIndexer` gain it automatically; only hand-written implementations, most often a fake standing in for the indexer in tests, are affected.
```
sean-: Flip this around: "users of `NewBulkIndexer` automatically flush their contents when.....  custom implementations of BulkIndexer must now call Flush()`" or something.  Why wouldn't we call Flush() automatically before Close() to preserve existing behavior?

CHANGELOG.md
ryanyuan: I reckon it be breaking for anyone implementing the interface (?)
Maybe we could move it to `### Added` and drop that marker.

opensearchutil/bulk_indexer.go
ryanyuan: It seems the `BulkIndexerDebugLogger` just a `Printf()` with no promise to append a newline.
https://github.com/LeeFred3042U/opensearch-go/blob/9d5799dce98e0dc685d4c4ab10f5ee1078e06057/opensearchutil/bulk_indexer.go#L179

UPGRADING_V5.md
ryanyuan: Updated the wording.

Re: calling it automatically, Close() already drains. It first closes the queues, waits for the workers to finish, then flushes.

sean-: So then we should just say:

> `BulkIndexer` is a buffered queue that flushes its queue on `Close()` and explicitly via `Flush()`.

because that's clear.

opensearchutil/bulk_indexer.go
```diff
@@ -72,6 +72,16 @@ type BulkIndexer interface {
 	// actions on one document are sent in the order they were added.
 	Add(context.Context, BulkIndexerItem) error
 
+	// Flush sends every item added before the call and leaves the indexer open,
```
sean-: > Flush sends every item in BulkIndexer's queue before the call was made and leaves the indexer open,

opensearchutil/bulk_indexer.go
```diff
@@ -72,6 +72,16 @@ type BulkIndexer interface {
...
+	// so one indexer can drain on demand and keep going. Items added
+	// concurrently may land in this drain or the next, and the bulk requests run
+	// on ctx.
+	//
+	// A non-nil error means the drain itself failed; documents the cluster
+	// rejected individually go to their OnFailure callback. Safe alongside Add
+	// and other Flush calls, but like Add it panics after Close.
```
sean-: > 	// and other concurrent Flush calls.  Like Add, any Flush calls after Close will result in a panic.

CHANGELOG.md
sean-: I would move this to Added, not breaking.  I would maybe mark the `BulkIndexer` interface as having expanded to include `Flush()`.

This changelog entry is wordy, too.

sean- (dismissed):
Comment nits are minor and can be addressed in the future if desired.

opensearchutil/bulk_indexer.go
sean-: Can we migrate this to our debuglog package before v5?

sean- (approved):

Issue #864, opened 2026-06-08
[BUG] Bench bugs in opensearchtransport benchmarks: hoist New() out of loop, close clients, fix shared-body and read-after-close


sean-:
## Summary

The benchmarks under `opensearchtransport/` (and the root `BenchmarkClient`) have several pre-existing bugs that were exposed when the transport refactor on `worktree-transport-correctness-fixes` made `transport.New()` spawn long-lived background goroutines (`scheduleNodeStats`, `scheduleClusterHealthRefresh`).

These benches need to be fixed before v5 is cut, because:

1. On main today, running `go test -bench=. -benchtime=5s -count=10 ./opensearchtransport/` **hangs** at high iteration counts. Each iteration calls `opensearchtransport.New(...)` and never closes the client; the per-client schedulers accumulate to hundreds of thousands of idle goroutines and the runtime starves the bench loop itself.
2. Two of the benchmarks have correctness bugs that mean they have not been measuring what their names suggest, going back to v2.

Once the v5 branch is cut these will compound (more constructor work in `New()`, more attached state) and the benches become useless for tracking regressions.

## Bug 1: `New()` is called per-iteration and never closed

`opensearchtransport/opensearchtransport_benchmark_test.go` and `logger_benchmark_test.go` both contain loops shaped like:

```go
b.Run("Defaults", func(b *testing.B) {
    for i := 0; i < b.N; i++ {
        tp, _ := opensearchtransport.New(opensearchtransport.Config{
            URLs:      []*url.URL{{Scheme: "http", Host: "foo"}},
            Transport: newFakeTransport(b),
        })
        req, _ := http.NewRequest(http.MethodGet, "/abc", nil)
        res, err := tp.Perform(req)
        // ...
        defer res.Body.Close()  // also: defer in a loop, see Bug 4
    }
})
```

Problems:
- We are not benchmarking `Perform`. We are benchmarking `New()` plus `Perform`. Constructor cost dominates the measurement.
- On the v5-bound transport, `New()` schedules background tickers (`scheduleNodeStats` and `scheduleClusterHealthRefresh`) tied to a context that is only cancelled by `Close()`. Without `Close()`, those goroutines live until process exit. After ~250k iterations the scheduler is buried.
- `Perform()` is what changed across the v2-to-v5 transport rewrites; that is what we want a stable benchmark for.

## Bug 2: `FakeTransport` shares a stateful `Body` across iterations

```go
[...]

PR #931, opened 2026-07-09
fix(opensearchtransport): default benchmark pprof to an ephemeral port

ryanyuan:
### Description

Back-to-back `go test -bench` runs could fail to start the pprof server because the previous run's socket was still in TIME_WAIT on a fixed port. This defaults pprof to an ephemeral loopback port (`localhost:0`), so repeated runs pick a free port instead of colliding.

The startup logic moves out of the benchmark test file into a small `internal/pprofutil` package. `PPROF_ADDR` sets the address: empty binds an ephemeral loopback port, and an explicit `host:port` (with or without an `http://` scheme) pins it, e.g. `localhost:6060`. Malformed input is logged and skipped rather than bound.

This is the last of the five benchmark bugs from #864; the other four were fixed in #859.

### Issues Resolved
Closes [#864]


internal/pprofutil/pprofutil.go
```diff
@@ -0,0 +1,74 @@
...
+		return nil
+	}
+
+	ln, err := net.Listen("tcp", u.Host)
+	if err != nil {
+		log.Printf("pprof server failed to listen: %v", err)
+		return nil
+	}
+
+	log.Printf("pprof server listening on http://%s/debug/pprof/", ln.Addr())
```
sean-: Why aren't we using `u.String()`?  The path` /debug/pprof` is also configurable, iirc, so we should use a local constant for this and also have a single-source-of-truth for the URL if not specified.

internal/pprofutil/pprofutil_test.go
```diff
@@ -0,0 +1,118 @@
...
+		})
+	}
+}
+
+// TestStartNoCollisionOnRepeatedCalls guards issue #864 Bug 5. It proves
+// ephemeral-port allocation hands out distinct ports, so repeated startups do
+// not fight over one address. The real-world trigger -- a TIME_WAIT socket held
+// by a prior `go test -bench` process -- cannot be reproduced within a single
+// test binary, so two in-process starts is the closest observable proxy.
+func TestStartNoCollisionOnRepeatedCalls(t *testing.T) {
```
sean-: Do we need to test the behavior of `:0`?

internal/pprofutil/pprofutil.go
ryanyuan: right now `/debug/pprof` is not configurable because the blank import (https://pkg.go.dev/net/http/pprof#section-documentation) hardcodes registration on DefaultServeMux in its init().

reference: https://pkg.go.dev/net/http/pprof#section-documentation

We could drop the blank import and register handlers by hand.

internal/pprofutil/pprofutil_test.go
ryanyuan: I think the `TestStart()` should cover the `:0` test case and make `:0` and `localhost:0` a duplicate here in this test case.

But we can also cover it here. WDYT?

sean-: Please.  I want minimal redundancy and don't want a fluff test that tests Go's `:0` built-in.

internal/pprofutil/pprofutil.go
sean-: I think that's what's done elsewhere... or maybe not in this project, but yes.  For a generic helper func like this, that's how I'd spin things up (plus it also gets us a context on the http server which we can use to hook up signals in the future if we need to).

sean- (approved):

PR #933, opened 2026-07-10
feat: add osapifix, a version-hop migration tool for opensearch-go v4->v5

ryanyuan:
### Description

Adds `osapifix`, a tool that migrates a Go module across opensearch-go major versions. It detects the source major from the module's imports and defaults the target to the newest supported version, so the usual invocation is `osapifix rewrite -w ./...`.

Two subcommands:

- `rewrite` — rewrites old-major API shapes (type names, method paths, field spellings) into the target's. Purely syntactic (go/parser + astutil + go/printer), so it runs before the code compiles against the target. Writes are sandboxed to the target module directory via `os.Root`.
- `vet` — runs go/analysis analyzers that catch runtime type-assertion hazards (the target's precise `*int64`/`*string` types flowing into `any` sinks like testify's `Equal`, which compile but panic at run time). `-fix` applies the safe rewrites.

Hops are added as data — the v4→v5 hop ships now; v3→v4 and v2→v3 will be coming soon. Also wires the osapifix module into `lint.local` and `test-unit`, regenerates `opensearchapi` for CPU/DFS acronym casing, and adds the matching osgen acronym-table entries.

### Issues Resolved
N/A — new tooling.


sean- (approved):
Given this is net-new functionality, merging for now (to make way for `v3` and `v2` support) and any bug fixes can be addressed before `rc2` or the `v5.0.0` release.

PR #934, opened 2026-07-10
feat: update osapifix, a version-hop migration tool for opensearch-go v3->v4

ryanyuan:
### Description
Adds the v3 -> v4 hop to osapifix, so the tool can now migrate v3 -> v4 and chain v3 -> v4 -> v5 in one run. Only new data plus its committed v3 API surface (surface_v3.json); no engine changes.

The v3 -> v4 boundary is quiet compared to v4 -> v5. The generated opensearchapi client and its sub-clients are byte-identical across the two versions, so there are no method regroups and no *Req argument changes. Most of the Hop tables are empty on purpose: the surface diff derives the mechanical changes on its own (about 90 fields that became pointers), and any field that genuinely vanished falls through to the fail-loud "unclassified" default until a real consumer proves a ruling is needed. Same discipline as hop_v4_to_v5.go.

The one change a human has to make is the error model. The API error types (Error, Err, RootCause, StringError) moved out of opensearchapi into the root opensearch package and were reshaped. The engine only rewrites type names and version-bumps import prefixes, never a package qualifier, so a cross-package rename would pass the drift guard but emit code
that does not compile. It is reported as a SemanticFollowup instead, documented in opensearchapi/UPGRADING_V3_TO_V4.md.

### Issues Resolved
N/A — new tooling for v3-> v4 migration.


sean- (approved):
This is the v3 -> v4 migration code.  Same store as the v4 -> v5 work: this is net-new functionality, so any bug fixes between now and the release can be fixed in a follow-up PR.

NOTE: this PR paves the way for v2 -> v3 to land as a draft (a more substantial lift than v3 -> v4 -> v5).

PR #950, opened 2026-07-12
feat[opensearchutil]: route subsequent bulk requests on the same documentID to the same worker.

yanaix10:
**Changes made :** 

- Struct Update - queue chan BulkIndexerItem -> queues []chan BulkIndexerItem sized to NumWorkers
- Implemented OpenSearch shardhash.Hash() for documentID.
- Item routed to hashValue % NumWorkers.
- If no DocumentID is present, it falls back to a lock-free round-robin distribution via atomic.Uint64.
- Updated close function : Iterates through and closes all worker channels.


yanaix10:
hey @sean-  I am not very sure about the test case I wrote. I hardcorded the documentID.

opensearchutil/bulk_indexer.go
```diff
@@ -166,10 +167,11 @@ type BulkIndexerDebugLogger interface {
...
 type bulkIndexer struct {
-	wg      sync.WaitGroup
-	queue   chan BulkIndexerItem
-	workers []*worker
-	ticker  *time.Ticker
+	wg        sync.WaitGroup
+	queues    []chan BulkIndexerItem
+	rrCounter atomic.Uint64 // added for round-robin fallback
+	workers   []*worker
+	ticker    *time.Ticker
```
sean-, line 182: Can we store an instance of an [`opensearchtransport.DocRouter`](https://pkg.go.dev/github.com/opensearch-project/opensearch-go/v5@v5.0.0-rc1/opensearchtransport#DocRouter) here instead and just use that?

opensearchutil/bulk_indexer.go
```diff
@@ -258,14 +260,28 @@ func NewBulkIndexer(cfg BulkIndexerConfig) (BulkIndexer, error) {
 //
 // Adding an item after a call to Close() will panic.
 func (bi *bulkIndexer) Add(ctx context.Context, item BulkIndexerItem) error {
+	var targetQueue chan BulkIndexerItem
```
sean-: Here, we'd take a `BulkIndexerItem`, encapsulate it as an `http.Request`, then pass it along to [DocRouter.Eval()](https://pkg.go.dev/github.com/opensearch-project/opensearch-go/v5@v5.0.0-rc1/opensearchtransport#DocRouter.Eval).  In fact, there may be a chance that the new routing code in v5 (and if opted-in, v4) may just "do this" out of the box on a best-effort basis, but to force this, we should have `opensearchtuil/` construct a `DocRouter` and firehose things into the `DocRouter`.

CHANGELOG.md
```diff
@@ -274,6 +274,8 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - Fix ISM RefreshSearchAnalyzers missing leading slash in URL path, causing HTTP/2 request failures ([#686](https://github.com/opensearch-project/opensearch-go/pull/686))
 - Default the benchmark pprof server to an ephemeral loopback port so back-to-back `go test -bench` runs no longer collide on a `TIME_WAIT` socket held by the prior run. The startup logic moves into an `internal/pprofutil` package that registers the pprof handlers on a private mux (off `http.DefaultServeMux`); `PPROF_ADDR` pins an explicit `host:port` when needed. ([#864](https://github.com/opensearch-project/opensearch-go/issues/864))
 
+- Consistently route bulk requests for the same DocumentID to the same worker to prevent race conditions ([#950](https://github.com/opensearch-project/opensearch-go/pull/950)).
+
```
sean-: See https://github.com/opensearch-project/opensearch-go/blob/main/guides/transport-routing.md#document-path-docrouter and https://github.com/opensearch-project/opensearch-go/blob/main/guides/transport-routing.md#shared-connections-across-policies

sean- (commented):
Admittedly, I haven't looked into the [`opensearchutil/BulkIndexer`](https://pkg.go.dev/github.com/opensearch-project/opensearch-go/v5@v5.0.0-rc1/opensearchutil) much, but I'm not a fan of it in its current incarnation because it's interface is not particularly rigid.  That said, I think its internals should be adapted to use the `opensearchtransport/DocRouter` until a v2 of this package can be put together to tighten up its interface.

yanaix10:
hey @sean-   I have made required changes based on the feedback. Please review.

opensearchutil/bulk_indexer.go
```diff
@@ -254,18 +263,48 @@ func NewBulkIndexer(cfg BulkIndexerConfig) (BulkIndexer, error) {
...
+
+	//nolint:nestif // keep routing logic inline for simplicity
+	if item.DocumentID != "" {
+		idx := item.Index
+		if idx == "" {
+			idx = bi.config.Index
+		}
+
+		hashInput := item.DocumentID
+		path := fmt.Sprintf("/%s/_doc/%s", idx, item.DocumentID)
```
sean-: Can we generate this somehow not using Sprintf? url.URL with a paths.Join would be good, but we haven't validated idx or the DocumentID, yet, so I'm uncomfortable with Sprintf here.

opensearchutil/bulk_indexer.go
```diff
@@ -254,18 +263,48 @@ func NewBulkIndexer(cfg BulkIndexerConfig) (BulkIndexer, error) {
...
+		req, err := http.NewRequestWithContext(ctx, http.MethodPost, path, nil)
+
+		if err == nil {
+			hop, evalErr := bi.docRouter.Eval(ctx, req)
```
sean-: I forget off the top of my head (on mobile), does Eval perform the rendezvous hashing with costs, or just murmur hashing?

opensearchutil/bulk_indexer.go
```diff
@@ -254,18 +263,48 @@ func NewBulkIndexer(cfg BulkIndexerConfig) (BulkIndexer, error) {
...
+			if evalErr == nil && hop.Conn != nil {
+				hashInput = fmt.Sprintf("%p", hop.Conn)
+			}
+		}
+
+		//nolint:gosec // intentional conversion from signed to unsigned for modulo
+		workerIndex := uint32(shardhash.Hash(hashInput)) % uint32(bi.config.NumWorkers)
```
sean-: Why do we need to hash the pointer if we have a connection? Why not just use the connection?

opensearchutil/bulk_indexer.go
```diff
@@ -166,10 +168,12 @@ type BulkIndexerDebugLogger interface {
 }
 
 type bulkIndexer struct {
-	wg      sync.WaitGroup
-	queue   chan BulkIndexerItem
-	workers []*worker
-	ticker  *time.Ticker
+	wg        sync.WaitGroup
+	queues    []chan BulkIndexerItem
```
sean-: Can this be:

```suggestion
	queues    map[*Connection]chan BulkIndexerItem
```

?

opensearchutil/bulk_indexer.go
yanaix10: alright I will use url.URL with a paths.Join.

opensearchutil/bulk_indexer.go
yanaix10: I checked the Eval implementation in DocRouter, and it performs extensive cost calculations (calcSingleKeyCost, connScoreSelect) and uses rendezvousTopK as a fallback. Running this per document inside the Add loop will definitely introduce too much CPU overhead and memory allocation.

opensearchutil/bulk_indexer.go
yanaix10: If we still keep Eval()  changing []chan to a map introduces a concurrency risk.  we could wrap around sync.RWMutex . but every single goroutine calling Add() has to acquire that lock which destroys concurrency. 
instead we can use sync.Map to avoid lock contention.

opensearchutil/bulk_indexer.go
sean-: Have you profiled this?  The cost used to be trivial.  Has there been a regression?

opensearchutil/bulk_indexer.go
sean-: We only need to hold the lock for the duration of the lookup.  Something like:

```
struct mu {
  sync.RWMutex
  queues    map[*Connection]chan BulkIndexerItem
}
```

Would be sufficient.

opensearchutil/bulk_indexer.go
yanaix10: My bad, I missed the benchmark file earlier. You are correct—Eval() does use rendezvous hashing, and the cost remains trivial (~450ns/op with 0 allocations). I am keeping the Eval() call as is.

opensearchutil/bulk_indexer.go
```diff
@@ -254,18 +269,62 @@ func NewBulkIndexer(cfg BulkIndexerConfig) (BulkIndexer, error) {
...
+	var targetQueue chan BulkIndexerItem
+
+	//nolint:nestif // keep routing logic inline for simplicity
+	if item.DocumentID != "" {
+		idx := item.Index
+		if idx == "" {
+			idx = bi.config.Index
+		}
+
+		u := url.URL{Path: path.Join("/", idx, "_doc", item.DocumentID)}
```
sean-: There's nothing here that escapes `idx`: we're just trusting the user input blindly.  Can you look into `url.PathUnescape()` and look at the `./internal/build/` package and how it is used.

yanaix10: I have used url.PathEscape() on idx and documentID to handle user input. manually constructing the http.Request and explicitly setting Path (via url.PathUnescape()) and RawPath.  Due to manual construction I could drop http.NewRequest entirely - bypassing url.Parse allocation overhead.

sean-: Because this is in the bulk-path, I'm not too worried about the performance, but it's up to you.

sean- (dismissed):
Thank you for adapting to using the router!

opensearchutil/bulk_indexer.go
```diff
@@ -254,18 +268,69 @@ func NewBulkIndexer(cfg BulkIndexerConfig) (BulkIndexer, error) {
...
+	//nolint:nestif // keep routing logic inline for simplicity
+	if item.DocumentID != "" {
+		idx := item.Index
+		if idx == "" {
+			idx = bi.config.Index
+		}
+
+		encodedPath := "/" + url.PathEscape(idx) + "/_doc/" + url.PathEscape(item.DocumentID)
+
+		if p, err := url.PathUnescape(encodedPath); err == nil {
```
ryanyuan: Question about the escape/unescape pair here: does the escaping survive to Eval?
Would putting the escaped form straight into Path work for what you were after, dropping PathUnescape and RawPath?

opensearchutil/bulk_indexer.go
```diff
@@ -254,18 +268,69 @@ func NewBulkIndexer(cfg BulkIndexerConfig) (BulkIndexer, error) {
...
+			req := (&http.Request{
+				Method: http.MethodPost,
+				URL: &url.URL{
+					Path:    p,
+					RawPath: encodedPath,
+				},
+			}).WithContext(ctx)
+
+			if hop, evalErr := bi.docRouter.Eval(ctx, req); evalErr == nil && hop.Conn != nil {
```
ryanyuan: Does this Eval ever return a connection? I think NewDocRouter() gets its connections through DiscoveryUpdate(), which only the owning transport calls, so activeConns looks like it stays empty and Eval returns early at policy_doc_router.go:144-150.

ryanyuan (dismissed):

sean- (approved):

PR #951, opened 2026-07-13
feat: update osapifix, a version-hop migration tool for opensearch-go v2->v3

ryanyuan:
### Description

Adds the v2 -> v3 hop to osapifix, so a v2 module can be migrated through to v5 (the engine chains v2 -> v3 -> v4 -> v5).

Most of v2 -> v3 is not mechanizable. The opensearchapi package was redesigned from a function-based API into a typed sub-client API, and of 182 exported v2 request types only 16 survive by name. The hop rewrites what it can prove and reports the rest:
  - Import paths are bumped v2 -> v3 across every sub-package.
  - The removed opensearchapi.*Request types (idiom 1) and the removed root-opensearch.Client API methods (idiom 2) are surfaced as MANUAL worklist items rather than rewritten. Each is a shape change (raw *Response -> decoded typed *Resp, functional options -> Req struct), not a rename.
  - The two seed ops Ping and Indices.Exists are rewritten best-effort into compiling v3: the call itself, its raw-response Status() handling, and the Config/NewClient client lifecycle. Where a transform is not mechanically certain, the pass plants an `_OSAPIFIX_RESOLVE` marker so go build breaks at that spot instead of emitting a wrong value.

The v2 -> v3 correspondence lives in three hand-authored data tables (call map, arg-detail, field dispositions). Each is drift-guarded against the committed `surface_v2.json`/`surface_v3.json` a table cannot silently diverge from the real API. An end-to-end corpus test runs the real rewrite over fixture modules and diffs the output against committed goldens. The fixtures compile against hand-written stubs of the source API, so the test runs offline with no opensearch-go download.

The hop is added as data, with no engine changes, following the existing v3 -> v4 and v4 -> v5 hops.

### Issues Resolved

None. This adds a new capability (the v2 -> v3 migration hop)


ryanyuan:
Hi @Jakob3xD , @sean- , PTAL.

cmd/osapifix/applydelta.go
```diff
@@ -178,7 +178,28 @@ func rewriteFileTyped(pkg *packages.Package, file *ast.File, rules rewriteRules)
...
 	astutil.Apply(file, func(c *astutil.Cursor) bool {
+		if isV2Hop {
+			if edits, imports, handled := rewriteIdiom2Node(c, info, apiName, apiImportPath); handled {
+				res.edits = append(res.edits, edits...)
+				needImports = append(needImports, imports...)
+				// Stop descending: the rewritten call's stale selector (or the
+				// synthetic subtree) must not be re-walked, or flagFieldAccess would
+				// double-report the same op. Synthetic nodes also carry no
+				// info.Selections, so the type gates are inert on them regardless.
+				return false
```
Jakob3xD, line 208: Inline `NewClient(Config{...})` isn't fully migrated. The `NewClient` handler above mutates the selector in place (`sel.X = ast.NewIdent(apiName)`) and returns `handled=true`, so this `return false` prunes the `CallExpr`'s children — the `Config{...}` argument literal is never visited by the `CompositeLit` branch that would reshape it. Output becomes `opensearchapi.NewClient(opensearchv2.Config{...})`, which does not compile against v3 (`opensearchapi.NewClient` takes `opensearchapi.Config`), and no `_OSAPIFIX_RESOLVE` marker is planted — so the tool emits wrong code with only a clean "repoint NewClient" edit, contrary to its plant-a-marker-instead-of-guessing invariant.

The corpus only covers the split form (`cfg := Config{}; ...; NewClient(cfg)`), where the literal is a standalone assignment and gets reshaped independently; the inline form (the canonical README idiom) is untested. Fix: for the in-place `NewClient` repoint, also reshape the `Config` argument (or let the walk descend into it) rather than pruning — pruning is only needed for the synthetic-replacement cases, not the in-place mutations.

cmd/osapifix/internal/apirev/delta.go
```diff
@@ -82,6 +82,15 @@ type StructDelta struct {
 // source type.
 type Delta struct {
 	Structs map[string]StructDelta `json:"structs"`
+	// RemovedTypes is the set of fully-qualified source types ("<pkgPath>.<Name>")
+	// that have NO counterpart in the target and are not covered by a TypeRename -
+	// i.e. types deleted outright across the hop (e.g. the v2 opensearchapi.*Request
+	// family removed in v3's client redesign). The value is the version-agnostic
```
Jakob3xD: `RemovedTypes` is `map[string]bool` and the value is always `true` (set membership) — it is not a "display name". The qualified name is the map *key*. This sentence looks like a leftover from an earlier `map[string]string` shape; suggest dropping it (the rest of the doc is accurate).

Jakob3xD (commented):
### Code review

Reviewed the hand-written Go (the generated `surface_v2.json` was excluded). Two issues are posted as inline comments; the rest are non-blocking notes below. This is net-new tooling in draft, so none of it gates the PR — the inline `NewClient` issue is the one worth fixing before it leaves draft.

**Minor comment/code inaccuracies**

1. `rewrite_idiom2.go:29` — the `markerPrefix` doc says "the type-aware walk scans for it to record a MANUAL edit", but nothing scans for the marker; the MANUAL edit is recorded at plant time in `rewriteIdiom2Node`, and the marker's role is to break the build so a human resolves it.

https://github.com/opensearch-project/opensearch-go/blob/7df72e73450de5d3291fa78b34ae9bcf8483e169/cmd/osapifix/rewrite_idiom2.go#L28-L30

2. `internal/apirev/delta.go:160` — `DeriveDelta`'s doc still says a source struct with no target counterpart "is skipped", but this PR changed that branch to record the type in `RemovedTypes`.

https://github.com/opensearch-project/opensearch-go/blob/7df72e73450de5d3291fa78b34ae9bcf8483e169/cmd/osapifix/internal/apirev/delta.go#L159-L161

3. `callmap_v2_to_v3.go` header — the `RawResponse` prose enumerates 9 ops (7 `Exists*` + `Ping` + `Nodes.HotThreads`) but the count is 11; the two `Cluster.*VotingConfigExclusions` rows are omitted from the description.

4. `applydelta.go` `flagFieldAccess` doc understates behavior — it now resolves the field via both the declaring type and the receiver type, but the comment only mentions the declaring type.

**Suggestion — drift guard for the hand-typed callmap columns**

`ReqPtr` and `RawResponse` are the two columns `TestCallMapV2toV3AgainstSurfaces` cannot validate (hand-typed booleans, no drift guard). Consider a check that inspects the pinned v3 method signatures. One row looks off: `Cat.Snapshots` is the only `Cat.*` row without `ReqPtr: true` — worth confirming against the v3 `opensearchapi` source. (Unverified here: I could only check against the on-disk v5, where Req names differ across the hop, so pointer-ness may legitimately differ.)

cmd/osapifix/applydelta.go
```diff
@@ -207,9 +233,159 @@ func rewriteFileTyped(pkg *packages.Package, file *ast.File, rules rewriteRules)
...
 	// above still saw the source paths.
 	res.edits = append(res.edits, rewriteImports(file, rules.importPrefixes)...)
+
+	// Inject imports the idiom-2 pass introduced (v3 opensearchapi for the reshaped
+	// Config/Req, fmt+net/http for the raw-response Status rewrite). Done after the
+	// walk and rewriteImports, like the import bump, so type resolution saw the
+	// original source. AddImport is idempotent, so a path already present is a
+	// no-op.
+	for _, p := range dedupe(needImports) {
+		astutil.AddImport(pkg.Fset, file, p)
```
Jakob3xD, line 262: Aliased `opensearchapi` import produces a duplicate, unused import on the v2->v3 hop — the rewritten file won't compile.

When the source file imports the opensearchapi package under an explicit alias (`osapi "...opensearch-go/v2/opensearchapi"`), `idiom2ImportNames` correctly returns `osapi` (line 383) and the synthetic nodes reference `osapi.PingReq{}`. `rewriteImports` then bumps that spec's path to v3 **but keeps the alias** (it only touches `imp.Path.Value`). Here `astutil.AddImport(pkg.Fset, file, apiImportPath)` injects an *unnamed* spec for the same v3 path. `astutil`'s dedup (`imports(f, name, path)`) matches on **both** name and path, so `(name="", v3path)` does not match the existing `(name="osapi", v3path)` → a second, unnamed `.../v3/opensearchapi` import is added. It is unused (all refs go through `osapi.`), so the output fails with `imported and not used`. The comment on line 240-241 ("already present is a no-op") holds only for the unnamed case; the corpus only covers the unnamed import, so this is untested.

Fix: inject with the resolved alias, e.g. thread `apiName` here and call `astutil.AddNamedImport(pkg.Fset, file, apiName, p)` for the opensearchapi path — then the dedup matches the bumped aliased spec and it becomes a no-op. The same unnamed-injection also affects the `fmt`/`net/http` inject (line 299) if the file binds `fmt`/`http` to another package, though that trigger is far narrower.

cmd/osapifix/applydelta.go
Jakob3xD, line 208: Descent-stop can skip a genuine v2 subexpression that is reused verbatim into the v3 Req, yielding non-compiling output with no marker.

Returning `false` here halts descent into all children of a handled node. The comment justifies this for the *synthetic* replacement's stale selector — correct. But `rewriteIdiom2Call` carries **original, live-typed** v2 subexpressions into the new Req unchanged: positional values (rewrite_idiom2.go:89-92), option values (`optionValue` returns `oc.Args[0]` verbatim), and the ctx arg. If a reused subexpression is itself a v2 root-client construct (e.g. an arg whose value embeds another `client.X(...)` call), descent stops and it is never rewritten or flagged — the file won't compile, and no `_OSAPIFIX_RESOLVE` marker is planted. That contradicts the pass's stated rewrite-or-mark invariant (rewrite_idiom2.go:20-23).

Currently narrow (only Ping/Indices.Exists are wired and their args are usually plain), but the transform is general and widening `argDetailV2toV3` exposes it. Consider re-walking the reused sub-args, or planting a marker when a reused arg still contains a v2 root-client reference.

Jakob3xD (commented):
### Code review

Re-review on the current head; my earlier findings (inline `NewClient(Config{...})` reshape, the `RemovedTypes`/comment doc nits) are all fixed, and the `Cat.Snapshots` `ReqPtr` row I flagged last time is in fact correct (v3 `catClient.Snapshots` takes a value `CatSnapshotsReq`, unlike every other `Cat.*`). Nicely done.

Found 3 issue(s) on the new v2->v3 idiom-2 pass — all posted as inline comments, none blocking.

cmd/osapifix/rewrite_idiom2.go
```diff
@@ -0,0 +1,338 @@
...
+		}
+		return &ast.CallExpr{
+			Fun:  fun,
+			Args: []ast.Expr{ctxArg, markerExpr(strings.Join(salvage, "; "))},
+		}, edits
+	}
+
+	// Emit the collected params as a single nested Params field.
+	if len(paramFields) > 0 {
+		paramsType := strings.TrimSuffix(entry.V3Req, "Req") + "Params"
```
Jakob3xD, line 201: Non-blocking: the `destParams` emit path leans on two naming conventions that are only half-guarded. The params type is derived as `<Op>Params` (this line) and the value is placed under a Req field literally named `Params` (line 162). `TestArgDetailV2toV3AgainstSurfaces` asserts the `<Op>Params` **type** exists, but never asserts the `<Op>Req` actually has a field named `Params` holding it. If a future op added to the table names that field differently, the emitted `Req{Params: ...}` would reference a non-existent field → non-compiling output, no marker, and the drift guard stays green. There is also no golden-corpus coverage of the destParams branch (only a unit test). Suggest extending the drift guard to also require `reqFields["Params"]` (typed `<Op>Params`) for any op carrying a destParams option — the guards exist precisely to keep this convention-derived emit safe as the table grows.

cmd/osapifix/applydelta.go
ryanyuan: Fixed. Inject now guards on a by-path check instead of astutil's dedupe. Also added aliasedimport.go golden to pin the single-import output.

cmd/osapifix/applydelta.go
ryanyuan: Fixed. 
- Added an unsafeReuse predicate to rewriteIdiom2Call and applydelta passes containsV2RootClientRef.
- It checks every subexpressions carried into the v3 Req (positionals, ctx, option values) for a v2 root-client type.
- On a hit it plants an `_OSAPIFIX_RESOLVE`marker instead of emitting the live ref.
- Added carriedrootclient.go golden plus a unit test.

cmd/osapifix/rewrite_idiom2.go
ryanyuan: Fixed.
TestArgDetailV2toV3AgainstSurfaces now asserts any destParams op has a <Op>Req.Params field typed <Op>Params.
Also Added paramsemit.go golden for corpus coverage.

cmd/osapifix/applydelta.go
```diff
@@ -207,9 +233,188 @@ func rewriteFileTyped(pkg *packages.Package, file *ast.File, rules rewriteRules)
...
+	//
+	// Guard on the resolved import PATH, not astutil's own dedup: astutil matches
+	// on (name, path), so AddImport injecting an unnamed spec for a path the file
+	// already imports under an alias (osapi ".../opensearchapi", which rewriteImports
+	// just bumped to v3 in place) would NOT dedupe - it would add a second, unnamed,
+	// unused import and the output would fail with "imported and not used". The
+	// synthetic nodes reference the file's existing alias (apiName, resolved by
+	// idiom2ImportNames from that same spec), so a path already imported under any
+	// name needs no injection at all; only a genuinely new path is added.
+	for _, p := range dedupe(needImports) {
```
Jakob3xD, line 258: Idiom-2 can leave the bumped root import unused, producing an `imported and not used` compile error. When a file's only use of the root package is the `*opensearch.Client` type, the idiom-2 pass rewrites that reference to `*opensearchapi.Client`, and `rewriteImports` bumps the root import path v2→v3 in place — but nothing removes it. This loop only *injects* `needImports`; there is no symmetric removal of an import the pass renders dead.

Evidence: `testdata/corpus/v2/paramsemit.go.golden:12` emits `opensearchv2 "…/opensearch-go/v3"`, which the body never references. That golden has no `_OSAPIFIX_RESOLVE` marker and no removed-type reference, so it is meant to compile — and the README states idiom-2 output compiles ("rewritten to compiling v3"). `TestRewriteCorpus` does not catch it because it never runs `go build` on the goldens.

Suggested fix: after the walk, `astutil.DeleteImport` the bumped root path when it is no longer referenced in the file. (`seedops.go.golden` is unaffected — its Config reshape keeps `opensearchv2.Config{…}` in the output, so the root import stays used.)

cmd/osapifix/callmap_v2_to_v3.go
```diff
@@ -0,0 +1,235 @@
...
+// correspondence between a v2 root/sub-client call path and its v3 typed
+// sub-client destination. It is the data the surface JSON cannot carry - the
+// surfaces record struct fields, not the method regrouping and *Request -> *Req
+// renames that the v3 redesign performed.
+//
+// It exists so the v2 -> v3 hop can eventually mechanize the CALL half of the
+// migration (client.<v2path>(...) -> client.<v3path>(ctx, <Req>{...})) for the
+// ops it covers, and report the rest. This file is only the map + its drift
+// guard (callmap_v2_to_v3_test.go); wiring it into the rewriter is a later
+// increment. Until then the whole v2 -> v3 boundary stays report-only (see
```
Jakob3xD: "the whole v2 -> v3 boundary stays report-only" is no longer accurate. `rewrite_idiom2.go:81` iterates `callMapV2toV3`, and `Ping` / `Indices.Exists` are rewritten by this PR (asserted by `TestRewriteCorpus`, e.g. the `rewrite client.Ping` edit). The map *is* wired into the rewriter for the seed ops. Consider scoping the claim to the ops not yet mechanized rather than the whole boundary.

Jakob3xD (commented):
### Code review (R3)

All blocking findings from the earlier rounds are verified resolved on this head: import dedup (`fileImportsPath`), the descent-stop marker (`containsV2RootClientRef`), and the `destParams` guard (`TestArgDetailV2toV3AgainstSurfaces`), each pinned by a new golden. This round has one compile bug and two doc-accuracy notes (all inline below), plus one scope finding without a single-line anchor. Nothing blocking.

**Cross-hop scope: `RemovedTypes` / `flagRemovedTypeRef` are not version-gated**

`DeriveDelta` now records every removed source type in `RemovedTypes` unconditionally, and `flagRemovedTypeRef` runs in the generic `*ast.SelectorExpr` branch (applydelta.go:217) — outside the `isV2Hop` gate. So it fires for every hop, not only v2→v3. The v4→v5 surface drops many types (v3→v4 drops none), so a v4→v5 run now emits new `MANUAL … removed on the target` worklist lines it did not produce before this PR. No v4 corpus covers this, and the PR description scopes the change to v2→v3. If this cross-hop effect is intended, a CHANGELOG note and a v4-consumer corpus case would lock the new behavior in.

https://github.com/opensearch-project/opensearch-go/blob/a1f58d2c23a5c3cd27c7f6a65953ac948ea7c68d/cmd/osapifix/internal/apirev/delta.go#L191-L196

https://github.com/opensearch-project/opensearch-go/blob/a1f58d2c23a5c3cd27c7f6a65953ac948ea7c68d/cmd/osapifix/applydelta.go#L214-L219

cmd/osapifix/hop_v2_to_v3.go
```diff
@@ -0,0 +1,152 @@
...
+//     and the receiver changes from the root *opensearch.Client to a separate
+//     *opensearchapi.Client. That is a per-op semantic rewrite, not a rename.
+//
+//   - Idiom 2 (root client): client.Ping(client.Ping.WithContext(ctx)) plus
+//     resp.IsError()/resp.Status(). The root opensearch.Client lost all 51 of its
+//     API method fields in v3 (only Transport survives); the functional-option
+//     args collapse into a Req struct, and the raw-response error check moves to
+//     the returned error. Also a control-flow change, not a rename.
+//
+// So this hop rewrites only what is genuinely mechanical - the import-path bump -
```
Jakob3xD: Minor: "reports everything else" reads against line 134 of this same file ("Seed ops (Ping, Indices.Exists) are rewritten best-effort; everything else is report-only"). The seed-op rewrites are the exception to "reports everything else" — worth naming that exception in this header paragraph so the two descriptions agree.

cmd/osapifix/applydelta.go
ryanyuan: Done. Now prune the bumped root import once nothing references it (alias-aware). Three goldens regenerated and paramsemit compiles clean.

cmd/osapifix/callmap_v2_to_v3.go
ryanyuan: Fixed. reworded to rewriter handles the seed ops, rest stays report-only.

cmd/osapifix/hop_v2_to_v3.go
ryanyuan: Fixed. Now names the seed-op rewrites, not just the import bump.

sean- (approved):

PR #952, opened 2026-07-13
fix(opensearchtransport): don't let newly discovered nodes mask seed fallback

sean-:
Node discovery inserts freshly discovered connections into a policy pool's dead list before they are ever health-checked (createConnection sets lcDead|lcNeedsWarmup|lcNeedsHardware). Each policy cached its "enabled" bit as len(ready)+len(dead) > 0, so a policy holding only never-verified discovered connections still reported enabled. Route then handed the request to the pool, which served the never-verified connection as a zombie (tryZombieWithLock) rather than returning ErrNoConnections -- so the request never cascaded to the user-supplied seed-URL fallback (which only triggers on ErrNoConnections).

When discovered publish_addresses are unroutable from the client (NAT'd or misconfigured clusters, as seen in CI against a k8s stack cluster), this hijacks the request stream onto dead pod IPs and every request fails with "no route to host", even though the seed URL is reachable.

Compute the enabled bit from availability instead of presence: a connection is available for routing if it is a user-supplied seed (assumed reachable) or a discovered connection confirmed reachable (lcNeedsHardware cleared). A dead, never-verified discovered connection no longer counts, so the policy reports not-enabled, Route returns ErrNoConnections, and the seed fallback serves the request until a discovered node health-checks clean and takes over.

- Add Connection.seed (immutable, set at the two seed construction sites) and Connection.availableForRouting().
- Add multiServerPool.hasAvailableConnsWithLock(); use it for the enabled bit in the round-robin, role, and coordinator policies, and apply the same per-connection check in the index/doc router's flat active list.

Genuinely-once-healthy dead connections (OnFailure re-sets lcNeedsHardware, but they are re-verified on the next health check) and steady-state zombie behavior are unchanged; seeds are always available so the fallback path is preserved.


opensearchtransport/policy_role.go
```diff
@@ -346,9 +347,10 @@ func (p *RolePolicy) discoveryUpdateRemove(removed []*Connection) {
 	gap := activeCountBefore - p.pool.mu.activeCount
 	p.pool.promoteStandbyGracefullyWithLock(p.pool.poolCtx(), gap)
 
-	// Update hasMatching state while holding the lock
-	hasConnections := len(p.pool.mu.ready) > 0 || len(p.pool.mu.dead) > 0
-	psSetEnabled(&p.policyState, hasConnections)
+	// Update hasMatching state while holding the lock. Dead-but-never-verified
+	// connections do not count as available so the request cascades to the seed
+	// fallback rather than being served as a zombie.
+	psSetEnabled(&p.policyState, p.pool.hasAvailableConnsWithLock())
```
Jakob3xD: RolePolicy's cached `enabled` bit can go stale against pool reality. This PR changed the predicate from a churn-invariant one (`len(ready)+len(dead) > 0`, unaffected by dead<->ready transitions) to a churn-sensitive one (`hasAvailableConnsWithLock` keys on `lcNeedsHardware`, which flips on the health-check/resurrect path).

But both of RolePolicy's recompute sites are gated: the one in `discoveryUpdateAdd` (:259) runs only inside the per-role-matching-added-connection loop, and this one in `discoveryUpdateRemove` sits after the `if !hasConnectionsToRemove { return }` early return (:294). On an unchanged-only discovery cycle — the steady state for a stable cluster — neither fires. `psSetEnabled` is never called anywhere on the health-check/resurrect path (verified: zero occurrences in pool_multi_server.go / opensearchtransport.go; `IsEnabled()` just reads the cached bit).

So when a discovered node transitions unverified -> verified via a health check (no membership change), RolePolicy's bit stays `false` until the next add/remove touching this pool, and `PolicyChain.Route` keeps skipping the policy — the healthy discovered node is never routed to and traffic pins to the seed fallback. RoundRobin and Coordinator avoid this because they recompute unconditionally after their `added == nil && removed == nil` short-circuit; RolePolicy does not. Degraded operation rather than an outage (the seed keeps serving), unless the seed also fails while a healthy node sits stale-disabled.

Consider recomputing RolePolicy's bit unconditionally per discovery cycle (like the other policies), or recomputing on the resurrect/health-check path when a connection's routability changes.

opensearchtransport/policy_roundrobin.go
```diff
@@ -196,8 +196,12 @@ func (p *RoundRobinPolicy) DiscoveryUpdate(added, removed, unchanged []*Connecti
 		}
 	}
 
-	// Update cached enabled state
-	psSetEnabled(&p.policyState, len(p.pool.mu.ready)+len(p.pool.mu.dead) > 0)
+	// Update cached enabled state. A policy is enabled only when it has a
+	// connection worth routing to -- a ready connection, or a dead connection
+	// that has been verified at least once. Dead-but-never-verified connections
```
Jakob3xD: The comment says a dead connection "that has been verified at least once" counts as available, but `availableForRouting` checks `!has(lcNeedsHardware)` — a current-state test, not an "ever verified" latch. Since `OnFailure` re-sets `lcNeedsHardware` on any live->dead transition, a connection that was verified once and then failed lands back in the dead list with the bit set and is *not* available. Suggest rewording to something like "a dead connection currently confirmed reachable (lcNeedsHardware clear)" to match the actual predicate.

Jakob3xD (commented):
### Code review

Reviewed the production changes (the new `*_test.go` files were read for behavior but not line-audited). One issue posted inline is worth addressing; a second is a comment-accuracy nit. Two non-blocking notes below.

**Question — `hasAvailableConnsWithLock` vs the index router treat `ready`/active connections differently**

`hasAvailableConnsWithLock` short-circuits `if len(cp.mu.ready) > 0 { return true }` — every ready connection counts unconditionally. `routerDiscoveryUpdate` instead calls `availableForRouting()` on every connection in its `all` list, ready ones included. `resurrectWithLock` moves a connection dead->ready while clearing only `lcUnknown|lcStandby` (pool_multi_server.go:600/607), i.e. it preserves `lcNeedsHardware`. If an active connection can reach the ready list still carrying `lcNeedsHardware`, the two policy families would compute availability differently for the same connection state. Is that reachable, and if so is the router's stricter per-connection check the intended semantics (vs. the pool's "any ready counts")? Not necessarily a bug — flagging for confirmation.

**Minor — `availableForRouting` doc wording**

connection.go says the `lcNeedsHardware` bit is "cleared only by a successful hardware/health check." It's actually cleared whenever the node *responds at all* (the clear also fires on a non-200 such as 403) — i.e. reachability, not success. The `availableForRouting` semantics are correct; only the word "successful" is loose.

Jakob3xD (dismissed):

opensearchtransport/policy_role.go
sean-: Yup, this was a real bug, and it was the culprit for the broken `TestShardExactRouting_FullPipeline_Integration` test. When a discovered node verified via health check with no membership change, RolePolicy's bit stayed `false`, `Route` skipped the policy, and the readiness probe never observed a `RouteEvent`.

Fixed by recomputing RolePolicy's enabled bit **unconditionally** at the end of `DiscoveryUpdate`, mirroring RoundRobin/Coordinator, and dropping the now redundant `discoveryUpdateAdd`/`discoveryUpdateRemove` recomputation points.

opensearchtransport/policy_roundrobin.go
sean-: You're right — the predicate is a current-state test, not an ever-verified latch, and `OnFailure` re-sets `lcNeedsHardware`. Reworded the comment to match:

```
    // Update cached enabled state. A policy is enabled only when it has a
    // connection worth routing to -- a ready connection, or a dead connection
    // currently confirmed reachable (lcNeedsHardware clear; see
    // availableForRouting). Dead connections not currently confirmed reachable
    // (freshly discovered, or verified-then-failed, possibly unroutable) do not
    // count, so the request cascades to the seed fallback instead of being
    // served as a zombie.
```

sean-:
Thanks for the review, both points have been addressed.

sean-:
Closing in favor of #954

PR #953, opened 2026-07-13
ci: remove unpinned compatibility target

### LINKED ISSUE #925: Simplify the v4 compatibility CI matrix
### Summary

The `v4` compatibility matrix in `.github/workflows/test-compatibility.yml` includes a `latest` OpenSearch target that fails on every run. It is currently masked with `continue-on-error`, so it never fails the workflow but consumes runner time and adds persistent red status on the job's matrix. This issue proposes removing the `latest` target and aligning the documented support range with the versions actually tested.

### Background

`latest` fails on every recent run because current OpenSearch emits a `native_memory` node-stats field that the v4 response struct does not model:

```
Error: Should be empty, but was [{"op":"remove","path":"/nodes/.../native_memory"} ...]
Test:  TestNodes/Stats/with_request
Test:  TestNodesStats/success
```

The failure is hidden by `continue-on-error: ${{ matrix.entry.opensearch_version == 'latest' }}`. The `latest` target is also above the highest pinned version in the matrix (3.6.0) and above the range documented in `COMPATIBILITY.md`.

All pinned versions from 1.3.20 through 3.6.0 pass (secured and insecure).

### Proposed changes

- [x] Remove the `latest` entry from the `test-compatibility.yml` matrix.
- [x] Remove the now-unused `continue-on-error` expression on the matrix entry.
- [x] Update `COMPATIBILITY.md` to state that the v4 client is tested against OpenSearch through 3.6.x; versions beyond that are untested and supported on a best-effort basis.

### Acceptance criteria

- The compatibility matrix tests only pinned versions (1.3.20 through 3.6.0).
- No `continue-on-error` entries remain in the matrix.
- `COMPATIBILITY.md` reflects the tested range (through 3.6.x) and notes best-effort support beyond it.
- A clean `v4` produces a green Integration for Compatibility check.

Seekers2001:
## Summary

- remove the unpinned `latest` OpenSearch target from the compatibility matrix
- remove the matching `continue-on-error` exception so every matrix job is required
- document the exact pinned 5.x compatibility range and best-effort behavior for newer releases
- record the CI change in the unreleased changelog

Closes #925.

## Validation

- `git diff --check`
- parsed `.github/workflows/test-compatibility.yml` with Ruby YAML and asserted that neither the `latest` target nor the job-level `continue-on-error` key remains
- confirmed the pinned matrix resolves to OpenSearch 2.19.5 and 3.0.0 through 3.7.0

The full compatibility integration matrix is intentionally left to GitHub Actions because it provisions eighteen three-node OpenSearch clusters.


sean- (changes requested):
We want latest on the main/v5 branch. v4, however, needs this exact treatment. Could you please target the V4 branch and try again if you're interested in fixing this issue?

Issue #977, opened 2026-07-17
V4.7.0 Migration guide recommends to adjust removed method `opensearch.BuildRequest`


mgoetzegb:
https://github.com/opensearch-project/opensearch-go/blob/f818ecb6af101d0f729c3eae32c3c4cc6ccc5ae2/UPGRADING.md?plain=1#L159-L171

The V4.7.0 Migration guide mentions the usage of `opensearch.BuildRequest`, however this function is no longer present in V4.7.0. So instructions for a migration of usages of `opensearch.BuildRequest` would be helpful. 

Also further methods like `opensearchapi.Client.Do` requires now the `method` parameter. Is there any guidance to determine the value to be passed? (I assume it is just checking the method set in the respective `GetRequest` in the code prior to V4.7.0`).  That would ease the migration.

sean-:
Thanks for the report, @mgoetzegb.  And yeah, the `BuildRequest` mention is a bug in the guide and wasn't revisited after its removal.

## `opensearch.BuildRequest`

`opensearch.BuildRequest` was **removed in 4.7.0** (it existed only through 4.6.0), so the 4.7.0 migration section should never referenced it.

The `Request` interface is the only thing that changed shape: `GetRequest` now receives the method from the caller. If you maintain a custom type that satisfies `opensearch.Request`, build the `*http.Request` with `net/http` directly.  For example:

```go
// Before (<= 4.6.0)
func (r customReq) GetRequest() (*http.Request, error) {
    return opensearch.BuildRequest(http.MethodGet, r.path, r.body, nil, nil)
}

// After (>= 4.7.0)
func (r customReq) GetRequest(method string) (*http.Request, error) {
    return http.NewRequest(method, r.path, r.body)
}
```

Pass a **relative** path (e.g. `/_plugins/my_plugin/status`) -- the transport prepends the base URL. `BuildRequest` also took `params map[string]string` and `headers http.Header`; if you relied on those, set them on the request after construction (`req.URL.RawQuery` via `url.Values`, and `req.Header`).

## Determining the `method` argument

And yes, the "`method`" to pass is whatever the operation used before. For the typed request structs the client ships, this is already
wired up for you -- only hand-written `opensearch.Request` implementations need attention, and there you pass the same method your old `GetRequest` hard-coded in its `http.NewRequest` call.

[...]

PR #978, opened 2026-07-17
docs: fix removed opensearch.BuildRequest reference in v4 upgrade guide

sean-:
See #977 

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -168,7 +168,23 @@ GetRequest() (*http.Request, error)
 GetRequest(method string) (*http.Request, error)
 ```
 
-This change is invisible to almost all callers: the typed `Req` structs that the client consumes (e.g. `opensearchapi.SearchReq`, the v5-preview `opensearchapi.IndexReq`) already implement the new signature. Only code that defines a custom type satisfying `opensearch.Request` is affected. If you maintain such a type, add a `method string` parameter and forward it to your underlying `http.NewRequest` call (or `opensearch.BuildRequest`).
+This change is invisible to almost all callers: the typed `Req` structs that the client consumes (e.g. `opensearchapi.SearchReq`, the v5-preview `opensearchapi.IndexReq`) already implement the new signature. Only code that defines a custom type satisfying `opensearch.Request` is affected.
+
+If you maintain such a type, add a `method string` parameter and forward it to your request builder. The `opensearch.BuildRequest` helper that earlier v4 releases exposed for this purpose was **removed in 4.7.0**; construct the request with `net/http` directly instead. Pass a relative path -- the transport prepends the base URL.
+
+```go
+// Before (<= 4.6.0): fixed method, built via the removed opensearch.BuildRequest helper
+func (r customReq) GetRequest() (*http.Request, error) {
+    return opensearch.BuildRequest(http.MethodGet, r.path, r.body, nil, nil)
+}
+
+// After (>= 4.7.0): method comes from the caller, built with net/http
+func (r customReq) GetRequest(method string) (*http.Request, error) {
+    return http.NewRequest(method, r.path, r.body)
+}
+```
+
+`opensearch.BuildRequest` also accepted `params map[string]string` and `headers http.Header` arguments. To preserve those, set them on the `*http.Request` after construction: encode params onto `req.URL.RawQuery` (via `url.Values`) and add headers to `req.Header`.
 
 ### Path segment values are percent-encoded
 
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -170,17 +170,27 @@ GetRequest(method string) (*http.Request, error)
 
 This change is invisible to almost all callers: the typed `Req` structs that the client consumes (e.g. `opensearchapi.SearchReq`, the v5-preview `opensearchapi.IndexReq`) already implement the new signature. Only code that defines a custom type satisfying `opensearch.Request` is affected.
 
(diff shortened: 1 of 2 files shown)


Jakob3xD (commented):
### Code review

Found 4 issue(s) — all posted as inline comments.

guides/json.md
```diff
@@ -52,14 +52,16 @@ The `Client.Do()` method accepts `any` for its response parameter, which means p
...
+	// opensearch.Request interface. The transport prepends the base URL.
+	// method is an HTTP method (e.g. http.MethodGet) forwarded by the caller.
 	type customReq struct {
 		path string
 		body io.Reader
 	}
 
 	func (r customReq) GetRequest(method string) (*http.Request, error) {
-		return opensearch.BuildRequest(method, r.path, r.body, nil, nil)
+		return http.NewRequest(method, r.path, r.body)
```
Jakob3xD: `http.NewRequest` does not set `Content-Type`, but `opensearch.BuildRequest` set `Content-Type: application/json` automatically whenever `body != nil` (see `request_builder.go` at v4.6.0). The PUT/POST examples in this same file (lines 154, 186) set it manually after `http.NewRequest`, but this migrated `customReq` example drops it. A custom request carrying a JSON body will be sent without `Content-Type`, and OpenSearch typically rejects it with 400/415. Suggest setting `req.Header.Set("Content-Type", "application/json")` when the body is non-nil (or a note pointing to it), to stay consistent with the other examples in this guide.

UPGRADING.md
```diff
@@ -168,7 +168,23 @@ GetRequest() (*http.Request, error)
...
+    return opensearch.BuildRequest(http.MethodGet, r.path, r.body, nil, nil)
+}
+
+// After (>= 4.7.0): method comes from the caller, built with net/http
+func (r customReq) GetRequest(method string) (*http.Request, error) {
+    return http.NewRequest(method, r.path, r.body)
+}
+```
+
+`opensearch.BuildRequest` also accepted `params map[string]string` and `headers http.Header` arguments. To preserve those, set them on the `*http.Request` after construction: encode params onto `req.URL.RawQuery` (via `url.Values`) and add headers to `req.Header`.
```
Jakob3xD, line 197: This note covers params and headers to re-add, but omits the one piece of hidden behavior `BuildRequest` had: it set `Content-Type: application/json` automatically when `body != nil` (verified in `request_builder.go` at v4.6.0). `http.NewRequest` does not. So a migrated POST/PUT custom request silently loses `Content-Type` and OpenSearch commonly rejects it with 400/415. Suggest adding: if the request has a body, also call `req.Header.Set("Content-Type", "application/json")`.

UPGRADING.md
```diff
@@ -168,7 +168,23 @@ GetRequest() (*http.Request, error)
 GetRequest(method string) (*http.Request, error)
 ```
 
-This change is invisible to almost all callers: the typed `Req` structs that the client consumes (e.g. `opensearchapi.SearchReq`, the v5-preview `opensearchapi.IndexReq`) already implement the new signature. Only code that defines a custom type satisfying `opensearch.Request` is affected. If you maintain such a type, add a `method string` parameter and forward it to your underlying `http.NewRequest` call (or `opensearch.BuildRequest`).
+This change is invisible to almost all callers: the typed `Req` structs that the client consumes (e.g. `opensearchapi.SearchReq`, the v5-preview `opensearchapi.IndexReq`) already implement the new signature. Only code that defines a custom type satisfying `opensearch.Request` is affected.
+
+If you maintain such a type, add a `method string` parameter and forward it to your request builder. The `opensearch.BuildRequest` helper that earlier v4 releases exposed for this purpose was **removed in 4.7.0**; construct the request with `net/http` directly instead. Pass a relative path -- the transport prepends the base URL.
```
Jakob3xD: Minor clarity: the transport builds the final URL via `req.URL.Path = base + req.URL.Path` (plain string concatenation in `opensearchtransport.setReqURL`, not `url.ResolveReference`), so the path must begin with a leading `/` (e.g. `/_plugins/my_plugin/status`). "relative path" can be misread as an RFC-relative reference. Consider showing a concrete leading-slash path so readers don't accidentally build a malformed URL.

UPGRADING.md
```diff
@@ -168,7 +168,23 @@ GetRequest() (*http.Request, error)
...
+
+```go
+// Before (<= 4.6.0): fixed method, built via the removed opensearch.BuildRequest helper
+func (r customReq) GetRequest() (*http.Request, error) {
+    return opensearch.BuildRequest(http.MethodGet, r.path, r.body, nil, nil)
```
Jakob3xD: Optional nit: this "Before" example hardcodes `http.MethodGet` and drops the method field, whereas the actual v4.6.0 `customReq` in guides/json.md stored the method in a struct field and passed `r.method`. The `BuildRequest` signature and the semantic point (the old interface had no caller-supplied method) are correct; just noting the illustration differs from the real prior code shape.

Jakob3xD (approved):
### Code review

All four findings from the previous review are resolved in `0cacc88`, and the follow-up introduces no new issues:

1. Content-Type — the migration examples in both `UPGRADING.md` and `guides/json.md` now set `Content-Type: application/json` for a non-nil body, faithfully reproducing what `opensearch.BuildRequest` did, with a clear explanatory comment. The added `http.NewRequest` error handling is a nice touch.
2. Path guidance — reworded to require a leading slash and explain the string-concatenation reason a missing one produces a malformed URL.
3. Before example — now uses `r.method` (matching the real v4.6.0 struct-field shape) and notes the old Content-Type behavior.

LGTM.

PR #979, opened 2026-07-17
docs: fix removed opensearch.BuildRequest reference and sweep upgrade guides

sean-:
The v4.7.0 migration section (and the raw-JSON usage guide) told readers to forward a custom Request.GetRequest to opensearch.BuildRequest, but that helper was removed in 4.7.0 -- the very release the section documents. Replace it with a net/http-based before/after example (relative path; the transport prepends the base URL).

A verification sweep of the remaining upgrade guides fixed:
- UPGRADING_V5.md: RawBody() comment wrongly listed error responses as nil.
- opensearchapi/UPGRADING_V4_TO_V5.md: SearchParams.Size is *int; the example used a bare int literal that would not compile.
- UPGRADING_V3.md: InsecureSkipVerify did not exist in v2.3.0/v3.0.0; error snippet had a variable-name mismatch and a missing return; Msearch/ MsearchTemplate casing in the v2.3.0 before-column.
- opensearchapi/UPGRADING_V3_TO_V4.md: scoped the CausedBy addition to v4.6.0.

Reported in #977.

### DIFF
--- a/UPGRADING_V4.md
+++ b/UPGRADING_V4.md
@@ -16,21 +16,33 @@ GetRequest(method string) (*http.Request, error)
 
 This change is invisible to almost all callers: the typed `Req` structs that the client consumes (e.g. `opensearchapi.SearchReq`, `opensearchapi.IndexReq`) already implement the new signature. Only code that defines a custom type satisfying `opensearch.Request` is affected.
 
-If you maintain such a type, add a `method string` parameter and forward it to your request builder. The `opensearch.BuildRequest` helper that earlier v4 releases exposed for this purpose was **removed in 4.7.0**; construct the request with `net/http` directly instead. Pass a relative path -- the transport prepends the base URL.
+If you maintain such a type, add a `method string` parameter and forward it to your request builder. The `opensearch.BuildRequest` helper that earlier v4 releases exposed for this purpose was **removed in 4.7.0**; construct the request with `net/http` directly instead.
 
 ```go
-// Before (<= 4.6.0): fixed method, built via the removed opensearch.BuildRequest helper
+// Before (<= 4.6.0): method stored on the struct, built via the removed
+// opensearch.BuildRequest helper (which set Content-Type for a non-nil body).
 func (r customReq) GetRequest() (*http.Request, error) {
-    return opensearch.BuildRequest(http.MethodGet, r.path, r.body, nil, nil)
+    return opensearch.BuildRequest(r.method, r.path, r.body, nil, nil)
 }
 
-// After (>= 4.7.0): method comes from the caller, built with net/http
+// After (>= 4.7.0): method comes from the caller, built with net/http.
 func (r customReq) GetRequest(method string) (*http.Request, error) {
-    return http.NewRequest(method, r.path, r.body)
+    req, err := http.NewRequest(method, r.path, r.body)
+    if err != nil {
+        return nil, err
+    }
+    // BuildRequest set this automatically for a non-nil body; http.NewRequest
+    // does not, so set it here or OpenSearch may reject a JSON body with 400/415.
+    if r.body != nil {
+        req.Header.Set("Content-Type", "application/json")
+    }
+    return req, nil
 }
 ```
 
 `opensearch.BuildRequest` also accepted `params map[string]string` and `headers http.Header` arguments. To preserve those, set them on the `*http.Request` after construction: encode params onto `req.URL.RawQuery` (via `url.Values`) and add headers to `req.Header`.
+
+> The path must begin with a leading slash (e.g. `r.path == "/_plugins/my_plugin/status"`). The transport builds the final URL by concatenating the base URL with the request path (`base + req.URL.Path` in `opensearchtransport.setReqURL`), not via `url.ResolveReference`, so a path without a leading slash produces a malformed URL.
 
 ### Path segment values are percent-encoded
 
--- a/UPGRADING_V4.md
+++ b/UPGRADING_V4.md
@@ -14,7 +14,23 @@ GetRequest() (*http.Request, error)
 GetRequest(method string) (*http.Request, error)
 ```
 
-This change is invisible to almost all callers: the typed `Req` structs that the client consumes (e.g. `opensearchapi.SearchReq`, `opensearchapi.IndexReq`) already implement the new signature. Only code that defines a custom type satisfying `opensearch.Request` is affected. If you maintain such a type, add a `method string` parameter and forward it to your underlying `http.NewRequest` call (or `opensearch.BuildRequest`).
+This change is invisible to almost all callers: the typed `Req` structs that the client consumes (e.g. `opensearchapi.SearchReq`, `opensearchapi.IndexReq`) already implement the new signature. Only code that defines a custom type satisfying `opensearch.Request` is affected.
+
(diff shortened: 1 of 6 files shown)


guides/usage-json.md
```diff
@@ -52,14 +52,16 @@ When you need to call an API that `opensearchapi` doesn't cover -- plugin endpoi
...
+	// opensearch.Request interface. The transport prepends the base URL.
+	// method is an HTTP method (e.g. http.MethodGet) forwarded by the caller.
 	type customReq struct {
 		path string
 		body io.Reader
 	}
 
 	func (r customReq) GetRequest(method string) (*http.Request, error) {
-		return opensearch.BuildRequest(method, r.path, r.body, nil, nil)
+		return http.NewRequest(method, r.path, r.body)
```
Jakob3xD: `http.NewRequest` does not set `Content-Type`, but `opensearch.BuildRequest` set `Content-Type: application/json` automatically whenever `body != nil` (see `request_builder.go` at v4.6.0). The PUT/POST examples in this same file (lines 154, 186) set it manually after `http.NewRequest`, but this migrated `customReq` example drops it. A custom request carrying a JSON body will be sent without `Content-Type`, and OpenSearch typically rejects it with 400/415. Suggest setting `req.Header.Set("Content-Type", "application/json")` when the body is non-nil (or a note pointing to it), to stay consistent with the other examples in this guide. (Same point was raised on the sibling PR #978.)

UPGRADING_V4.md
```diff
@@ -14,7 +14,23 @@ GetRequest() (*http.Request, error)
...
+    return opensearch.BuildRequest(http.MethodGet, r.path, r.body, nil, nil)
+}
+
+// After (>= 4.7.0): method comes from the caller, built with net/http
+func (r customReq) GetRequest(method string) (*http.Request, error) {
+    return http.NewRequest(method, r.path, r.body)
+}
+```
+
+`opensearch.BuildRequest` also accepted `params map[string]string` and `headers http.Header` arguments. To preserve those, set them on the `*http.Request` after construction: encode params onto `req.URL.RawQuery` (via `url.Values`) and add headers to `req.Header`.
```
Jakob3xD, line 43: This note covers params and headers to re-add, but omits the one piece of hidden behavior `BuildRequest` had: it set `Content-Type: application/json` automatically when `body != nil` (verified in `request_builder.go` at v4.6.0). `http.NewRequest` does not. So a migrated POST/PUT custom request silently loses `Content-Type` and OpenSearch commonly rejects it with 400/415. Suggest adding: if the request has a body, also call `req.Header.Set("Content-Type", "application/json")`.

opensearchapi/UPGRADING_V4_TO_V5.md
```diff
@@ -67,12 +67,14 @@ client.Search(ctx, &opensearchapi.SearchReq{
 // v5
 client.Search(ctx, &opensearchapi.SearchReq{
     Indices: []string{"products"},
-    Params:  &opensearchapi.SearchParams{Size: 20},
+    Params:  &opensearchapi.SearchParams{Size: new(20)},
```
Jakob3xD: Two related issues here:

1. Asymmetric fix: this `// v5` block is corrected to `Size: new(20)`, but the `// v4` block just above (line 64) still reads `Params: opensearchapi.SearchParams{Size: 20}`. `SearchParams.Size` has been `*int` since v4.0.0 (`git show v4.0.0:opensearchapi/api_search-params.go`), so `Size: 20` is an identical compile error there too — it should become `Size: opensearch.ToPointer(20)` (or `new(20)` on Go 1.26+).

2. The note below ("`Size` is also now a `*int` ... the same pointerization applies to optional scalars") is misleading: `Size` was already `*int` throughout v4, so it was not pointerized as part of the v4→v5 migration. The `*bool` pointerization is a genuine v5 change, but `Size` is not — consider dropping `Size` from that grouping or rewording.

UPGRADING_V4.md
```diff
@@ -14,7 +14,23 @@ GetRequest() (*http.Request, error)
 GetRequest(method string) (*http.Request, error)
 ```
 
-This change is invisible to almost all callers: the typed `Req` structs that the client consumes (e.g. `opensearchapi.SearchReq`, `opensearchapi.IndexReq`) already implement the new signature. Only code that defines a custom type satisfying `opensearch.Request` is affected. If you maintain such a type, add a `method string` parameter and forward it to your underlying `http.NewRequest` call (or `opensearch.BuildRequest`).
+This change is invisible to almost all callers: the typed `Req` structs that the client consumes (e.g. `opensearchapi.SearchReq`, `opensearchapi.IndexReq`) already implement the new signature. Only code that defines a custom type satisfying `opensearch.Request` is affected.
+
+If you maintain such a type, add a `method string` parameter and forward it to your request builder. The `opensearch.BuildRequest` helper that earlier v4 releases exposed for this purpose was **removed in 4.7.0**; construct the request with `net/http` directly instead. Pass a relative path -- the transport prepends the base URL.
```
Jakob3xD: Minor clarity: the transport builds the final URL via `req.URL.Path = base + req.URL.Path` (plain string concatenation in `opensearchtransport.setReqURL`, not `url.ResolveReference`), so the path must begin with a leading `/` (e.g. `/_plugins/my_plugin/status`). "relative path" can be misread as an RFC-relative reference. Consider showing a concrete leading-slash path so readers don't accidentally build a malformed URL.

UPGRADING_V4.md
```diff
@@ -14,7 +14,23 @@ GetRequest() (*http.Request, error)
...
+
+```go
+// Before (<= 4.6.0): fixed method, built via the removed opensearch.BuildRequest helper
+func (r customReq) GetRequest() (*http.Request, error) {
+    return opensearch.BuildRequest(http.MethodGet, r.path, r.body, nil, nil)
```
Jakob3xD: Optional nit: this "Before" example hardcodes `http.MethodGet` and drops the method field, whereas the actual v4.6.0 `customReq` in the guide stored the method in a struct field and passed `r.method`. The `BuildRequest` signature and the semantic point (the old interface had no caller-supplied method) are correct; just noting the illustration differs from the real prior code shape.

Jakob3xD (commented):
### Code review

Found 5 issue(s) — all posted as inline comments.

The sweep is largely accurate and fixes several real doc bugs (the `InsecureSkipVerify` removal for v2/v3, the `resp`/`return fmt.Errorf` fix, the `Msearch` casing, the `RawBody()` nil condition, `CausedBy` scoped to v4.6.0, and the `new(20)`/`ToPointer` note). The comments below are the remaining gaps.

Jakob3xD (approved):
### Code review

All five findings from the previous review are resolved in `7e7a44b`, and the follow-up introduces no new issues:

1. Content-Type — the migration examples in both `UPGRADING_V4.md` and `guides/usage-json.md` now set `Content-Type: application/json` for a non-nil body (with error handling and an explanatory comment), matching the old `opensearch.BuildRequest` behavior.
2. Size example (`UPGRADING_V4_TO_V5.md`) — the v4 and v5 blocks are unified on `opensearch.ToPointer(20)` (verified available since v4.0.0), and the misleading prose is replaced with the accurate note that `Size` has been `*int` since v4.0.0 and only `Params` became a pointer. The `new(20)` form is now presented as an optional Go 1.26+ alternative.
3. Path guidance — rewritten as a dedicated note requiring a leading slash and citing the `base + req.URL.Path` concatenation in `opensearchtransport.setReqURL`.
4. Before example — now uses `r.method` (matching the real v4.6.0 struct-field shape) and documents the old Content-Type behavior.

The rest of the sweep (InsecureSkipVerify, resp/return fix, Msearch casing, RawBody nil condition, CausedBy scoping) was already accurate. LGTM.

PR #981, opened 2026-07-17
Fix/pool field lock races

sean-:
CI picked up this race today and ran it to ground.  This bug has been present since 7f9970c3 (for ~4mo) and needs to be backported to the v4 branch, too.


Jakob3xD (approved):
### Code review

LGTM. Verified the fix achieves its stated goal:

- `snapshot()` now reads `activeListCap` while holding `RLock` instead of after `RUnlock` — this is the actual race fix.
- Moving the fields (`activeListCap`, `warmupRounds`, `warmupSkipCount`, `healthCheck`) into the `mu` sub-struct makes the guard structural.
- Policy-pool `recalculateWarmupParams`/`getWarmupParams` are correctly serialized under `p.pool.Lock()` (nicely documented in `policy_role.go`).
- Flat-pool `snapshot()` in `metrics.go` is serialized against the discovery writer via `c.mu`.
- Test call sites and CHANGELOG updated consistently.

No issues introduced by this change.

---

Non-blocking, pre-existing (not introduced by this PR): in `createOrUpdateMultiNodePoolWithLock`, after the pool lock is released, `recalculateWarmupParams` (write) and `getWarmupParams` (read) — plus the `mu.activeCount` write a few lines down — touch `mu`-guarded fields without holding `pool.mu`. Those run under `c.mu(W)`, which serializes them against `metrics.snapshot()` (`c.mu.RLock`), but not against `resurrectWithLock`, which reads the same fields under `pool.mu` only. The rename `cp.x` → `cp.mu.x` is race-neutral, so this is unchanged from before; the `activeCount` write (never touched here) has the same shape. Worth a follow-up if you want the "all access under `pool.mu`" invariant to hold end-to-end.

https://github.com/opensearch-project/opensearch-go/blob/4eb7d4e5dfb6b122030f291a3b2f9f404ea84c96/opensearchtransport/discovery.go#L1047-L1089

PR #989, opened 2026-07-20
feat(osgen,opensearchapi): export and document sub-client types

ryanyuan:
### Summary

Exports the generated opensearchapi and plugin sub-client types so each gets its own godoc page, adds a map of which sub-client owns which operations, and fixes a set of `osgen` codegen bugs.

### Changes

**Export sub-client types.** `documentClient`, `catClient`, plugin types like `roleClient`, etc. were unexported, so godoc rendered no page and their methods were only findable by grepping the generated files. Renames them to exported names (`DocumentClient`, ...) so each gets a godoc page with a navigable method list. The `client` field stays unexported, so external construction is still impossible and only `NewClient` wires a usable instance. Additive and non-breaking: field access (`client.Doc.Get`) is unchanged; alias fields (`Document`, `Indices`, ...) keep their names and share the canonical type.

**Sub-client doc map.** Adds a hand-written "Sub-clients" section to `opensearchapi/doc.go` mapping field path → sub-client type → API family, with doc links to each godoc page. Teaches `osgen` to generate the equivalent package doc for plugins that have sub-clients.

**Codegen fixes.**
- `IP2Geo` acronym casing (`Ip2geoDatasource` → `IP2GeoDatasource`).
- `SearchProfile` type-name collision that silently dropped a type and mis-typed the response `Profile` field. Adds an override table; container becomes `SearchProfileResult`, no existing public name changes.
- `<Group>Resp` response-body collisions that degraded dropped types to raw `json.RawMessage` (flow_framework, security_analytics). Adds a response-body override table merged into one ref → name lookup.
- Nil `Inspect().Response` on plugin dispatch (template discarded the transport response).


sean- (dismissed):

Jakob3xD (dismissed):
### Code review

LGTM. Verified the four codegen fixes are correct (spot-checked generator logic and generated output):

- IP2Geo acronym: added to the acronyms table and applied only at PascalCase boundaries, so no collateral over-matching; generated geospatial types read `IP2GeoDatasource`.
- SearchProfile and response-body collisions (flow_framework, security_analytics): resolved via the override tables, with the completeness panic-guard ensuring any new colliding ref must be enumerated.
- nil `Inspect().Response` on plugin dispatch: the template now wires `resp.response, err = request(...)` in all four dispatch paths (root ptr/value, sub-client ptr/value) instead of discarding it.
- Type export is additive/non-breaking; the `client`/`apiClient` field stays unexported, and the generated godoc is accurate.

Two non-blocking nits — stale example comments in the generator source that the rename sweep missed (the user-facing generated godoc is unaffected):

1. `SubClient` struct examples still show the old unexported names (`"catClient"`, `"indicesClient"`); should be `"CatClient"` / `"IndicesClient"`.

https://github.com/opensearch-project/opensearch-go/blob/cec26db36554fcb3bf419e187b8dc7600c23d70b/cmd/osgen/emit/frag_clients.go#L15-L20

2. The plugin sub-client comments read `// unexported type name (e.g. "actionGroupClient")`, which is now inaccurate — plugin sub-client types are exported (`ActionGroupClient`). `resourceToTypeName`'s own doc was updated to match; these two were missed.

https://github.com/opensearch-project/opensearch-go/blob/cec26db36554fcb3bf419e187b8dc7600c23d70b/cmd/osgen/emit/frag_plugin.go#L18-L22

https://github.com/opensearch-project/opensearch-go/blob/cec26db36554fcb3bf419e187b8dc7600c23d70b/cmd/osgen/route.go#L573-L576

sean- (approved):

sean-:
Thank you!  This is much easier to consume now: https://pkg.go.dev/github.com/opensearch-project/opensearch-go/v5@v5.0.0-rc2.0.20260720173644-20e0e1fd8a77/opensearchapi#DocumentClient

PR #998, opened 2026-07-21
feat(osgen)!: emit oneOf-of-const schemas as permissive typed string enums

sean-:
## Summary

osgen collapsed enum-like `oneOf`-of-`{type: string, const: X}` schemas (`NodeRole`, `Result`, `VersionType`, `TimeUnit`, …) to a plain Go `string`, so the individual wire values were never surfaced as typed constants.

This change emits each such schema as a **named string type with one exported const per value**:

```go
type NodeRole string

const (
      NodeRoleData   NodeRole = "data"
      NodeRoleSearch NodeRole = "search"
      NodeRoleWarm   NodeRole = "warm"
      // ...
)
```

It also bundles the `search` and `warm` node roles from apispec [#1006](https://github.com/opensearch-project/opensearch-api-specification/pull/1006) (merged upstream) into the vendored spec, which was the original driver.

## Permissive by design

The generated type is **permissive**, not closed. Because it is backed by `string`, `encoding/json` round-trips any value: an unknown role a newer server or a plugin introduces decodes cleanly rather than erroring. No custom `(Un)MarshalJSON` is generated.

This is deliberately different from the existing closed int-backed enum (`RestStatus`), whose `UnmarshalJSON` rejects unknown values and aborts the enclosing decode. Node roles are pluggable/dynamic on the server, so a closed enum would turn "cluster returns an unknown role" into a hard decode failure on `GET _nodes` / `_cat/nodes`.

## Breaking change

> [!WARNING]
> This is API-breaking. Fields and query params across `opensearchapi` and `plugins` change from `string` to named enum types.

- `Roles []string` → `Roles []NodeRole`
- `Result string` → `Result Result`
- `Time string` → `Time TimeUnit`, and similar for `OpType`, `SearchType`, `VersionType`, etc.

Values still assign and compare as strings via the named type. Only type-strict comparisons against untyped string literals need the const:

```go
// before
require.Equal(t, "updated", resp.Result)
// after
require.Equal(t, opensearchapi.ResultUpdated, resp.Result)
```

The one such hand-written assertion in the suite (`api_document_lifecycle_test.go`) is updated.

## Types emitted

33 string-enum types now generated (`NodeRole`, `Result`, `VersionType`, `TimeUnit`, `SortMode`, `Conflicts`, `OpType`, `SearchType`, …), alongside the unchanged closed `RestStatus` int enum.


cmd/osgen/ir_bridge.go
```diff
@@ -302,21 +305,25 @@ func convertType(gt *goType) *ir.Type {
...
 // It panics if a value yields an invalid Go identifier (e.g. an empty value, or
 // one whose segments produce nothing) or if two distinct values collapse to the
 // same const name (e.g. "FOO_BAR" and "FOO__BAR"). Both would otherwise emit
 // uncompilable Go — failing loudly at generation time matches osgen's handling
-// of other invalid-identifier cases (see naming.go).
-func convertEnumMembers(typeName string, values []string) []ir.EnumMember {
+// of other invalid-identifier cases (see naming.go). The string-enum path
+// pre-validates and deduplicates its values in parseConstOneOf, so those guards
+// only ever fire for the int-backed (x-enum-name) path.
+func convertEnumMembers(typeName string, values, descriptions []string) []ir.EnumMember {
```
Jakob3xD: This comment is inaccurate for the sentinel guard. The doc says the panic guards "only ever fire for the int-backed (x-enum-name) path" -- but convertEnumMembers seeds the dedup map with typeName+"Unknown" unconditionally (a few lines below). That seed is only needed by the int-backed path (which emits the <Type>Unknown sentinel), not the string-backed path. A const value "unknown" passes parseConstOneOf cleanly (it starts with an empty seen map, not mirroring this reservation) and then panics when it hits "<unknown sentinel>" in the seed. See main review body for details and suggested fix.

cmd/osgen/string_enum.go
```diff
@@ -0,0 +1,142 @@
...
+		raw = append(raw, constEnumValue{Value: cv, Description: s.Description})
+	}
+	if !sawConst {
+		return nil, false
+	}
+
+	// Reduce to values with a valid, unique const identifier, keeping the first
+	// branch to win each identifier. seen maps const name -> the value that took
+	// it, so a later distinct value colliding on the same name is reported.
+	seen := make(map[string]string, len(raw))
```
Jakob3xD, line 134: seen here is not seeded with typeName+"Unknown", but convertEnumMembers is. This means a const value "unknown" (PascalCases to "Unknown", constName "TypeNameUnknown") passes through parseConstOneOf unscathed and then hits the typeName+"Unknown": "<unknown sentinel>" seed in convertEnumMembers, causing a generation-time panic. Either seed the same reserved names here (worse for the permissive design) or make the typeName+"Unknown" seed in convertEnumMembers conditional on the int-backed path (recommended -- for the string path, FooUnknown is a desirable const, not a reserved name).

Jakob3xD (commented):
Two non-blocking findings from a historical-context review (looking at how prior commits/PRs set invariants that this PR's changes could interact with).

**Finding 1 – latent generation-time panic for `const: "unknown"` values (non-blocking, currently safe)**

`convertEnumMembers` seeds its dedup map with `typeName+"Unknown"` (value `"<unknown sentinel>"`). That sentinel exists because the int-backed path emits a `<Type>Unknown` sentinel + closed `UnmarshalJSON` that rejects it. The string-backed path emits no such sentinel, but the seed is unconditional.

`parseConstOneOf` starts its dedup map empty — it does not mirror the `typeName+"Unknown"` reservation. So a const branch `{type: string, const: "unknown"}` survives `parseConstOneOf` cleanly (identifier `Unknown`, constName `FooUnknown`), then hits the seed in `convertEnumMembers` and panics:

```
convertEnumMembers: enum values "<unknown sentinel>" and "unknown" on type "Foo" both map to const "FooUnknown"
```

No current enum in the bundled spec has this value, so generation succeeds today. But it contradicts the PR's headline design goal of being permissive as the spec evolves — the value literally named `"unknown"` would trigger a hard generation-time panic rather than round-tripping transparently.

The comment added to `convertEnumMembers` says those guards "only ever fire for the int-backed path" — that claim is wrong for the sentinel guard specifically.

Cleanest fix: make the `typeName+"Unknown"` seed conditional on the enum kind. It is only needed when the int-backed `<Type>Unknown` sentinel will actually be emitted. For the string-backed path, `FooUnknown` is a perfectly valid, desirable const name that should be allowed.

**Finding 2 – stale `rawmessage_allowlist.txt` entry (cosmetic, non-blocking)**

`cmd/osgen/rawmessage_allowlist.txt` line 60 (`ClusterGetDecommissionAwarenessResp/[entries] # map`) was added by PR #890 (commit `869a579b`) because `ClusterGetDecommissionAwarenessResp.Entries` was unresolved and fell back to `map[string]json.RawMessage`. This PR resolves the element type to `ClusterDecommissionAwarenessDecommissionStatus`, so the entry is now phantom. The guard treats stale entries as a NOTE-only warning (not a failure), but re-running with `-update-raw-message-allowlist` is the maintenance expectation established in PR #890.

opensearchapi/unions_gen.go
```diff
@@ -7277,6 +7277,117 @@ func (u SearchInnerHitsFieldsItem) MarshalJSON() ([]byte, error) {
...
+
+func (u *SearchHighlightBaseType) UnmarshalJSON(data []byte) error {
+	u.raw = data
+	u.value = nil
+	u.typ = SearchHighlightBaseTypeUnknownType
+	if len(data) == 0 || bytes.Equal(data, build.NullJSON) {
+		return nil
+	}
+	switch {
+	case data[0] == '{':
```
Jakob3xD: Dead `Builtin` branch — wrong JSON token dispatch. `SearchBuiltinHighlighterType` is a string enum (`type ... string`, values `"plain"`/`"fvh"`/`"unified"`), so its wire form is a quoted string starting with `"`. But this branch is gated on `data[0] == '{'` (object). A valid builtin value can never reach here — it always falls to the `data[0] == '"'` `String` branch, so `Type()` never returns `SearchHighlightBaseTypeBuiltinType` and `Builtin()` always returns the zero value.

Root cause is in the generator: `classifyRefBranch` / `tokenClassForSchemaValue` (cmd/osgen/resp_union.go) derives the dispatch token from the raw schema, and a const-oneOf schema has `type: nil` → it defaults to `"object"`. It's unaware that this PR's `resolveStringEnumConst` now retypes the same schema to a string enum. Fix: when the resolved branch type is a string enum (`IsStringEnum`), classify it as the `"string"` token (`data[0] == '"'`).

opensearchapi/unions_gen.go
```diff
@@ -19728,6 +19839,117 @@ func (u IndicesIndexSegmentSortOrder) MarshalJSON() ([]byte, error) {
...
+
+func (u *IndicesIndexSettingsStoreType) UnmarshalJSON(data []byte) error {
+	u.raw = data
+	u.value = nil
+	u.typ = IndicesIndexSettingsStoreTypeUnknownType
+	if len(data) == 0 || bytes.Equal(data, build.NullJSON) {
+		return nil
+	}
+	switch {
+	case data[0] == '{':
```
Jakob3xD: Same wrong-token dispatch as `SearchHighlightBaseType`. `IndicesBuiltinStorageType` is a string enum (`"fs"`/`"hybridfs"`/`"mmapfs"`/`"niofs"`), but the `Builtin` branch is gated on `data[0] == '{'`, so it is unreachable and every value decodes into the `String` branch. This union has a live decode path: `store.type` is returned by GET `_settings`, so `Type()`/`Builtin()` are observably broken for real responses. Same generator fix applies (classify string-enum branches as the `"string"` token).

cmd/osgen/string_enum_test.go
```diff
@@ -0,0 +1,198 @@
...
+		schema := &openapi3.Schema{OneOf: openapi3.SchemaRefs{
+			{Value: openapi3.NewStringSchema()},
+			{Value: openapi3.NewIntegerSchema()},
+		}}
+		_, ok := parseConstOneOf("X", "x___X", schema, VersionRange{}, nil)
+		require.False(t, ok)
+	})
+
+	t.Run("honors the deny-list", func(t *testing.T) {
+		t.Parallel()
```
Jakob3xD, line 119: Data race. This `t.Parallel()` subtest mutates the package-global `stringEnumDenyList` map (`stringEnumDenyList[key] = struct{}{}` a few lines below, plus the `defer delete`). Its sibling subtests are also `t.Parallel()` and call `parseConstOneOf`, which reads `stringEnumDenyList` at its first line (cmd/osgen/string_enum.go:68). Parallel subtests resume concurrently after the parent returns, so this is a concurrent map write + read with no synchronization. Go's runtime raises the always-on "concurrent map read and map write" fatal (not only under `-race`), so the test binary can crash nondeterministically; CI's `-race` will also flag it. Fix: drop `t.Parallel()` on this subtest only (it mutates global state), or swap in a fresh test-scoped map instead of writing the global.

opensearchapi/enums_gen.go
```diff
@@ -210,3 +210,537 @@ type UnknownRestStatusError struct {
...
+	// The node can act as a remote cluster client.
+	NodeRoleRemoteClusterClient NodeRole = "remote_cluster_client"
+
+	// The node can perform transform operations.
+	NodeRoleTransform NodeRole = "transform"
+
+	// The node can only vote in cluster decisions.
+	NodeRoleVotingOnly NodeRole = "voting_only"
+
+	NodeRoleMaster NodeRole = "master"
```
Jakob3xD, line 592: Deprecation metadata is dropped. The spec marks `master` deprecated (`deprecated: true`, `x-version-deprecated: '2.0'`, `x-deprecation-message: Use \`cluster_manager\` instead.`), but no `// Deprecated:` godoc is emitted here, so users get no signal to prefer `NodeRoleClusterManager`. `parseConstOneOf` only carries each branch's `Description`, and this branch has none. Non-blocking, but consider reading `x-deprecation-message` / `x-version-deprecated` into the member doc comment (as `AvailabilityNote` does elsewhere).

Jakob3xD (commented):
### Code review

Note: this is a second review on the PR. An earlier automated review (the latent `const:"unknown"` panic + stale `rawmessage_allowlist.txt` entry) was submitted before these findings were ready; the four below are additional.

Found 4 issue(s) — 3 posted as inline comments, 1 below.

4. `buildIntegParams` has no `IsStringEnum` branch (latent, non-blocking). A *required* string-enum query param falls through to the `default` case, emitting `<Field>: name` where `name` is a plain `string` variable. Assigning a `string` var to a named `type X string` field does not compile (unlike an untyped string literal). No required string-enum query param exists in the bundled spec today, so it is not triggered now, but it diverges from `frag_params.go`, which casts string-enum params. A `case p.IsStringEnum:` branch emitting a converted value would match that handling.

https://github.com/opensearch-project/opensearch-go/blob/e511625010422389662c6c421103043ed689221c/cmd/osgen/emit/build.go#L1288-L1306

cmd/osgen/ir_bridge.go
sean-: Yup.  Thank you for tracing it to the unconditional seed rather than just the symptom. You're right that the "only fires for the int-backed path" comment was wrong for the sentinel guard specifically.

Fixed as you recommended: the `<Type>Unknown` reservation is now conditional on the emitting path rather than unconditional. `convertEnumMembers` takes an explicit `reserved` set of the identifiers a fragment emits in the same scope; the int-backed path passes `{<Type>Unknown}` (it emits that sentinel), the string-backed path passes nothing. So a const value `"unknown"` now yields a valid `FooUnknown` const and round-trips, matching the permissive design. Corrected the misleading doc comment too.

Added a regression test asserting the string path accepts a `"unknown"` value without panicking.

opensearchapi/unions_gen.go
sean-: Confirmed and fixed, though the fix landed somewhere different from the token-class tweak once I worked through it.

Your root-cause was right: the branch resolved to a string enum but dispatched on `{`. But reclassifying it as the `"string"` token just moved the bug: both branches are then JSON strings, so the generator fell back to try-each, where the permissive enum (which accepts *any* string) always wins and the plain-string branch goes dead instead. These two branches are genuinely indistinguishable on the wire.

Since the permissive enum already makes unknown/custom values first-class, the `oneOf[enum, string]` is saying the same thing twice — the custom-string branch only existed to express "a value outside the known set," which the enum now covers. So these unions now collapse to the enum type itself:

- `SearchHighlight.Type` → `*SearchBuiltinHighlighterType`
- `IndicesIndexSettingsStore.Type` → `IndicesBuiltinStorageType`

The union types are no longer emitted. Round-trip stays lossless, and as a bonus the field is now a named enum the `exhaustive` linter can check at `switch` sites. Implemented as a new branch-collapse rule alongside the
existing `collapseEquivalentBranches`.

cmd/osgen/string_enum_test.go
sean-: Confirmed and fixed. Rather than serialize the subtest, I removed the shared-mutable-state root cause: the deny-list is now threaded through `parseConstOneOf` as a parameter instead of read from the package global.
Production passes the global as the starting point; the test clones it (`stringEnumDenyList.clone()`) and adds to the copy, so nothing mutates global state. The subtest keeps `t.Parallel()`, and the package passes under
`-race`.

(Also introduced a small generic `set[T]` here to carry this)

opensearchapi/enums_gen.go
sean-: Good call — done. `parseConstOneOf` now carries each branch's `x-version-added` / `x-version-deprecated` / `x-deprecation-message`, and the string-enum fragment renders the same `AvailabilityNote` other fragments use. `master` now emits:

```
    // Deprecated: since 2.0.0. Use `cluster_manager` instead.
    NodeRoleMaster NodeRole = "master"
```

sean-:
Thanks for the review.

All eight points are addressed in the latest push:

- **Union dead-branch (blocking, `IndicesIndexSettingsStoreType` had a live `_settings` decode path):** the two `enum | string` unions now collapse to their enum type, since the permissive enum supersedes the custom-string branch. Fields become `*SearchBuiltinHighlighterType` and `IndicesBuiltinStorageType`.
- **`const:"unknown"` panic:** sentinel reservation made conditional on the emitting path; string enums now accept an `"unknown"` value. Doc comment corrected.
- **Test data race:** deny-list threaded as a parameter; global no longer mutated. Passes under `-race`.
- **`buildIntegParams`:** string-enum case added with the converting cast.
- **`master` deprecation:** version/deprecation metadata now flows into the const doc.
- **Stale allowlist entry:** regenerated; removed.

Verified: `make regen` reproducible + idempotent, `go build`/`vet`/`golangci-lint` clean (osgen + generated, all build-tag sets), unit + integration suites green against a live cluster (including the nodes, index-settings, and search paths these types touch).

Also added CHANGELOG entries (Added + a BREAKING Changed note).
- **Test data race:** deny-list threaded as a parameter; global no longer mutated. Passes under `-race`.
- **`buildIntegParams`:** string-enum case added with the converting cast.
- **`master` deprecation:** version/deprecation metadata now flows into the const doc.
- **Stale allowlist entry:** regenerated; removed.

Verified: `make regen` reproducible + idempotent, `go build`/`vet`/`golangci-lint` clean (osgen + generated, all build-tag sets), unit + integration suites green against a live cluster (including the nodes, index-settings, and search paths these types touch).

ryanyuan (approved):

PR #999, opened 2026-07-21
feat(opensearchtransport,osprom,osotel): request observability

sean-:
## Summary

Enriches the request-observability surface across the transport and both metrics registries, and expands request routing to cover system/admin endpoints. Groups six commits:

1. **`feat(opensearchtransport)`** — add `String()` names for admin `OperationID`s (cat/nodes/tasks/snapshot/script/dangling/data-stream), replacing the numeric `admin_N` fallback with stable snake_case labels.
2. **`feat(opensearchtransport)`** — route system endpoints (cluster/cat/nodes/tasks/snapshot/index-mgmt/mapping/alias/template/data-stream) to the correct node role, adding a cluster-manager routing target.
3. **`feat(opensearchtransport)`** — add `RouteName`, `Index`, and `PoolName` to `RequestEvent`; make `Path` the raw URL-escaped caller input.
4. **`feat(osprom,osotel)`** — pluggable status classifier and buffer-overflow handlers.
5. **`style(opensearchtransport)`** — extract shared operation-name constants.
6. **`style(opensearchtransport)`** — name every `OperationID` `String()` label
   as a constant.

## Motivation

The observability API (currently `v5.0.0-rc3`) exposed only `Method`, `Path`, `Host`, `Attempt`, and `RequestBytes` on `RequestEvent`. `Path` was documented as raw but set from the decoded, post-rewrite `req.URL.Path`; `Index` was documented but never populated; there was no route/operation identity and no way to see which pool served a request. Both metrics registries hardcoded status-code reduction and only incremented a counter on buffer overflow. Doing this before GA keeps the additions non-breaking.

### `RequestEvent` (opensearchtransport)

- **`Path`** is now the URL-escaped caller input (`req.URL.EscapedPath()`), captured before the URL is rewritten to the selected backend, with a corrected doc comment (it is raw and high-cardinality, not templated).
- **`RouteName`** — the classified operation (`"search"`, `"doc_index"`, …), suitable as a low-cardinality metric label; `"other"` when unclassified.
- **`Index`** — extracted from the pre-rewrite path; now actually populated.
- **`PoolName`** — the pool that served the final attempt, threaded from the routing hop through `streamResult`.
- Identity is captured once per request, before the backend rewrite, and only when an observer is registered (the no-observer path stays allocation-free).
- `RouteName` is derived from a process-wide `OperationClassifier`; callers with non-standard paths can override it via the new `Config.OperationClassifier` (affects the label only, never routing).

### Routing (opensearchtransport)

- Adds ~110 routes for system/admin endpoints previously absent from the table (they fell through to the default pool and classified as `other`).
- Introduces a cluster-manager routing target (split read/write pools); cluster-state operations prefer `cluster_manager` nodes with a data-node fallback.
- Node-local reads use the management pool; shard-derived reads use data nodes.
- Because the router and the operation classifier share the route table, these routes also give each operation a meaningful `RouteName`.
- Guards the route trie against empty/relative paths so the new bare `GET /{index}` route cannot match `Clean("") == "."`.

### Metrics registries (osprom, osotel)

- **`WithStatusClassifier`** — override how status codes reduce to the `status` label (default unchanged: `2xx`/`4xx`/`error`).
- **`WithOverflowHandler`** / **`WithStreamOverflowHandler`** — invoked on drop with the queue length and the concrete event, *in addition to* the existing dropped counter, so callers can log overflow without losing the metric.
- Renamed the internal pooled `sample` to `envelope` to disambiguate from the public `RequestSample`.


opensearchtransport/opensearchtransport.go
```diff
@@ -383,6 +383,15 @@ type Config struct {
...
 	Observer  ConnectionObserver // Optional observer for connection lifecycle events
 
+	// OperationClassifier maps each request's method and path to the
+	// [RequestEvent.RouteName] label reported to observers. When nil, a
+	// process-wide classifier built from the standard OpenSearch REST layout is
+	// used. Override it only when you route non-standard paths (e.g. a custom
+	// Router with its own patterns) and need RouteName to reflect them; the
+	// classifier is used for the observability label only and never affects
+	// routing. A supplied classifier must be safe for concurrent use.
+	OperationClassifier *OperationClassifier
```
Jakob3xD, line 393: `OperationClassifier` is a new field on `opensearchtransport.Config`, but it is not threaded through the `opensearch` entry point: `opensearch.Config` has no matching field and `opensearch.NewClient` does not pass it into `opensearchtransport.New(...)` (that block threads `Signer`, `Router`, and `Observer` — the closely-related advanced hooks — but not this one). The CHANGELOG states callers "can override it via the new `Config.OperationClassifier`," yet a user going through the documented `opensearch.NewClient(opensearch.Config{...})` path cannot set it — only a direct `opensearchtransport.New` caller can. This is the same half-threaded config gap flagged and fixed on #902 (DNS knobs). Suggest adding the field to `opensearch.Config` and wiring it in the transport-construction block.

osprom/registry.go
```diff
@@ -111,20 +111,47 @@ type Registry struct {
...
+
 	dropped prometheus.Counter
 }
 
 // Option configures a [Registry].
 type Option func(*options)
 
+// StatusClassifier reduces an HTTP status code to a low-cardinality "status"
+// label. It receives the response status code (0 when no response was received)
+// and returns the label. The default classifier maps to "2xx"/"4xx"/"error";
```
Jakob3xD: The default is broader than "2xx"/"4xx"/"error". `statusClass()` returns `"error"` for code 0, `"unknown"` for out-of-range, and otherwise `strconv.Itoa(code/100) + "xx"` — so `"3xx"` and `"5xx"` are also emitted for real traffic (redirects, server errors). A caller writing a custom classifier off this doc could miss those buckets. Suggest wording like "maps to the per-hundred bucket (`2xx`, `3xx`, `4xx`, `5xx`), `error` (no response), or `unknown`." Same doc + the CHANGELOG's "default `2xx`/`4xx`/`error` unchanged" phrasing, and the parallel `osotel/registry.go:127`.

osprom/registry.go
```diff
@@ -111,20 +111,47 @@ type Registry struct {
...
+// override it with [WithStatusClassifier] to keep or collapse specific codes.
+type StatusClassifier func(statusCode int) string
+
+// RequestOverflowHandler is invoked on the caller's request goroutine each time
+// a buffered ([Registry.OnRequestResponse]) event is dropped because the
+// registry buffer is full. queueLen is the channel occupancy at drop time (equal
```
Jakob3xD: `queueLen` is not guaranteed "equal to the configured buffer size." The `default` branch of the send `select` fires because the channel was full at that instant, but `len(r.ch)` is read afterward and the dispatch goroutine may have drained one or more entries in between, so `queueLen` can be less than capacity under concurrent load. Suggest "(at most the configured buffer size)" to avoid callers relying on it as an exact full-buffer signal. Same wording in `osotel/registry.go:135`.

opensearchtransport/router.go
```diff
@@ -424,6 +448,186 @@ func buildRoleRoutes(r roleRoutes) []Route {
...
+		NewRoute("PUT /_snapshot/{repository}", r.clusterMgrWrite).Op(OpSnapshotRepoCreate).MustBuild(),
+		NewRoute("POST /_snapshot/{repository}", r.clusterMgrWrite).Op(OpSnapshotRepoCreate).MustBuild(),
+		NewRoute("DELETE /_snapshot/{repository}", r.clusterMgrWrite).Op(OpSnapshotRepoDelete).MustBuild(),
+		NewRoute("POST /_snapshot/{repository}/_verify", r.clusterMgrWrite).Op(OpSnapshotRepoVerify).MustBuild(),
+		NewRoute("POST /_snapshot/{repository}/_cleanup", r.clusterMgrWrite).Op(OpSnapshotRepoClean).MustBuild(),
+		NewRoute("GET /_snapshot/{repository}/{snapshot}", r.clusterMgrRead).Op(OpSnapshotGet).MustBuild(),
+		NewRoute("PUT /_snapshot/{repository}/{snapshot}", r.clusterMgrWrite).Op(OpSnapshotCreate).MustBuild(),
+		NewRoute("POST /_snapshot/{repository}/{snapshot}", r.clusterMgrWrite).Op(OpSnapshotCreate).MustBuild(),
+		NewRoute("DELETE /_snapshot/{repository}/{snapshot}", r.clusterMgrWrite).Op(OpSnapshotDelete).MustBuild(),
+		NewRoute("GET /_snapshot/{repository}/{snapshot}/_status", r.clusterMgrRead).Op(OpSnapshotStatus).MustBuild(),
```
Jakob3xD, line 537: The snapshot family is nearly complete but two valid status endpoints are missing: `GET /_snapshot/_status` (all in-progress snapshots) and `GET /_snapshot/{repository}/_status` (all snapshots in a repo). Only the per-snapshot form is registered here, so those two fall through the trie to `OpOther` (`RouteName = "other"`) and get no `clusterMgrRead` routing. Suggest adding both with `r.clusterMgrRead` / `OpSnapshotStatus`.

osprom/registry.go
```diff
@@ -209,58 +265,73 @@ func (r *Registry) drain() {
...
+	if n, dropped := r.enqueue(s); dropped && handler != nil {
+		handler(n, e)
 	}
 }
 
 // OnRequestResponse records a buffered (full-read) request. Safe for concurrent
 // use; returns without blocking.
 func (r *Registry) OnRequestResponse(e opensearchtransport.RequestResponseEvent) {
-	r.enqueue(sample{
+	enqueueOrOverflow(r, envelope{
```
Jakob3xD: Question / doc-clarity: `RouteName`, `Index`, and `PoolName` are not copied into the `envelope` here (nor exposed on `RequestSample`), so the bundled `osprom`/`osotel` registries cannot attach `RouteName` as a metric label — even though it is advertised (RequestEvent doc + CHANGELOG) as "suitable as a low-cardinality metric label." This is plausibly an intentional cardinality choice (keep `RequestSample` coarse; ~110 routes is a new label dimension). If so, a short note pointing users who want per-route labels to implement `ConnectionObserver` directly would close the gap between the advertised property and what the shipped observers can consume. Same for `osotel`.

Jakob3xD (commented):
### Code review

Solid PR — correctness and concurrency look clean (the identity capture is properly guarded for the no-observer path, and the new `sr.*` fields are stack-local, so no #901/#981-style lock regressions). All findings below are **non-blocking**: five inline, plus one here.

6. `osotel/registry_internal_test.go` drops `TestRegistryCloseIdempotent` (replaced wholesale by `TestWithStatusClassifier`), while `osprom/registry_internal_test.go` keeps both. `Close` idempotency is a real invariant (`stop = sync.OnceFunc(...)`), so osotel loses that regression coverage asymmetrically. Consider keeping the four-line idempotency test alongside the new one.

https://github.com/opensearch-project/opensearch-go/blob/42b427a6f2c1a86e76c7a60d4583560caff73cdb/osotel/registry_internal_test.go#L139-L140

opensearchtransport/opensearchtransport.go
sean-: Good catch, thank you. Fixed: `opensearch.Config` now has an `OperationClassifier` field and `opensearch.NewClient` passes it into `opensearchtransport.New(...)` alongside `Router`/`Observer`/`Signer`, so the documented `NewClient` path can set it.

osprom/registry.go
sean-: You're right, the doc was wrong - the default also emits `3xx`/`5xx` for real traffic plus `unknown` for out-of-range. Reworded to:
> the per-hundred bucket (`2xx`/`3xx`/`4xx`/`5xx`), `error` (no response), or `unknown`

in both modules and the CHANGELOG. Note the classifier also moved: with the sink rework it's now a `RequestObserver` option (`WithStatusClassifier`) rather than a registry option, since labeling is the observer's concern.

osprom/registry.go
sean-: Correct - a dispatch worker can drain an entry between the full-buffer `select` default and the `len(r.ch)` read, so it's a ceiling, not an exact value. Reworded the `RequestOverflowHandler` doc to "at most the configured buffer size" in both `osprom` and `osotel`.

opensearchtransport/router.go
sean-: Thanks - added both. `GET /_snapshot/_status` and `GET /_snapshot/{repository}/_status` now route via `r.clusterMgrRead` with `OpSnapshotStatus`, matching the per-snapshot form. Confirmed the trie disambiguates `_status` from the `{snapshot}` wildcard (a per-snapshot status still classifies as `snapshot_status`, and `/_snapshot/{repo}/{snap}` still resolves to `snapshot_get`).

osprom/registry.go
sean-: This one prompted a broader rework. Rather than widen `RequestSample`, the registries now use an event-sink `Observer`: `OnRequestResponse`/`OnStreamResponse` receive the full `*opensearchtransport.RequestResponseEvent`/`*StreamResponseEvent`, so `RouteName`/`Index`/`PoolName` are available to any sink. `RequestSample` is removed.

The shipped `RequestObserver` deliberately stays coarse (method/status/mode) to avoid a per-route cardinality explosion. To label by route/index/pool, implement `Observer` over your own metrics; the README's "Custom observers" section now shows exactly that. A `PoolObserver` (USE) has also been added to get connection-pool utilization/saturation/errors.

ryanyuan (approved):

PR #1063, opened 2026-08-15
sign background health-check and discovery requests

### LINKED ISSUE #1062: [BUG] SigV4 signer not applied to background health-check/discovery/node-stats requests (continuous 403s on Amazon OpenSearch Service)
### What is the bug?

`opensearchtransport` applies the configured `Signer` (e.g. `signer/awsv2` SigV4) only on the `Perform` path. The background health-check, discovery, and node-stats pollers build their own request and dispatch it with a raw `c.transport.RoundTrip(req)`, calling `setReqAuth` but never `signRequest`.

`setReqAuth` only ever sets HTTP basic auth:

```go
func (c *Client) setReqAuth(u *url.URL, req *http.Request) {
	if _, ok := req.Header["Authorization"]; !ok {
		if u.User != nil { ... req.SetBasicAuth(...); return }
		if c.username != "" && c.password != "" { req.SetBasicAuth(c.username, c.password); return }
	}
}
```

When the client is configured for SigV4, there is no username/password, so these background requests leave the process with no credentials at all. Against Amazon OpenSearch Service (IAM-principal-based access policy) every one is rejected with HTTP 403.

The failures are swallowed by the pollers, so there is no error surfaced to the application.

### Affected call sites

`signRequest` is called from only two places, both on the `Perform` path (`opensearchtransport.go`, currently ~L1542 and ~L1878). Seven other sites dispatch via raw `RoundTrip` after `setReqAuth` only:

| File | Approx. lines (`main`) | Request |
|---|---|---|
| `opensearchtransport/cluster_health.go` | 202 / 209 | `GET /_nodes/_local/stats/jvm,breaker,thread_pool` (`fetchAndEvaluateNodeStats`) |
| `opensearchtransport/discovery.go` | 1125 / 1128 | discovery probe |
| `opensearchtransport/discovery.go` | 1914 / 1917 | discovery probe |
| `opensearchtransport/discovery.go` | 2167 / 2170 | discovery probe |
| `opensearchtransport/opensearchtransport.go` | 2214 / 2221 | health probe |
| `opensearchtransport/opensearchtransport.go` | 2293 / 2300 | health probe |
| `opensearchtransport/opensearchtransport.go` | 2425 / 2432 | health probe |

Shape of the defect, from `cluster_health.go`:

```go
req, err := http.NewRequestWithContext(ctx, http.MethodGet, "/_nodes/_local/stats/jvm,breaker,thread_pool", nil)
...
c.setReqURL(conn.URL, req)
c.setReqAuth(conn.URL, req)      // basic auth only -- signer never invoked
c.setReqUserAgent(req)
...
res, err := c.transport.RoundTrip(req)   // bypasses Perform, so bypasses signRequest
```

1. Point a client at an Amazon OpenSearch Service domain (managed, VPC, fine-grained access control disabled, IAM-based access policy).
2. Configure SigV4 via `signer/awsv2`, with no `Username`/`Password` set.
3. Issue normal indexing/search traffic so the transport establishes ready connections and starts its background pollers.
4. Observe requests to `GET /_nodes/_local/stats/jvm,breaker,thread_pool` returning 403, at the `NodeStatsInterval` cadence (default auto-derived, 5–30s), per ready connection.

Background health-check, discovery, and node-stats requests should be signed with the same `Signer` as application requests, so they authenticate identically. Concretely, the raw-dispatch paths should call `signRequest(req)` before `c.transport.RoundTrip(req)`.

### Impact

- Per-pool AIMD congestion control and overload detection (`NodeStatsInterval`, `OverloadedHeapThreshold`, `OverloadedBreakerRatio`) never receive a single sample on Amazon OpenSearch Service with SigV4, so no congestion window is ever adjusted and no overloaded node is ever demoted. The feature is inert while appearing configured.
- Continuous authorization failures against the domain for the lifetime of every client, invisible to the application because the poller discards the error.
- No configuration or IAM change can work around it: the identity and domain policies can already permit the path, but an unsigned request presents no principal to authorize.

The only available mitigation is to disable the feature entirely, e.g. `OPENSEARCH_GO_DISCOVERY_CONFIG=-node_stats`.

- `github.com/opensearch-project/opensearch-go/v4 v4.7.1`
- Go client against Amazon OpenSearch Service `OpenSearch_3.7`, VPC domain, FGAC disabled, IAM/SigV4 data-plane access.

iprithv:
background health-check, discovery, and node-stats pollers were constructing their own `http.Request` and dispatching it with raw `RoundTrip` after calling `setReqAuth`. Since `setReqAuth` does not invoke the configured `Signer`, SigV4 signing was only applied on the `stream()` and `performSeedFallback()` paths.

when connecting to Amazon OpenSearch Service with IAM-only access, these background requests returned `403`. failures were swallowed, which caused several features to silently stop working:

- resurrection health checks never succeeded, so dead nodes could not come back into the pool.
- hardware discovery never ran, leaving the capacity model at the 8-core default.
- node-stats AIMD never received samples, so congestion windows never adjusted.
- `/_cat/shards` and routing metadata were never loaded, so murmur3 shard-exact routing was not enabled.

user search/index traffic continued to work because it goes through `stream()`.

fixes #1062


opensearchtransport/cluster_health.go
```diff
@@ -198,12 +198,8 @@ func (c *Transport) fetchAndEvaluateNodeStats(conn *Connection, pool *multiServe
 		return sample, false
 	}
 
-	c.setReqURL(conn.URL, req)
-	c.setReqAuth(conn.URL, req)
-	c.setReqUserAgent(req)
-
-	if c.healthCheckRequestModifier != nil {
-		c.healthCheckRequestModifier(req)
+	if err = c.prepareInternalRequest(conn.URL, req, c.healthCheckRequestModifier); err != nil {
```
ryanyuan, line 201: The `err` gets dropped with no trace. Could you log it through loadDebugLogger()?

opensearchtransport/opensearchtransport.go
```diff
@@ -2050,6 +2050,29 @@ func (c *Transport) setReqAuth(u *url.URL, req *http.Request) {
...
+// uses -- URL rewrite, basic auth, User-Agent, Config.Header, and the
+// configured Signer -- to a request that will be dispatched via raw
+// RoundTrip rather than stream(). Background health-check, discovery,
+// and node-stats pollers must go through this so SigV4 (and Config.Header)
+// apply to them the same way they apply to user traffic.
+//
+// applyModifier, if non-nil, runs after header injection and before
+// signing so any headers it adds are included in the signature.
+func (c *Transport) prepareInternalRequest(u *url.URL, req *http.Request, applyModifier func(*http.Request)) error {
+	c.setReqURL(u, req)
```
ryanyuan, line 2070: Shall we call `c.setReqAuth()` after `c.setReqGlobalHeader()` to match what `stream()` (`c.setReqGlobalHeader()` at :1420, `c.setReqAuth()` at :1531) does?

Both are add-if-absent, so the order decides which credential wins for a client configured with both a Config.Header Authorisation and basic auth.

Worth adding some tests for this as well. Nothing in the package pins `stream()`'s order today, so the two can drift apart again without anything failing.

opensearchtransport/prepare_internal_request_internal_test.go
```diff
@@ -0,0 +1,401 @@
...
+	"github.com/stretchr/testify/require"
+
+	"github.com/opensearch-project/opensearch-go/v5/signer"
+)
+
+// recordingSigner records every request it is asked to sign and stamps
+// Authorization so the RoundTripper can assert the stamp made it onto
+// the wire. err, if set, is returned after the request is recorded.
+type recordingSigner struct {
+	mu   sync.Mutex
```
ryanyuan: Could these move the guarded fields inside mu?

i.e.
```
type recordingSigner struct {
	mu struct {
		sync.Mutex
		reqs []recordedSign
		err  error
	}
}
```

The accesses become `s.mu.Lock()`, `s.mu.reqs`, etc. so what the lock covers is visible at the declaration. It also makes anything outside `mu` is by definition not protected.

opensearchtransport/prepare_internal_request_internal_test.go
```diff
@@ -0,0 +1,401 @@
...
+		out[i] = r.path
+	}
+	return out
+}
+
+// captureTripper records the request that actually left the process and
+// returns a canned response. Header values are snapshotted at RoundTrip
+// time so a later mutation of the original request cannot hide a miss.
+type captureTripper struct {
+	mu     sync.Mutex
```
ryanyuan: same as https://github.com/opensearch-project/opensearch-go/pull/1063/changes#r3793721089

opensearchtransport/prepare_internal_request_internal_test.go
```diff
@@ -0,0 +1,401 @@
...
+		header:             header,
+		userAgent:          "opensearch-go-test",
+		healthCheckTimeout: time.Second,
+		ctx:                t.Context(),
+	}
+	tp.mu.connectionPool = newSingleServerPool(conn, nil)
+	return tp, u, conn
+}
+
+func TestPrepareInternalRequest(t *testing.T) {
```
ryanyuan, line 191: As these tests have the same shape, could they become table-driven?

opensearchtransport/prepare_internal_request_internal_test.go
```diff
@@ -0,0 +1,401 @@
...
+				"http": {"publish_address": "127.0.0.1:9200"}
+			}
+		}
+	}`)
+	catShardsBody := []byte(`[{"index":"idx","shard":"0","prirep":"p","state":"STARTED","node":"n1"}]`)
+	routingMetaBody := []byte(`{"metadata":{"indices":{"idx":{"routing_num_shards":640,"settings":{"index":{"number_of_shards":"5"}}}}}}`)
+	hardwareBody := []byte(`{"nodes":{"n1":{"os":{"allocated_processors":8},"thread_pool":{"search":{"max":13}}}}}`)
+	nodeStatsBody := []byte(`{"nodes":{"n1":{"jvm":{"mem":{"heap_used_percent":10}},"breakers":{},"thread_pool":{}}}}`)
+
+	t.Run("each background path is signed before RoundTrip", func(t *testing.T) {
```
ryanyuan: As these tests have the same shape, could they become table-driven?

signer/awsv2/sdkv2signer_test.go
```diff
@@ -249,6 +249,37 @@ func TestV4SignerAwsSdkV2(t *testing.T) {
 		require.Error(t, err)
 		require.True(t, body.closed, "request body must be closed even when the read fails")
 	})
+
+	t.Run("honors a cancelled request context for credential retrieval", func(t *testing.T) {
```
ryanyuan: As these tests have the same shape, could they become table-driven?

opensearchtransport/prepare_internal_request_internal_test.go
```diff
@@ -0,0 +1,401 @@
...
+			"/_nodes/_local/http,os,thread_pool",
+			"/_cluster/health",
+			"/_nodes/_local/stats/jvm,breaker,thread_pool",
+			"/_nodes/http",
+			"/_cat/shards",
+			"/_cluster/state/metadata/idx",
+		}, sig.paths())
+	})
+
+	t.Run("nil signer still dispatches (basic-auth / no-auth clusters)", func(t *testing.T) {
```
ryanyuan: As these tests have the same shape, could they become table-driven?

ryanyuan (commented):
Would you also have some tests to cover the signing-failure branches? The test coverage has slightly dropped a little bit with this PR.

sean- (commented):
Thank you for tackling this, @iprithv . Please see @ryanyuan 's comments but this looks good!

opensearchtransport/cluster_health.go
iprithv: understood, sure. done!

opensearchtransport/prepare_internal_request_internal_test.go
iprithv: sure, done.

opensearchtransport/prepare_internal_request_internal_test.go
iprithv: yes, done. thanks!

ryanyuan (approved):
LGTM. Thanks for working on this one, @iprithv

ryanyuan:
@iprithv since the the v4 also has the same problem, would you like to raise another PR to apply this fix on the `v4` branch?

Issue #828, opened 2026-04-29
Inconsistent newline handling in debugLogger.Logf


sean-:
`debuggingLogger.Logf` (`opensearchtransport/logger.go:434`) uses raw `fmt.Fprintf` without appending a trailing newline. This pushes newline responsibility onto every caller, and the convention is already inconsistent:
74 of 76 call sites include `\n`, while two in `policy_chain.go` (lines 176 and 203) do not.

The standard library's `log.Printf` appends a newline when the format string does not end with one. Our `Logf` should do the same.

### Proposed fix

1. Append `\n` in `Logf` when `format` does not already end with one (matching `log.Printf` semantics).
2. Strip the trailing `\n` from all 74 call sites that currently include it.
3. No change needed for the two `policy_chain.go` callers that already omit it.

### Risk

`Logf` is part of the exported `Logger` interface. Callers with custom implementations that already append their own newline would see double newlines until they update. This is a minor cosmetic issue and the fix aligns the interface with Go conventions (`log.Printf`, `testing.T.Logf`).

### Scope

Purely internal cleanup. No behavior change beyond log formatting.

krisfreedain:
@Jakob3xD @VachaShah @VijayanB - please take a look 

[Catch All Triage](https://github.com/opensearch-project/.github/blob/main/TRIAGING.md) - [1](https://github.com/krisfreedain) [2](https://github.com/andrross)

ssunno:
It seems like this problem has quietly remained for a long time.

I would like to discuss two things regarding logging:

1. Why are we use `fmt.Fprintf` instead of standard logging? I think using a logger will be more stable and user-friendly than `fmt.Fprintf`. (also it can solve `\n` issue)
2. One step more, I think now is the time to modernize the logger. current logger was designed quite a long time ago, and I think it would be good to adopt a structure that is compatible with the log package in modern Go.

sean-:
If I could cash a nickel for every time I though during one of the refactors that I've done recently...  I would love to have a zerolog logger, or something that adapts to zerolog and slog, but before that happens, I want to get the v5 branch cut per #835.  Can we hold on this until then?

ssunno:
Sure, changing the logo will not be a small task. Let's apply this to v5.

sean-:
@ssunno have you had a chance to look at #1087 ?

PR #1087, opened 2026-08-21
feat(logger)!: replace DebuggingLogger with the debuglog builder interface

ryanyuan:
### Description
Replaces `opensearchtransport.DebuggingLogger` with `debuglog`, a new package defining a zerolog-shaped builder interface the client emits through, and ships zerolog and slog adapters as separate modules.

### Why

`DebuggingLogger` was two `...any` methods:

```go
type DebuggingLogger interface {
	Log(a ...any) error
	Logf(format string, a ...any) error
}
```

`Logf` appended no newline, so every call site carried its own. Of v4's 79 `dl.Logf` sites, 77 ended their format string with `\n` and the two in `policy_chain.go` did not, so those records ran into the next one. #828 reported the inconsistency and proposed appending `\n` inside `Logf`.

The discussion asked for more than that: a structure compatible with the log package in modern Go, and a logger that adapts to zerolog and slog. This branch takes that route, and the newline bug goes away with the printf formatting, because a record now terminates where it is emitted.

There was a second limitation. Debug records were formatted with `fmt` and written only to the client's own stream, so an application had no way to route them into the logger it already runs. An interface the client owns removes that without putting a logging library in the core dependency graph.

The interface is shaped after zerolog's builder rather than after a flat `Debug(msg string, kv ...any)`. A flat variadic boxes every value into an `any`, which would hand zerolog the shape of a structured logger and none of its encoding speed, making the faster of the two libraries pay for the slower one's API.

#### The interface

New package `github.com/opensearch-project/opensearch-go/v5/debuglog`. It belongs to the root module, not a separate one, and imports only the standard library.

```go
type Logger interface {
	Debug() Event
}

type Event interface {
	Str(key, val string) Event
	Strs(key string, val []string) Event
	Int(key string, val int) Event
	Int32(key string, val int32) Event
	Int64(key string, val int64) Event
	Uint32(key string, val uint32) Event
	Float64(key string, val float64) Event
	Dur(key string, val time.Duration) Event
	Time(key string, val time.Time) Event
	Stringer(key string, val fmt.Stringer) Event
	Err(err error) Event
	Msg(msg string)
}
```

Two properties drove the rest of the design. Each key's type is fixed by the method carrying it, so a dangling key or a non-string key cannot be written and there is no `!BADKEY` case anywhere. And `Stringer` defers `String()` to emit time, which is what makes an expensive field free when logging is off.

Eleven field methods is the cost. It is also what a census of the client's own records needs: 85 emitting chains carry 230 field calls, every one of the eleven methods is used at least once, and `Int32` alone accounts for 17, so the set does not trim to zerolog's common four. There is deliberately no `Any(key string, val any)` escape hatch, because it would reintroduce the boxing the typed methods exist to avoid and would become the path of least resistance for every future site. The one value that fit nothing (`warmupManager`, a packed bitfield) got a `String()` instead.

#### Installing one

`Config.DebugLogger` is new on both `opensearch.Config` and `opensearchtransport.Config`; v4 had no equivalent. A supplied logger enables debug output and takes precedence over `EnableDebugLogger`, which still selects the built-in stderr logger.

```go
client, err := opensearch.NewClient(opensearch.Config{
	DebugLogger: logzerolog.Default(),
})
```

Two adapter modules present the same `New`/`Default` pair: [`log-zerolog`](log-zerolog/README.md) and [`log-slog`](log-slog/README.md). Both compile against `debuglog` alone rather than against `opensearchtransport`, so an adapter, including a third-party one, needs no package of the client beyond the interface it implements.

The root module implements none of it. It defines the interface, calls it, and ships a hand-rolled stderr fallback so the environment variable keeps working with no imports. Same reasoning as the AWS signers living in submodules.

#### The guard is gone

`LoadDebugLogger()` is replaced by package-level `opensearchtransport.Debug()`, which never returns nil, returning `debuglog.Nop()` when no logger is installed. All 85 internal emitting sites drop the `if dl := ...; dl != nil` guard they used to need:

```go
// Before
if dl := loadDebugLogger(); dl != nil {
	_ = dl.Logf("Resurrecting connection %s\n", conn.URL)
}

// After
Debug().Stringer("conn", conn.URL).Msg("Resurrecting connection")
```

`Msg` returns no error, so the `errcheck` exclusion for `dl.Logf` in `.golangci.yml` is deleted rather than replaced.

#### `OPENSEARCH_GO_LOG`

New variable taking a level name. `debug`, matched case-insensitively, installs the built-in logger. The client emits at one level, so this is a switch wearing a level's name rather than a ladder: every other level, and any unrecognized value, leaves the records off, which means raising the level is how you turn them off.

`OPENSEARCH_GO_DEBUG` keeps working and is read whenever `OPENSEARCH_GO_LOG` is unset or empty, so nothing exporting it has to change. When both carry a value `OPENSEARCH_GO_LOG` decides, which is what lets a deployment be quieted without first unsetting the boolean. Both resolve through one function so the transport's `init` and `testutil.IsDebugEnabled` cannot drift on what "enabled" means.

### Performance

Measured on darwin/arm64, Apple M4 Max, go1.26.4, writing to `io.Discard`. `benchstat` medians of 10 runs throughout, including the pre-pooling baselines, which were re-measured at the pre-pooling commit rather than quoted from the shorter run they were first taken from. All three implementations benchmark identical record shapes.

| record                | log-zerolog             | log-slog (JSONHandler)    | log-slog (TextHandler)    | built-in text           |
| --------------------- | ----------------------- | ------------------------- | ------------------------- | ----------------------- |
| 1 field               | 83.4 ns, 32 B, 1 alloc  | 356.5 ns, 32 B, 1 alloc   | 386.7 ns, 32 B, 1 alloc   | 110.7 ns, 32 B, 1 alloc |
| 1 field, no Stringer  | 55.5 ns, 0 B, 0 allocs  | 323.9 ns, 0 B, 0 allocs   | 353.8 ns, 0 B, 0 allocs   | 84.0 ns, 0 B, 0 allocs  |
| 4 fields              | 139.9 ns, 32 B, 1 alloc | 486.4 ns, 32 B, 1 alloc   | 624.0 ns, 56 B, 2 allocs  | 144.0 ns, 32 B, 1 alloc |
| 4 fields, no Stringer | 111.5 ns, 0 B, 0 allocs | 450.7 ns, 0 B, 0 allocs   | 588.9 ns, 24 B, 1 alloc   | 117.6 ns, 0 B, 0 allocs |
| 8 fields              | 195.6 ns, 32 B, 1 alloc | 715.4 ns, 168 B, 3 allocs | 805.0 ns, 184 B, 3 allocs | 199.6 ns, 32 B, 1 alloc |
| level rejects it      | 10.9 ns, 0 B, 0 allocs  | 6.3 ns, 0 B, 0 allocs     | 6.3 ns, 0 B, 0 allocs     | n/a                     |
| no logger installed   | n/a                     | n/a                       | n/a                       | 3.7 ns, 0 B, 0 allocs   |

Each "no Stringer" row is the record above it with the deferred `Stringer` swapped for an already-resolved string. The paired rows isolate what a logger itself allocates from what resolving the connection address costs, and the one-field pair does it best: a record of one `Stringer` field is otherwise the cheapest thing any of these can emit.

Nothing a logger itself does allocates. Both no-Stringer rows are 0 B and 0 allocations for `log-zerolog`, for `log-slog` under `JSONHandler`, and for the built-in logger; the one allocation the other rows have in common is `(*url.URL).String` building the connection address, paid by the caller's value.

The client's own records no longer pay even that. `Connection` already caches its address as `URLString`, set once at construction, so all 50 emitting sites that name a connection pass it through `Str` and land on the no-Stringer row. `Stringer` stays in the interface for values that are expensive to render and may never be rendered, which is what the connection-state and lifecycle fields are; the three sites still using it take a bare `*url.URL` in discovery's address-rewrite path, with no connection behind it to have cached anything.

#### Per field

Whole-record numbers cannot tell you which method is dear, so each row below is a one-field record: subtracting the "no fields" row gives what that method adds on top of `Debug` and `Msg`. Allocation counts are exact counters rather than samples, and were identical across all 10 runs.

| field      | log-zerolog       | log-slog (JSONHandler) | log-slog (TextHandler) | built-in text      |
| ---------- | ----------------- | ---------------------- | ---------------------- | ------------------ |
| no fields  | 41.1 ns, 0 allocs | 271.4 ns, 0 allocs     | 303.4 ns, 0 allocs     | 77.0 ns, 0 allocs  |
| `Str`      | 49.9 ns, 0 allocs | 320.4 ns, 0 allocs     | 347.6 ns, 0 allocs     | 84.4 ns, 0 allocs  |
| `Strs`     | 60.9 ns, 0 allocs | 415.7 ns, 1 alloc      | 572.2 ns, 5 allocs     | 100.4 ns, 0 allocs |
| `Int`      | 50.5 ns, 0 allocs | 316.6 ns, 0 allocs     | 347.8 ns, 0 allocs     | 85.4 ns, 0 allocs  |
| `Int32`    | 49.8 ns, 0 allocs | 317.3 ns, 0 allocs     | 346.4 ns, 0 allocs     | 85.4 ns, 0 allocs  |
| `Int64`    | 49.4 ns, 0 allocs | 316.5 ns, 0 allocs     | 348.0 ns, 0 allocs     | 84.8 ns, 0 allocs  |
| `Uint32`   | 48.5 ns, 0 allocs | 316.4 ns, 0 allocs     | 346.9 ns, 0 allocs     | 84.5 ns, 0 allocs  |
| `Float64`  | 65.3 ns, 0 allocs | 383.4 ns, 1 alloc      | 367.6 ns, 0 allocs     | 103.8 ns, 0 allocs |
| `Dur`      | 71.7 ns, 0 allocs | 321.4 ns, 0 allocs     | 355.0 ns, 0 allocs     | 91.3 ns, 0 allocs  |
| `Time`     | 73.3 ns, 0 allocs | 365.6 ns, 0 allocs     | 396.2 ns, 0 allocs     | 183.6 ns, 0 allocs |
| `Stringer` | 82.5 ns, 1 alloc  | 363.0 ns, 1 alloc      | 390.8 ns, 1 alloc      | 111.0 ns, 1 alloc  |
| `Err`      | 59.9 ns, 0 allocs | 334.1 ns, 0 allocs     | 464.2 ns, 1 alloc      | 82.7 ns, 0 allocs  |

On the built-in logger and on `log-zerolog`, every method except `Stringer` is allocation-free, `Float64` included: both append through the `strconv.Append` functions. `Stringer` is the only method that allocates on all four, and that allocation belongs to the caller's `String`.

The rest is `log-slog`'s handlers rather than the adapter. `JSONHandler` allocates on `Strs` and on `Float64`, which it routes through `encoding/json` rather than `strconv` because `json.Marshal` does not always agree with `strconv.AppendFloat`. `TextHandler` renders floats itself but spends five allocations on `Strs` and one on `Err`, both through `fmt`: it has no `error` case, where `JSONHandler` writes `Error()` directly. That asymmetry is why `TextHandler` costs one allocation more than `JSONHandler` on the four-field record, which carries an error and no float, and why the two come out even at eight fields, which carry both.

Widening `Int32` and `Uint32` to 64 bits costs nothing measurable. `Time` is the dearest field on the built-in logger, roughly doubling a bare record to format the timestamp layout, and it allocates nothing.

Both the built-in logger and `log-slog` originally allocated an event and a growing buffer per record. Review caught the first; sweeping for the same shape caught the second. Both now take the event from a `sync.Pool`:

| implementation          | 1 field                   | 4 fields                  | 8 fields                  |
| ----------------------- | ------------------------- | ------------------------- | ------------------------- |
| built-in text, before   | 211.7 ns, 384 B, 10 allocs  | 278.6 ns, 576 B, 12 allocs  | 334.8 ns, 608 B, 12 allocs  |
| built-in text, after    | 104.8 ns, 32 B, 1 alloc   | 136.8 ns, 32 B, 1 alloc   | 190.3 ns, 32 B, 1 alloc   |
| log-slog (JSON), before | 367.4 ns, 112 B, 3 allocs | 540.4 ns, 352 B, 5 allocs | 808.1 ns, 809 B, 8 allocs |
| log-slog (JSON), after  | 354.1 ns, 32 B, 1 alloc   | 483.8 ns, 32 B, 1 alloc   | 712.4 ns, 168 B, 3 allocs |

`Event.Time` was the one field type left allocating, because it rendered through `time.Time.String`. It appends through `AppendFormat` now, which also drops the monotonic-clock suffix `String` adds; the client's only `Time` call site reads a timestamp that round-trips through Unix nanoseconds, so no record renders differently. No benchmark shape had covered `Time`, which is why the table reported one allocation per record without ever exercising the field that broke that.

Pooling ruled out `strings.Builder` in the built-in logger, because its `Reset` drops the buffer it just grew. Fields accumulate in a `[]byte` written through the `strconv.Append` functions instead, so a number reaches the buffer without a string in between. An event grown past a cap is dropped rather than pooled, so one long record cannot park its buffer for the life of the process.

What remains in `log-slog` past five fields is not the adapter's to fix: `slog.Record` stores five attributes inline and moves the overflow to the heap, and `JSONHandler` encodes each `float64` through `json.Marshal`. The record is handed to the handler by value to keep `AddSource` pointing at the client code that emitted it rather than at the adapter.

Per record `log-slog` costs three to six times what the other two do, the multiple widening as the record gets cheaper. That, rather than allocation count, is the reason to reach for `log-zerolog` when debug logging is left on somewhere that matters; `log-slog` adds no dependency in exchange. `log-zerolog` leads the built-in logger at every size, by 2 to 3% at four and eight fields and by a third on the shortest record, so what an adapter mostly buys over the built-in logger is where the records go rather than what they cost.

Nothing is emitted and nothing allocated when no logger is installed, which is what makes the 85 unguarded sites affordable.

### Breaking changes

All are relative to v4. See [`UPGRADING_V5.md`](UPGRADING_V5.md) for the migration.

- **`opensearchtransport.DebuggingLogger` is removed.** Implement `debuglog.Logger` instead. A custom implementation trades the two `...any` methods for one method per field type and drops the error return, since a debug logger that cannot write has nowhere left to report it.
- **`opensearchtransport.LoadDebugLogger()` is removed.** Call `opensearchtransport.Debug()`, which never returns nil, so a caller that guarded on nil can drop the guard.
- **`EnableDebugLogger: true` writes to stderr.** It previously wrote to stdout while `OPENSEARCH_GO_DEBUG=true` wrote to stderr. Both now agree, and stderr is what the docs describe. Anything parsing the client's debug output from stdout needs to read stderr.
- **The built-in record format changed** from a formatted line to a message plus key/value pairs:
  ```
  [15:04:05.000] DEBUG    Node overloaded: heap over threshold conn=https://localhost:9200 heap_used_percent=93 threshold=85
  ```
- **`OPENSEARCH_GO_POLICY_DUMP` writes its tree straight to stderr** rather than through the debug logger, so it stays one contiguous block a reader can copy a path out of. Through a structured logger it would collapse into a single record with every newline escaped. A `Config.DebugLogger` therefore switches the dump on without receiving it.

`opensearchutil.BulkIndexerConfig.DebugLogger` is a different field and is unchanged. It still takes a `BulkIndexerDebugLogger` and logs the indexer's own worker activity, not the client's internal records.

### Other notes

**A chain that never reaches `Msg` emits nothing, and the compiler cannot say so**, because a chain is a valid expression statement whether or not it terminates. A test in the `debuglog` package parses every module in the repository looking for that shape, so a missing terminator fails the test suite rather than shipping. The guard matches a chain written as a statement, which is how every emitting site is written; a chain assigned to a variable first is outside what it can see.

**Embedding an `Event` to inherit most of the methods fails quietly.** A promoted field method returns the embedded value rather than the embedding type, so the chain leaves the implementation at the first field and its `Msg` never runs. Documented on the interface, because nothing in the compiler catches it.

**`debuglog.StringerText` renders `<nil>` rather than panicking.** A nil `*url.URL` satisfies `fmt.Stringer` while `(*url.URL).String` dereferences its receiver, and a `*url.URL` is the client's most common debug value. Debug logging must not be able to take down the program it is describing.

**Pooling changes the blast radius of calling a method after `Msg`** from harmless to a cross-goroutine race, since the event may already have been handed to another record. This is inherent to the builder-plus-pool pattern and zerolog's own `*Event` has the same exposure. It is documented, not enforced.

**The error field's key belongs to the implementation.** The built-in logger and `log-slog` use `"err"`; `log-zerolog` uses zerolog's configured `ErrorFieldName`, `"error"` by default, so `ErrorMarshalFunc` and `ErrorStackMarshaler` keep working.

**`Config.DebugLogger` is not hashable**, so a config carrying one is never cached by the default-client cache.

**`osotel.WithLogger` and `osprom.WithLogger` now take a `debuglog.Logger`** and default to `opensearchtransport.Debug` resolved per message, so a registry's lifecycle messages follow the same switch as the client's own records. This is not a v4 break: neither module exists on v4, both are new in the same unreleased cycle, so no released caller can hit the old `*slog.Logger` signature.

### Issues Resolved
Closes [#828]


debuglog/chain_internal_test.go
```diff
@@ -0,0 +1,241 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+// fieldMethods are the [Event] methods that return an Event. A chain ending in
+// one of them, used as a statement, has discarded the Event without emitting:
+// the record is silently dropped and, for the zerolog adapter, its pooled event
+// is never returned to the pool.
+//
+// Msg is deliberately absent -- it is the terminator this guard looks for.
+var fieldMethods = map[string]bool{
```
sean-: Use `map[string]struct{}` instead

debuglog/chain_internal_test.go
```diff
@@ -0,0 +1,241 @@
...
+}
+
+// chainRoots are the names of the calls that begin a debug chain: the
+// opensearchtransport accessor, and the per-message resolver the metrics
+// registries hold.
+//
+// Matching the root is what keeps this off unrelated builders. A config hash
+// builder chains Int the same way a debug record does, so the method names alone
+// do not identify a debug chain.
+var chainRoots = map[string]bool{
```
sean-: Ditto: `map[string]struct{}`

debuglog/README.md
```diff
@@ -0,0 +1,109 @@
...
+
+This is the one place these numbers live. Both adapter READMEs, the root README, and the user guide point here rather than repeating them, so a re-run updates one table.
+
+Measured on darwin/arm64, Apple M4 Max, go1.26.4, writing to `io.Discard`, medians of 10 runs through `benchstat`. All three implementations benchmark identical record shapes; reproduce with `go test -run=none -bench=. -benchmem -count=10 ./...` in `log-zerolog`, `log-slog`, and the root module.
+
+The "no Stringer" row is the same four-field record with the deferred `Stringer` swapped for an already-resolved string. Every other row pays one allocation inside `(*url.URL).String`, so that row is the one that shows what a logger itself allocates rather than what resolving the connection address costs.
+
+| record                | log-zerolog             | log-slog (JSONHandler)    | log-slog (TextHandler)    | built-in text           |
+| --------------------- | ----------------------- | ------------------------- | ------------------------- | ----------------------- |
+| 1 field               | 82.3 ns, 32 B, 1 alloc  | 354.1 ns, 32 B, 1 alloc   | 387.9 ns, 32 B, 1 alloc   | 104.8 ns, 32 B, 1 alloc |
```
sean-: Are these alloc numbers still accurate after the `sync.Pool` work?

sean- (dismissed):

iprithv (approved):

debuglog/README.md
ryanyuan: Yes. The numbers were generated in [c76e9cb](https://github.com/opensearch-project/opensearch-go/pull/1087/commits/c76e9cb763ecfd2afe4aca1aede64dda6381b547), which is the last commit that touched allocation. Just re-ran to confirm.

debuglog/chain_internal_test.go
ryanyuan: fixed in [9265712](https://github.com/opensearch-project/opensearch-go/pull/1087/commits/92657124cbf26146130cbd59f22b57e7c094f285)

debuglog/chain_internal_test.go
ryanyuan: fixed in [9265712](https://github.com/opensearch-project/opensearch-go/pull/1087/commits/92657124cbf26146130cbd59f22b57e7c094f285)

debuglog/README.md
sean-: Where is the allocation coming from?

ryanyuan: It was `url.URL.String()`, which was used by the benchmarking, resolving `conn.URL` through `Stringer`, so the allocation belonged to the caller's value, not the logger.

I've added a `1 field, no Stringer` row to check the cached address and it had 0 allocation as `Connection` cached it as `URLString` at construction, so they use `Str`.

sean- (approved):

PR #1089, opened 2026-08-21
stop stacking connection path prefixes on retry

iprithv:
`stream()` reuses one `*http.Request` and `setReqURL` prepends the selected connection's base path in place. For a reverse-proxy address like `https://host/prefix`, attempt 0 correctly sent `/prefix/_search` and attempt 1 sent `/prefix/prefix/_search` (default `MaxRetries` is 6, 502/503/504 retry).

After the first rewrite, `Route()` also matched on the leftover path. That is not a miss: the mux treats `/prefix/_search` as `/{index}/_search` with index `prefix`, so retry classified the wrong index before stacking another copy of the prefix. A prefix-less discovered node after a prefixed seed kept the leftover prefix because `baseLen == 0` returns without assigning `Path`.

The caller-supplied `Path`/`RawPath` are captured once and restored at the start of every attempt (before `Route()` / `setReqURL`). Bare `https://host:9200` is unchanged.


opensearchtransport/opensearchtransport.go
```diff
@@ -2011,6 +2024,21 @@ func (c *Transport) URLs() []*url.URL {
...
 
+// restoreReqPath writes the caller-supplied path back onto req. stream()
+// reuses one *http.Request across retries and setReqURL prepends the
+// connection base path in place, so without a restore a prefixed address
+// (https://host/prefix) becomes /prefix/prefix/_search on the next attempt.
+// Route() matches on Path: leftover /prefix/_search is not a miss, the mux
+// treats it as /{index}/_search with index "prefix".
+func restoreReqPath(req *http.Request, path, rawPath string) {
+	req.URL.Path = path
+	req.URL.RawPath = rawPath
```
sean-, line 2035: It seems like `stream()` and should be decomposed to `streamOnce()` and then in the child we can use something like:

```
var retryRequired bool
defer func() {
  if retryRequired {
	req.URL.Path = path
	req.URL.RawPath = rawPath
  }
}()
```

that way it's consistently applied on all error paths.

opensearchtransport/transport_coverage_internal_test.go
```diff
@@ -598,3 +824,65 @@ func TestCalculateNodeStatsInterval(t *testing.T) {
...
+func (r *pathRecordingRouter) RotateStandby(_ context.Context, _ int) (int, error)  { return 0, nil }
+
+var _ Router = (*pathRecordingRouter)(nil)
+
+// onceThenEmptyRouter returns conn on the first Route, then ErrNoConnections,
+// so stream() takes the retryable-status path once and seed fallback next.
+type onceThenEmptyRouter struct {
+	conn *Connection
+	n    atomic.Int32
+	mu   sync.Mutex
```
sean-: Even in tests, can you use the idiom:

```
struct mu {
  sync.Mutex
  // things protected by mu
  saw []string
}
```

opensearchtransport/transport_coverage_internal_test.go
```diff
@@ -89,6 +91,230 @@ func TestSetReqURL(t *testing.T) {
...
+	}
+
+	require.Equal(t, []string{"/_search", "/_search", "/_search"}, router.paths())
+}
+
+// TestStreamRetryThenSeedFallbackDoesNotStackPrefix covers 502 then
+// ErrNoConnections then seed fallback. The loop restores before Route on
+// the ErrNoConnections iteration, so performSeedFallback must still see
+// /_search and send /prefix/_search, not /prefix/prefix/_search.
+func TestStreamRetryThenSeedFallbackDoesNotStackPrefix(t *testing.T) {
```
sean-: How many of these tests should be turned into table-driven tests?

iprithv: All four Stream cases. They share one fixture (prefixed URL + Stream + inspect paths) and only differ in which path they pin (wire, Route(), seed fallback, percent-encoded RawPath), so they are now one table-driven TestStreamRetryPathPrefix.

Left TestSetReqURL as t.Run subtests: those are unit tests of setReqURL+restoreReqPath in isolation (stable rewrite loop, prefix-less switch, RawPath) and are not Stream hops.

opensearchtransport/transport_coverage_internal_test.go
iprithv: sure, done. thanks!

opensearchtransport/opensearchtransport.go
iprithv: agreed, changed. thanks!

sean- (dismissed):

sean- (dismissed):

ryanyuan:
I benchmarked a few variants of the fix against main (`c83eb08a`) to see where the cost actually comes from.

Setup: darwin/arm64, Apple M4 Max, mock `RoundTripper` (no real network call, so only the fix's own overhead shows), benchstat medians of 12 runs, `-cpu=1`. Ad hoc harness, not part of the committed test suite; numbers are the sec/op geomean across four cases (bare host, prefixed host, prefixed host with 2 retries, prefixed host with 6 retries), relative to unmodified main.

| variant | change | geomean vs main |
|---|---|---|
| capture-only | holds `origPath`/`origRawPath` live, never restores | +1.47% |
| gated (shipped in `c659de2c`) | restores only when `i > 0` | +2.04% (+2.29% on the no-retry case alone) |
| unconditional | restores on every attempt, including attempt 0 | +2.45% |
| earlier PR head (shipped in `d721340`) | unconditional restore + `streamOnce` split out of `stream()` | +8.77% |

The gate does what it's meant to: attempt 0 never calls `restoreReqPath`, and that shows up as the gap between "unconditional" and "gated." But most of the original 8.8-point cost wasn't the restore call at all. It was splitting the retry body into a separate `streamOnce` function, which this branch has since folded back into `stream()`. The capture-only row shows the floor: just keeping the two path locals live across the loop, with no restore logic at all, already costs 1.5%. That part doesn't go away with any gating scheme, since it comes from remembering the caller's path in the first place.

Allocations are unaffected everywhere: 6/7/11/19 per op at 0/0/2/6 retries, identical across every variant. Bytes per op actually improve on the shipped version: -1.73% at 2 retries and -8.42% at 6 retries (2.133 KiB -> 1.953 KiB), because the bug this fixes was making the path string grow on every retry, so fixing it shrinks the string too.

In absolute terms the no-retry cost is 303.0ns on main vs 309.9ns gated, about 7ns, against a mock transport that does no I/O. A real request will spend orders of magnitude more time on the network round trip.

sean- (approved):

PR #1090, opened 2026-08-21
stop ranking from reordering the live connection list

iprithv:
`rankByHash` sorts in place. Callers pass the live RTT-ordered `activeConns`/`sortedConns` slice, then drop the lock. When `/_cat/shards` has not populated the index (first requests, a new index, `-cat_shards`), `rendezvousTopK` aliased that slice instead of copying into the pooled buffer the shard-names path already used. Concurrent `Route()` raced, and the RTT-bucket order that slot filling requires - rebuilt on health checks, not per request - was destroyed.

Both branches now copy before ranking. `rankByHash` still sorts in place, only on that private buffer.


opensearchtransport/rendezvous_hash_internal_test.go
```diff
@@ -125,6 +126,65 @@ func TestRankByHash(t *testing.T) {
 	})
 }
 
+func TestRendezvousTopKDoesNotReorderInput(t *testing.T) {
```
ryanyuan, line 133: Can these 2 tests be table-driven?

opensearchtransport/rendezvous_hash.go
```diff
@@ -76,13 +76,13 @@ func rendezvousTopK(
 		k = len(conns)
 	}
 
-	// Split into shard-hosting and non-shard partitions, preserving
-	// the caller's RTT sort order within each partition.
+	// Copy before ranking. rankByHash sorts in place, and callers pass
```
ryanyuan: Shall we just comment the `end-state behaviour` and `why` here instead of how the code evolved?

CHANGELOG.md
```diff
@@ -258,6 +258,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ### Fixed
 
+- Fix `rendezvousTopK` sorting the live connection list when shard placement is unknown. `rankByHash` sorts in place, and the empty-placement path (`/_cat/shards` not yet populated: first requests, a new index, or `-cat_shards`) aliased the caller's `activeConns`/`sortedConns` slice instead of copying into the pooled buffer the shard-names path already used. Concurrent `Route()` then raced with discovery, and the RTT-bucket order that rendezvous filling "MUST" preserve — rebuilt on health checks, not per request — was destroyed. Both branches now copy before ranking
```
ryanyuan: Would you add the PR url here as reference?

sean- (approved):

PR #1106, opened 2026-08-31
fix(opensearchutil): do not fire OnError for PartialBulkError

iprithv:
When a bulk batch has some items that succeed and some that fail (errors: true), the indexer already tells you about each item through OnSuccess and OnFailure. After that it still returned the partial error, so OnError ran too. That made a mixed batch look like the whole request failed.

OnError is for real request failures (network, HTTP error, bad JSON). Per-item failures should only go to the item callbacks.

This change returns nil after those item callbacks when the error is a partial bulk failure. OnError is no longer called in that case. There is a test for it.


sean- (dismissed):

opensearchutil/bulk_indexer.go
```diff
@@ -815,6 +815,13 @@ func (w *worker) flush(ctx context.Context) error {
 		}
 	}
 
+	// PartialBulkError is success-with-failed-items: the loop above already
+	// called OnSuccess/OnFailure. Returning it would fire indexer OnError at
+	// every flush site (worker.run, auto-flush, Close) as if the request
+	// itself failed.
+	if errors.As(err, &partial) {
```
ryanyuan: `err` here has to be either nil or *PartialBulkError
after the check here: https://github.com/iprithv/opensearch-go/blob/673ac0e126d4112a47aff6cb389cbd41da190ed8/opensearchutil/bulk_indexer.go#L768

opensearchutil/bulk_indexer_internal_test.go
```diff
@@ -679,6 +680,88 @@ func TestBulkIndexerCallbacks(t *testing.T) {
 				require.Equal(t, 1, onErrorCount, "OnError call count")
 			},
 		},
+		{
+			// v5 reports *PartialBulkError when errors:true. flush already
+			// skips handleBulkError and dispatches OnSuccess/OnFailure, but
+			// used to return the partial error, so Close/auto-flush/worker.run
```
ryanyuan: Could we remove this historical narration?

opensearchutil/bulk_indexer_internal_test.go
```diff
@@ -679,6 +680,88 @@ func TestBulkIndexerCallbacks(t *testing.T) {
...
+			// treated a mixed batch as a transport failure.
+			name: "OnError is not called on partial bulk item failures",
```
ryanyuan: This test case is nearly a copy of `per-item OnSuccess and OnFailure`. So maybe we can fold it in instead.

opensearchutil/bulk_indexer.go
```diff
@@ -758,18 +753,30 @@ func (w *worker) flush(ctx context.Context) error {
...
 
-	blk, err = w.bi.config.Client.Doc.Bulk(ctx, req)
-	// Treat opensearchapi.PartialBulkError as success-with-failed-items:
-	// the indexer's whole job is per-item dispatch, so the per-item loop
-	// below already handles `info.Error != nil`. A real flush failure
-	// (transport error, HTTP error, JSON parse error) flows through
-	// handleBulkError as before.
+	blk, err := w.bi.config.Client.Doc.Bulk(ctx, req)
+
+	return w.reportBulkOutcome(ctx, blk, err)
```
sean-: I'm not sure this helped readability.

1. There's a newline above
2. `reportBulkOutcome` is not well named, either.  How about `processBulkResp(...)`?

opensearchutil/bulk_indexer.go
```diff
@@ -758,18 +753,30 @@ func (w *worker) flush(ctx context.Context) error {
...
 		return w.handleBulkError(ctx, fmt.Errorf("flush: %w", err))
 	}
 
 	for i, blkItem := range blk.Items {
+		// The cluster answers one item per action line, so a response carrying
+		// more items than were sent cannot be matched to a caller's item. Stop
+		// rather than index past the buffer.
+		if i >= len(w.items) {
+			break
+		}
```
sean-: This seems like an unnecessary precaution.  Hypothetically this could be a problem, but I'm not sure what scenario we're protecting here.  A rogue server over-emitting responses to a bulk operation?

opensearchutil/bulk_indexer.go
```diff
@@ -758,18 +753,30 @@ func (w *worker) flush(ctx context.Context) error {
...
+// land. A *PartialBulkError is the former, so its rejected documents reach their
+// OnFailure callback and the flush sites (worker.run, auto-flush, Close) see no
+// error for it. It must be called under a lock.
+func (w *worker) reportBulkOutcome(ctx context.Context, blk *opensearchapi.BulkResp, err error) error {
 	var partial *opensearchapi.PartialBulkError
 	if err != nil && !errors.As(err, &partial) {
 		return w.handleBulkError(ctx, fmt.Errorf("flush: %w", err))
 	}
 
 	for i, blkItem := range blk.Items {
```
sean-, line 771: If we're doing a bulk operation on every item, shouldn't we just pre-allocate an error array and then accumulate errors, if any, and then use `slices.Collect` and `errors.Join` to return any errors to the caller?  The individual error handling of a single item in a bulk resp should be a single function, and then the handling of one or more failures can be handled easily.  We could also pool the error array to make this more memory efficient for users of this interface.

opensearchutil/bulk_indexer_internal_test.go
```diff
@@ -623,297 +625,313 @@ func TestBulkIndexerContext(t *testing.T) {
...
-					countFailed++
-					failedIDs = append(failedIDs, item.DocumentID)
-					failedItemBodies = append(failedItemBodies, string(buf))
-					bodyFailureCount[string(buf)]++
-				}
+// itemCallback is what one item's OnSuccess or OnFailure callback reported. A
+// rejection the cluster made carries a status, and an error type when the
+// response item named one; a rejection the indexer made while writing the
+// request carries an error instead.
+type itemCallback struct {
```
sean-: If this is a type for testing, rename this to `testItemCallback` so it's clear it's something test-specific vs something from the `opensearchutil` package.

opensearchutil/bulk_indexer.go
ryanyuan: Happy to drop it. The cases it'd cover are all broken/hostile senders: 

- a server or version bug
- a proxy mangling the response
- a corrupted body that still parses
- a spoofed endpoint. 

The one wrinkle is that the failure is a panic on the worker `goroutine`, so it crashes the whole process rather than just the call, but that's a stretch to protect against.  I'll remove the block, the test, and the changelog line.

opensearchutil/bulk_indexer.go
ryanyuan: Per-item failures already go to each item's `OnFailure`. Flush returns `nil` on partial failure so `OnError` stays quiet, which is the whole fix here. Joining them back to the caller fires `OnError` again and reopens the bug because every flush site routes a non-nil return straight to `OnError`.

opensearchutil/bulk_indexer_internal_test.go
ryanyuan: Renamed it to `testItemCallback`.

opensearchutil/bulk_indexer.go
ryanyuan: 1. Dropped the new line.
2. Renamed the function to `processBulkResp`.

ryanyuan (approved):

PR #1118, opened 2026-09-04
ci: refresh pkg.go.dev automatically when a release is published

### LINKED ISSUE #499: [BUG] Releases require a manual navigation to the released package URL
### What is the bug?

When releasing 3.1.0 I ran into https://github.com/opensearch-project/opensearch-go/pull/498#issuecomment-2030597365.

The release was not appearing in https://pkg.go.dev/github.com/opensearch-project/opensearch-go/v3. A workaround was to navigate to https://pkg.go.dev/github.com/opensearch-project/opensearch-go/v3@v3.1.0 and hit "request".

I read in https://go.dev/doc/modules/publishing and per instructions tried running `GOPROXY=proxy.golang.org go list -m github.com/opensearch-project/opensearch-go@v3.1.0` but that fails with `not found: ... invalid version: go.mod has post-v3 module path "github.com/opensearch-project/opensearch-go/v3" at revision v3.1.0`. 

1. Make a release per https://github.com/opensearch-project/opensearch-go/blob/main/RELEASING.md#releasing.
2. Observe it not appearing on https://pkg.go.dev/github.com/opensearch-project/opensearch-go/v3

Release appears on https://pkg.go.dev/github.com/opensearch-project/opensearch-go/v3.

### Solution

Either:

a) A release needs to ping the go service and tell it there's a new release. Likely a new workflow that runs every time a release is made.
b) An update to RELEASING.md that says to go to the release URL

magic-peach:
## Description
Releasing v3.1.0 left the version unindexed on pkg.go.dev until a maintainer manually opened the module page and clicked "request" -- pkg.go.dev only lists a module version once the public Go module proxy has fetched it, and the proxy fetches a version lazily, on first request. RELEASING.md step 9 already worked around this with a `go list -m` command a release manager had to remember to run by hand.

This adds a `Refresh pkg.go.dev` GitHub Actions workflow that runs the same request automatically whenever a GitHub release is published: it checks out the tag the release points at, reads the module path from that tag's go.mod, and runs `go list -m <module>@<tag>` against proxy.golang.org, which is what actually makes the release show up on pkg.go.dev. RELEASING.md step 9 now points at the workflow and keeps the manual command as a documented fallback for when the workflow fails or a release is published without going through it.

## Issues Resolved
Fixes #499

## Testing
Ran the exact command the workflow issues, `go list -m github.com/opensearch-project/opensearch-go/v5@v5.0.0-rc5`, against the real module proxy and confirmed it resolves cleanly. Validated the workflow YAML. No Go source changed, so no `go build`/`go vet`/`go test` impact; `go build ./...` and `go vet ./...` still pass clean on this branch.


magic-peach:
@ryanyuan could you please review this pr when you get time
Thanks!

magic-peach:
@sean- could you please review this as well?

sean- (dismissed):

sean- (approved):

PR #1119, opened 2026-09-04
Wrap transport stream errors with the request method and URL

magic-peach:
## Description
Transport errors returned from `opensearchtransport`'s `stream()` carried no information about which request failed, making it hard to correlate a bare error like "unexpected EOF" or "context deadline exceeded" back to a specific call when several requests are in flight concurrently.

This wraps the error with the request method and URL before returning it from `stream()`, using `fmt.Errorf`'s `%w` so `errors.Is`/`errors.As` still see through to the original cause. The URL has its userinfo and query string stripped first: userinfo could carry embedded basic-auth credentials, and the query string can carry secrets that never touch userinfo, such as SigV4 presigned-request signatures or an API key passed as a query parameter. Scheme, host, and path are enough to identify the failing request without risking a credential leaking into logs or an error-tracking system.

This also updates two tests whose expected error strings depended on the old, unwrapped error format: one directly in `opensearchtransport`, and one in `opensearchutil/bulk_indexer_internal_test.go`, whose mock transport error flows through the same `stream()` path.

## Issues Resolved
No tracking issue -- found while reviewing `opensearchtransport`'s error paths. Happy to open one first if maintainers would rather track it before reviewing the code.

## Testing
`go build ./...` and `go vet ./...` pass clean. `go test ./opensearchtransport/... ./opensearchutil/...` passes, including the new `TestTransportStreamErrorRequestContext`. One unrelated pre-existing failure, `TestMetrics/String()`, reproduces identically on `main` without this change (a timezone-dependent formatting assertion), so it is not something this PR touches or affects.

### DIFF
--- a/opensearchtransport/opensearchtransport.go
+++ b/opensearchtransport/opensearchtransport.go
@@ -1814,8 +1814,45 @@ func (c *Transport) stream(req *http.Request) (*http.Response, streamResult, err
 		res, err = c.performSeedFallback(req.Context(), req, &sr)
 	}
 
-	// TODO: Consider wrapping the error with request context.
+	// Wrap the error with the request method and a credential-redacted URL so
+	// callers debugging a failure don't have to correlate it back to the
+	// request by hand. Skipped when err is nil (the common case) since that
+	// would otherwise turn a successful response into a non-nil error.
+	if err != nil {
+		err = fmt.Errorf("%s %s: %w", req.Method, redactedRequestURL(req), err)
+	}
+
 	return res, sr, err
+}
+
+// redactedRequestURL renders req.URL for inclusion in error messages, with
+// userinfo and the query string stripped.
+//
+// setReqURL only ever copies Scheme/Host/Path from the connection URL onto
+// req.URL, never User, so basic-auth credentials configured on a connection
+// (e.g. https://user:pass@host:9200) do not normally reach req.URL. Userinfo
+// is still stripped here as a defense in depth, since req.URL is caller
+// input: it is whatever *http.Request was passed to Stream/Request, which a
+// caller building requests directly against Transport (bypassing
+// opensearchapi) could construct with embedded credentials.
+//
+// The query string is dropped unconditionally because it routinely carries
+// secrets that never touch userinfo -- SigV4 presigned requests place the
+// signature and credential scope in query parameters (e.g. X-Amz-Signature,
+// X-Amz-Credential), and some deployments accept API keys as a query
+// parameter. Method, scheme, host, and path are enough to identify which
+// request failed without risking a leak into logs or error-tracking systems.
+func redactedRequestURL(req *http.Request) string {
+	if req == nil || req.URL == nil {
+		return "<no url>"
+	}
+	u := *req.URL
+	u.User = nil
+	u.RawQuery = ""
+	u.ForceQuery = false
+	u.Fragment = ""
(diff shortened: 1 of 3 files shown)


magic-peach:
@ryanyuan could you please review this pr when you get time
Thanks!

opensearchtransport/opensearchtransport_internal_test.go
```diff
@@ -1060,6 +1063,129 @@ func TestTransportStreamRetries(t *testing.T) {
 	})
 }
 
+// TestTransportStreamErrorRequestContext covers the request-context wrapping
+// applied at the end of Transport.stream: a non-nil error should come back
+// prefixed with "METHOD url: ", and that URL should never carry userinfo or
+// query-string secrets.
+func TestTransportStreamErrorRequestContext(t *testing.T) {
```
ryanyuan, line 1070: This file already leans on testify pretty heavily, so the manual `if err.Error() != ... { t.Fatalf }` and `if !errors.Is(...) { t.Error }` read as out of place here. `require.ErrorIs(t, err, io.EOF)` and `require.True(t, strings.HasPrefix(err.Error(), wantPrefix))` would match the rest of the package and fail fast on the first mismatch instead of reporting and continuing.

opensearchtransport/opensearchtransport_internal_test.go
```diff
@@ -1060,6 +1063,129 @@ func TestTransportStreamRetries(t *testing.T) {
 	})
 }
 
+// TestTransportStreamErrorRequestContext covers the request-context wrapping
```
ryanyuan, line 1066: Can you turn these four subtests, which have the same shape, to be table-driven?

opensearchtransport/opensearchtransport.go
```diff
@@ -1814,10 +1814,47 @@ func (c *Transport) stream(req *http.Request) (*http.Response, streamResult, err
 		res, err = c.performSeedFallback(req.Context(), req, &sr)
 	}
 
-	// TODO: Consider wrapping the error with request context.
+	// Wrap the error with the request method and a credential-redacted URL so
+	// callers debugging a failure don't have to correlate it back to the
+	// request by hand. Skipped when err is nil (the common case) since that
+	// would otherwise turn a successful response into a non-nil error.
+	if err != nil {
+		err = fmt.Errorf("%s %s: %w", req.Method, redactedRequestURL(req), err)
```
ryanyuan, line 1822: shall we use %q instead of %s?

ryanyuan, line 1822: The `"METHOD url: "` format is pinned in three spots with nothing linking them: the `fmt.Errorf` here, the exact-string assert in `opensearchtransport_internal_test.go` (:497 plus the new test's `wantPrefix`), and the prefix/suffix match in `opensearchutil/bulk_indexer_internal_test.go`:872,890. Change the format and all three have to move in sync. Not worth a named type, but a shared format constant (or a small test helper) would keep it in one place.

ryanyuan (commented):
Can you also add an entry to the `CHANGELOG.md`?

magic-peach:
Added an entry under Unreleased > Fixed describing the wrapping and why the URL is scrubbed first.

ryanyuan:
@magic-peach can you look at the comments above?

ryanyuan (approved):

PR #1120, opened 2026-09-05
fix(osgen): omitempty follows required, not nullability

### LINKED ISSUE #1114: [BUG] Range query body generated from opensearchapi.CommonQueryDSLDateRangeQuery always includes an undocumented "from" and "to" field
### What is the bug?
When constructing search request with a range query, the range query will always include a `from` and `to` field in its body, regardless of whether these are set.  These fields appear to be undocumented (no mention on https://docs.opensearch.org/latest/query-dsl/term/range/).

The following program constructs a search request with a date range query less than the current time:

```go
package main

import (
	"fmt"
	"io"
	"os"
	"time"

	"github.com/opensearch-project/opensearch-go/v5/opensearchapi"
)

func main() {
	req := &opensearchapi.SearchReq{
		Indices: []string{"test"},
		Body: &opensearchapi.SearchBody{
			Query: &opensearchapi.CommonQueryDSLQueryContainer{
				Range: map[string]opensearchapi.CommonQueryDSLRangeQuery{
					"field": opensearchapi.NewCommonQueryDSLRangeQueryFromDateRangeQuery(opensearchapi.CommonQueryDSLDateRangeQuery{
						Lt: new(time.Now().Format(time.RFC3339Nano)),
					}),
				},
			},
		},
	}

	httpReq, err := req.GetRequest("POST")
	if err != nil {
		panic(err)
	}

	body, err := io.ReadAll(httpReq.Body)
	if err != nil {
		panic(err)
	}
	defer httpReq.Body.Close()

	fmt.Fprintf(os.Stdout, "request body:\n%s\n", body)

}
```

This outputs the following body:

```
request body:
{"query":{"range":{"field":{"from":null,"lt":"2026-09-02T13:38:49.613275916+10:00","to":null}}}}
```

Note that `from` and `to` where not set but appear as null values in the request body.

While it doesn't appear that this results in a malformed query, its likely these values should either be removed or include an _omitempty_ json tag on the struct?

- Linux, Fedora 44
- opensearch-go v5.0.0-rc5

N/A

N/A

iamrajiv:
### Description

The generator treated a schema's `null` branch as "always present" and dropped `omitempty` from every nullable property (`omitEmpty := !isRequired && !nullable` in `cmd/osgen/resp_walk.go`). As a result, optional nullable request fields such as `CommonQueryDSLDateRangeQuery.From`/`.To` marshalled as `"from":null,"to":null` on every range query, whether or not the caller set them.

### Issues Resolved

Closes #1114


cmd/osgen/resp_walk_test.go
```diff
@@ -750,6 +759,102 @@ func TestRawMessageOmitEmptyPreservesExplicitNull(t *testing.T) {
 	}
 }
 
+// A nullable scalar is a pointer so nil can stand for null, but whether it
```
ryanyuan, line 762: nit: the second sentence narrates the old behaviour.
maybe something like
```
// A nullable scalar is a pointer so nil can stand for null, but whether it
// carries omitempty follows `required` alone: an optional nullable field is
// omitted when unset, and only a required one keeps the bare tag so nil
// marshals as the explicit null the schema demands. Guards #1114.
```

opensearchapi/api_typed_body_marshal_test.go
```diff
@@ -320,3 +320,61 @@ func TestScriptBodiesMarshalDocumentedShapes(t *testing.T) {
 		})
 	}
 }
+
+// An optional nullable request field is a pointer so a caller can leave it
```
ryanyuan, line 324: nit: same.

maybe like 
```
// An optional nullable request field is a pointer so a caller can leave it
// unset, and unset must mean absent on the wire. The spec types range bounds
// `from`/`to` as oneOf [value, null], so a nil bound is omitted rather than
// sent as `"from":null,"to":null`. Guards #1114.
```

ryanyuan (commented):
LGTM.
Thanks for your contribution, just a couple of nits.

iamrajiv:
thanks for review @ryanyuan all done

ryanyuan (dismissed):

iamrajiv:
@sean- @Jakob3xD can you review once

sean- (approved):

PR #1122, opened 2026-09-08
fix(osgen): normalize the git root before the containment check on Windows

MohammedAlkindi:
`make gen-api` fails on every native Windows checkout: `osgen` writes all 1706 files, then `resolveGenRoot` rejects the output directory and exits 1.

`repoRootGit` returns `git rev-parse --show-toplevel` verbatim, which Git for Windows prints with forward slashes. `filepath.Abs` returns backslashes, so the `strings.HasPrefix` check never matches any real subdirectory:

```
osgen: refusing to operate on "C:\Users\...\opensearchapi": outside git root "C:/Users/..." (set OSGEN_SKIP_GIT_CHECK=1 to bypass)
```

Normalizing the git root with `filepath.FromSlash` and `filepath.Clean` before the comparison fixes it; both are no-ops on Linux and macOS. With the change the same command exits 0 and regenerates with no content diff. The developer guide suggests WSL2 on Windows, but `writeIfChanged` in this file already special-cases Windows and the unit matrix runs `windows-latest`.

The existing `GenerateSuite` stubs `repoRoot` to `t.TempDir()`, so it never sees this. The new suite case `TestResolveGenRoot_GitRootWithForwardSlashes` stubs it to `filepath.ToSlash(tmpDir)`, matching Git for Windows. It fails on unmodified `main` with the error above and passes with the change; the rest of `cmd/osgen` and `go vet` are clean. Integration tests not run.

Drafted with AI assistance (Claude Code). I reproduced `make gen-api` failing on unmodified `main` and ran the tests before and after myself.


cmd/osgen/api_cmd.go
```diff
@@ -409,6 +409,12 @@ func resolveGenRoot(root string) (string, error) {
 		if err != nil {
 			return "", fmt.Errorf("%w (set %s=1 to bypass)", err, envSkipGitCheck)
 		}
+		// git rev-parse --show-toplevel prints forward slashes even on Windows
+		// (Git for Windows normalizes its output), while filepath.Abs above
+		// returns a backslash-separated path. Without this, the prefix check
+		// below never matches on Windows and every legitimate subdirectory is
+		// rejected as "outside git root".
+		gitTop = filepath.Clean(filepath.FromSlash(gitTop))
```
ryanyuan, line 417: Normalizing in resolveGenRoot fixes this one caller, but repoRelPath (line 447) takes the same repoRoot() output and hits the same Windows mismatch. Would it be better to normalize once at the producer so both are covered?

e.g.
```
func repoRootGit() (string, error) {
    // ...
    return filepath.Clean(filepath.FromSlash(strings.TrimSpace(string(out)))), nil
}
```

Then drop the line here. (Heads up: the new test stubs repoRoot, so it'd need to move to a unit test on the normalization.)

cmd/osgen/cmd_test.go
```diff
@@ -209,6 +209,22 @@ func (s *GenerateSuite) TestGenerateAPI_RemovesStaleFiles() {
 	s.Require().NotEmpty(entries)
 }
 
+// TestResolveGenRoot_GitRootWithForwardSlashes covers what `git rev-parse
+// --show-toplevel` prints under Git for Windows: a forward-slash path, while
+// filepath.Abs returns the native separator. The containment check must still
+// accept a subdirectory of the repository. On Linux and macOS the forward-slash
+// form is already native, so this passes there regardless.
+func (s *GenerateSuite) TestResolveGenRoot_GitRootWithForwardSlashes() {
```
ryanyuan, line 217: With the normalization moving into repoRootGit (see other comment), this test's repoRoot stub bypasses it and no longer covers the fix. Replace it with a unit test on the helper, which runs on every platform instead of only windows-latest:

```
func TestNormalizeRepoRoot(t *testing.T) {
    // git for Windows prints forward slashes
    got := normalizeRepoRoot("  C:/Users/x/repo\n")
    require.Equal(t, filepath.Clean("C:/Users/x/repo"), got)
}
```

ryanyuan (changes requested):

MohammedAlkindi:
Both done: normalization moved into `repoRootGit` behind a `normalizeRepoRoot` helper, the caller-side line dropped, and the suite test replaced by a unit test on the helper so it runs on every platform.

One correction, which I had repeated in a commit message before checking it. `repoRelPath` does not share the mismatch: `filepath.Rel` calls `Clean` on both arguments and `Clean` converts forward slashes on Windows, so it returns `a\b.go` either way. Only `resolveGenRoot` compares raw strings, through `strings.HasPrefix`.

Producer-side is still the better place, so the change stands. The comment and changelog now give that reason instead.

ryanyuan (approved):

sean- (approved):

PR #1134, opened 2026-09-15
close timed-out HTTP/2 connections before retry

### LINKED ISSUE #1121: [BUG] Timeout retries reuse the same stale HTTP/2 connection after endpoint cutover
### What is the bug?

With `RequestTimeout` and `EnableRetryOnTimeout` enabled, timeout retries can reuse the exact same stale HTTP/2 connection for every attempt.

This matters when a stable endpoint hostname moves from an old backend to a new backend (for example, during a blue/green cutover). Existing HTTP/2 connections can remain pinned to the old backend after new dials would resolve or route to the new backend. If the old backend stops responding, all retry attempts remain on that stale connection and no new dial occurs.

`RequestTimeout` bounds each attempt, but the retry does not provide failover in this case.

The following deterministic local reproducer uses two TLS/HTTP2 test servers behind one logical hostname:

1. Warm up a pooled connection to the old backend.
2. Make the old backend stop responding.
3. Switch all future dials for the hostname to the healthy new backend.
4. Send a request with `RequestTimeout`, `EnableRetryOnTimeout`, and two retries.

`go.mod`:

```go
module example.com/opensearch-go-timeout-repro

go 1.25.9

require github.com/opensearch-project/opensearch-go/v4 v4.7.3
```

`main.go`:

```go
package main

import (
	"context"
	"crypto/tls"
	"fmt"
	"io"
	"log"
	"net"
	"net/http"
	"net/http/httptrace"
	"net/http/httptest"
	"sync/atomic"
	"time"

	opensearch "github.com/opensearch-project/opensearch-go/v4"
	"github.com/opensearch-project/opensearch-go/v4/opensearchapi"
)

func main() {
	var oldBlackhole atomic.Bool
	oldBackend := newHTTP2Server(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		if oldBlackhole.Load() {
			<-r.Context().Done()
			return
		}
		_, _ = io.WriteString(w, "{}")
	}))
	defer oldBackend.Close()

	newBackend := newHTTP2Server(http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
		_, _ = io.WriteString(w, "{}")
	}))
	defer newBackend.Close()

	var useNew atomic.Bool
	runCase("keepalive", false, oldBackend.Listener.Addr().String(), newBackend.Listener.Addr().String(), &useNew, &oldBlackhole)
	runCase("no_keepalive", true, oldBackend.Listener.Addr().String(), newBackend.Listener.Addr().String(), &useNew, &oldBlackhole)
}

func runCase(label string, disableKeepAlives bool, oldAddr, newAddr string, useNew, oldBlackhole *atomic.Bool) {
	oldBlackhole.Store(false)
	useNew.Store(false)

	var dials atomic.Int64
	dialer := &net.Dialer{Timeout: time.Second}
	transport := http.DefaultTransport.(*http.Transport).Clone()
	transport.ForceAttemptHTTP2 = true
	transport.DisableKeepAlives = disableKeepAlives
	transport.TLSClientConfig = &tls.Config{InsecureSkipVerify: true} // local test only
	transport.DialContext = func(ctx context.Context, _, _ string) (net.Conn, error) {
		dials.Add(1)
		addr := oldAddr
		if useNew.Load() {
			addr = newAddr
		}
		return dialer.DialContext(ctx, "tcp", addr)
	}

	client, err := opensearchapi.NewClient(opensearchapi.Config{Client: opensearch.Config{
		Addresses:            []string{"https://cluster.test"},
		Transport:            transport,
		MaxRetries:           2,
		EnableRetryOnTimeout: true,
		RequestTimeout:       150 * time.Millisecond,
	}})
	if err != nil {
		log.Fatal(err)
	}

	if _, err := client.Info(traceContext(label+" warmup"), nil); err != nil {
		log.Fatal(err)
	}

	oldBlackhole.Store(true)
	useNew.Store(true)

	ctx, cancel := context.WithTimeout(traceContext(label+" cutover"), time.Second)
	started := time.Now()
	_, err = client.Info(ctx, nil)
	cancel()
	fmt.Printf("%s result: err=%v elapsed=%s dials=%d\n", label, err, time.Since(started).Round(time.Millisecond), dials.Load())
	_ = client.Close()
}

func traceContext(label string) context.Context {
	return httptrace.WithClientTrace(context.Background(), &httptrace.ClientTrace{
		GotConn: func(info httptrace.GotConnInfo) {
			fmt.Printf("%s remote=%s reused=%v was_idle=%v\n", label, info.Conn.RemoteAddr(), info.Reused, info.WasIdle)
		},
	})
}

func newHTTP2Server(handler http.Handler) *httptest.Server {
	server := httptest.NewUnstartedServer(handler)
	server.EnableHTTP2 = true
	server.StartTLS()
	return server
}
```

Observed output (ports abbreviated):

```text
keepalive warmup remote=127.0.0.1:<old> reused=false was_idle=false
keepalive cutover remote=127.0.0.1:<old> reused=true was_idle=true
keepalive cutover remote=127.0.0.1:<old> reused=true was_idle=false
keepalive cutover remote=127.0.0.1:<old> reused=true was_idle=false
keepalive result: err=context deadline exceeded elapsed=458ms dials=1

no_keepalive warmup remote=127.0.0.1:<old> reused=false was_idle=false
no_keepalive cutover remote=127.0.0.1:<new> reused=false was_idle=false
no_keepalive result: err=<nil> elapsed=7ms dials=2
```

All three attempts in the keep-alive case reuse the old HTTP/2 connection. The dial target has already changed, but `DialContext` is never called again. Disabling keep-alives is the control: it dials the new backend and succeeds immediately.

After an attempt times out and is selected for retry, the retry should not reuse the same timed-out HTTP/2 connection. It should establish a fresh connection (thereby invoking `DialContext`/DNS again), or the client should otherwise provide a scoped way for timeout retries to bypass the failed pooled connection.

At minimum, a single stale HTTP/2 connection should not consume every configured retry attempt.

- opensearch-go v4.7.3 (latest stable release)
- Go 1.25.9
- macOS 26.6.2, arm64
- Deterministic local reproduction with TLS and HTTP/2; no OpenSearch server is required
- Separately reproduced against an isolated Amazon OpenSearch Service 3.3 VPC domain

No.

I also reproduced this against a fresh, isolated Amazon OpenSearch Service 3.3 VPC domain. A one-node domain was changed from one instance type to another to trigger a Blue/Green deployment. An in-VPC opensearch-go v4.7.3 probe ran the keep-alive and no-keep-alive variants concurrently every five seconds. Both negotiated HTTP/2.

Sanitized managed-service results:

- DNS and the no-keep-alive client moved to the replacement backend about three minutes before the old pooled connection stopped responding.
- The keep-alive client then failed 28 consecutive requests (84 attempts). Every attempt used the old backend and reported `reused=true`; no new dial or DNS lookup occurred during those failures.
- The no-keep-alive control recorded zero failures and continued succeeding against the replacement backend in roughly 10-15 ms.
- The keep-alive client recovered after about seven minutes. On the recovery request, the first attempt timed out on the old connection and the second attempt performed a fresh DNS lookup/dial to the replacement backend and succeeded.

Resource names, account identifiers, endpoint names, and addresses are intentionally omitted. The exact failure duration appears environment-dependent; the connection-reuse and retry behavior matched the local reproduction.

The practical workaround is `DisableKeepAlives: true`, but that adds a TCP/TLS handshake to every request. `IdleConnTimeout` and periodic `CloseIdleConnections` do not reliably address a hot HTTP/2 connection, because it may not become idle while requests keep arriving.

The relevant interaction appears to be between timeout retry handling and the underlying HTTP/2 `ClientConn`: canceling a timed-out stream does not retire the pooled connection, so the next attempt is multiplexed onto it again.

iprithv:
`RequestTimeout` cancels the attempt context, which RSTs the HTTP/2 stream but leaves the `ClientConn` in `http.Transport`'s pool. With `EnableRetryOnTimeout`, every retry was multiplexed onto the same connection and `DialContext` never ran.

That is the failure after an Amazon OpenSearch Service blue/green cutover: DNS already points at the replacement backend, but keep-alive HTTP/2 stays pinned to the old one. The issue reproduced it locally (all retries `reused=true`, `dials=1`) and against a VPC domain (28 consecutive failures, recovery only after ~7 minutes). `DisableKeepAlives` works around it by dialing every request. `Request.Close` and `CloseIdleConnections` do not retire a hot HTTP/2 connection.

A timed-out attempt now closes the `net.Conn` recorded by `httptrace.GotConn`, so the next RoundTrip dials. The default path (no `RequestTimeout`, no `EnableRetryOnTimeout`) is unchanged.

`TestTimeoutRetryDoesNotReuseHTTP2Conn` is the two-server HTTP/2 repro from the issue. On unmodified main it fails with `context deadline exceeded` and a single dial; with this change the retry dials the new backend and succeeds. HTTP/2 is asserted on the warmup response so the test cannot pass on HTTP/1.1 by accident.

Fixes #1121


opensearchtransport/timeout_conn.go
```diff
@@ -0,0 +1,69 @@
...
+	"context"
+	"net"
+	"net/http/httptrace"
+	"sync"
+)
+
+// attemptConn holds the net.Conn httptrace.GotConn reports for one RoundTrip.
+// GotConn may run on a different goroutine than RoundTrip's caller (HTTP/2),
+// so the slot is mutex-guarded.
+type attemptConn struct {
```
ryanyuan: The repo convention is to nest guarded state inside the mutex field, so every access spells out the lock it needs:

```go
type attemptConn struct {
    mu struct {
        sync.Mutex
        conn net.Conn
    }
}
```

Then `a.mu.conn` reads as guarded right at the call site, instead of leaving the mu/conn pairing as a convention the reader has to infer. `set` and `close` become `a.mu.conn = c`. The internal test that reaches `slot.mu.Lock(); slot.conn` would move to `slot.mu.conn`.

opensearchtransport/timeout_retry_h2_internal_test.go
```diff
@@ -0,0 +1,148 @@
...
+	"net/http/httptrace"
+	"net/url"
+	"sync/atomic"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestAttemptConnCloseNil(t *testing.T) {
```
ryanyuan: Could these get `t.Parallel()`? Each test stands up its own servers with no shared state, and the rest of the package already parallelizes (see `address_resolver_internal_test.go`). The two unit tests are the easy wins, and the H2 reproducer can too, since its listeners are per-test.

opensearchtransport/timeout_retry_h2_internal_test.go
```diff
@@ -0,0 +1,148 @@
...
+	var a attemptConn
+	require.NotPanics(t, a.close)
+}
+
+func TestWithAttemptConnTracePreservesGotConn(t *testing.T) {
+	var seen atomic.Bool
+	ctx := httptrace.WithClientTrace(context.Background(), &httptrace.ClientTrace{
```
ryanyuan: Prefer `t.Context()` over `context.Background()` here. It cancels when the test ends, so a hung `GotConn` hook or request can't outlive the test.

opensearchtransport/timeout_retry_h2_internal_test.go
```diff
@@ -0,0 +1,148 @@
...
+	}
+
+	u, err := url.Parse("https://cluster.test")
+	require.NoError(t, err)
+	tp, err := New(Config{
+		URLs:                 []*url.URL{u},
+		Transport:            transport,
+		MaxRetries:           2,
+		EnableRetryOnTimeout: true,
+		RequestTimeout:       150 * time.Millisecond,
```
ryanyuan: 150ms as the per-attempt timeout is tight for CI. Shared-CPU runners give roughly 1s scheduling granularity, so the healthy retry to `newBackend` has to complete inside 150ms or the test flakes. Raising `RequestTimeout` to ~500ms keeps the same assertion (retry dials the replacement) with margin.

opensearchtransport/timeout_retry_h2_internal_test.go
```diff
@@ -0,0 +1,148 @@
...
+		HealthCheck:          NoOpHealthCheck,
+		NodeStatsInterval:    -1,
+	})
+	require.NoError(t, err)
+	t.Cleanup(func() { _ = tp.Close() })
+
+	warmup, err := http.NewRequest(http.MethodGet, "/", nil)
```
ryanyuan: Same `t.Context()` point as above, one layer down. `http.NewRequest` seeds the request with `context.Background()`, so `warmup` and `cutover` both start from Background before the `WithTimeout` wrap. `http.NewRequestWithContext(t.Context(), ...)` ties them to the test lifecycle. (Same applies to `cutover` on line 119.)

opensearchtransport/timeout_conn.go
```diff
@@ -0,0 +1,69 @@
...
+	mu   sync.Mutex
```
sean-: Please use this idiom:

```go
struct mu {
  sync.Mutex

  conn net.Conn
}
```

opensearchtransport/timeout_conn.go
```diff
@@ -0,0 +1,69 @@
...
+
+// close retires the captured connection. Canceling an HTTP/2 stream does not
+// retire the ClientConn, so a timeout retry would otherwise be multiplexed
+// onto the same (possibly black-holed) connection. Closing the net.Conn
+// forces the next RoundTrip to dial. A nil or empty slot is a no-op.
+func (a *attemptConn) close() {
+	a.mu.Lock()
+	conn := a.conn
+	a.conn = nil
+	a.mu.Unlock()
```
sean-: why isn't this a `defer a.mu.Unlock()` so it fires after `conn.Close()` below?  Right now, there's a race where an `attemptConn`'s `.conn` could be replaced.

opensearchtransport/timeout_conn.go
iprithv: sure, updated. conn now lives inside mu, so set/close assign a.mu.conn and the test reads slot.mu.conn.

opensearchtransport/timeout_retry_h2_internal_test.go
iprithv: git it, updated. all three tests call t.Parallel(). Each one owns its servers/listeners.

opensearchtransport/timeout_retry_h2_internal_test.go
iprithv: Done, the trace is now attached to t.Context() so a hung GotConn/request is canceled with the test.

opensearchtransport/timeout_retry_h2_internal_test.go
iprithv: makes sense, updated. RequestTimeout is 500ms. Also raised the cutover’s outer deadline to 3s so the first 500ms timeout doesn’t eat the whole Stream window.

opensearchtransport/timeout_retry_h2_internal_test.go
iprithv: Done, warmup and cutover both use http.NewRequestWithContext(t.Context(), ...).

sean- (dismissed):

sean- (approved):

PR #1148, opened 2026-09-15
fix(osgen): send version=0 on the wire

iprithv:
External versioning allows `version ≥ 0`, but the generator typed it as `int` and dropped `0` under the `!= 0` guard. `IndexParams{Version: 0, VersionType: VersionTypeExternal}` sent `version_type=external` and omitted `version`, so the server used internal versioning. Same on delete, get, create, exists, termvectors, etc.

Promote those `version` params to `*int` (same path as `if_seq_no=0` / search `size=0`), regenerate, and add a `version=0` params test.


sean-:
Why not `,omitzero`?

iprithv:
> Why not `,omitzero`?

Thanks for checking @sean-, omitzero wouldn't help here. these are query-string params, not JSON: get() used if r.Version != 0 { set("version", ...) }, so a deliberate 0 never reached the wire. omitzero is a JSON struct tag, and on a plain int it still drops 0 because 0 is the zero value.

We need to tell “unset” apart from “send 0”. That requires a pointer (nil omits, &0 sends 0), which is the same path we already use for if_seq_no, search size, and requests_per_second.

ryanyuan:
@iprithv Thank you for clearing that up. Just chatted with Sean about this offline. He's good with the approach since these are query-string params rather than JSON body fields, so a pointer is what you need to tell "unset" apart from a deliberate 0.

ryanyuan (approved):

PR #1155, opened 2026-09-19
fix(opensearchtransport): format dead_since and overloaded_since in UTC

magic-peach:
## Description
ConnectionMetric.String() formatted the DeadSince and OverloadedSince timestamps with .Local() before printing them, even though nanoToTime always reconstructs these timestamps in UTC. This made the debug string depend on the host machine's timezone, and TestMetrics/String() asserted a hardcoded hour expecting a whole hour offset, so it failed deterministically on any machine with a non whole hour UTC offset such as IST (UTC+5:30).

This drops the .Local() call so the string is formatted directly from the already UTC timestamp, and tightens the test from a timezone tolerant regex to an exact string match now that the output is deterministic.

## Issues Resolved
No tracking issue. This flaky test was previously called out as a known pre-existing failure in the discussion on PR #1119, and reproduces identically on main today.

## Testing
go vet ./opensearchtransport/... and gofmt -l pass clean. go test ./opensearchtransport/... passes, including the updated TestMetrics/String() assertion, verified on a non UTC machine where it previously failed deterministically.

## By submitting this pull request, I confirm my contribution is made under the terms of the Apache 2.0 license.


sean- (dismissed):
Thank you for catching this!

sean- (approved):

